PR #129, opened 2020-10-30
obsolete  -  Allow distinct clearing price per lease duration bucket

guggero:
This PR contains the client part of the change to switch over to distinct lease duration buckets for batches.

**NOTE: This PR contains breaking RPC changes for anything batch related.**
The batch version that is checked during the account subscription authentication prevents old clients from connecting to the new server once it's upgraded. This results in the following error message: `version 0 mismatches server version`.

TODO:
 - Request shutdown when error `version 0 mismatches server version` is detected.
 - Verify batch validation against server PR.


poolrpc/auctioneer.proto
```diff
@@ -441,48 +441,56 @@ message OrderMatchPrepare {
...
     */
     uint32 clearing_price_rate = 2;
+}
+
+message OrderMatchPrepare {
+    /*
+    Maps the distinct lease duration markets to the orders that were matched
+    within and the discovered market clearing price.
+    */
+    map<uint32, MatchedMarket> matched_markets = 1;
```
Roasbeef, line 451: 👍

order/rpc_parse.go
```diff
@@ -228,21 +228,43 @@ func ParseRPCBatch(prepareMsg *poolrpc.OrderMatchPrepare) (*Batch,
...
+			if err != nil {
+				return nil, fmt.Errorf("error parsing matched "+
+					"order: %v", err)
+			}
+
+			// It is imperative that all matched orders have the
+			// correct lease duration. Otherwise assumptions in the
+			// batch verifier won't be correct. That's why we
+			// validate this here already.
+			for _, matchedOrder := range b.MatchedOrders[ourOrder] {
```
Roasbeef, line 255: Why validate this here (when _parsing_ the batch) vs later in the validation logic?

clientdb/batch_snapshot.go
```diff
@@ -382,8 +383,19 @@ func getSnapshotBuckets(tx *bbolt.Tx) (*bbolt.Bucket, *bbolt.Bucket,
 }
 
 func serializeLocalBatchSnapshot(w io.Writer, b *LocalBatchSnapshot) error {
+	// The previous batch versions had a single clearing price. We'll
+	// extract it from the map which will only contain one entry for that
+	// oder version. For new batch versions we just store a zero price.
```
Roasbeef, line 388: oder -> older

order/rpc_parse.go
Roasbeef: I guess the parsing is part of validation in a sense, as it can cause batch validation failures. Also with this placement, we can catch an invalid batch earlier in the processing cycle.

Roasbeef (changes requested):

order/rpc_parse.go
guggero: To keep the diff reasonably small, we discard the `duration->orders` mapping after verifying that all orders within a bucket have the correct duration. It's not exactly parsing logic. So we could thread it through to the batch verifier. But as you say, like this we catch format inconsistencies earlier.

guggero:
zNeeded to re-open because my repo wasn't recognized as a fork anymore.

PR #134, opened 2020-11-02
account: ensure only one outstanding account reservation exists

wpaulino:
The auctioneer only expects one outstanding account reservation to exist per trader LSAT, so we should enforce this on the trader's side as well.

### DIFF
--- a/account/manager.go
+++ b/account/manager.go
@@ -140,6 +140,12 @@ type Manager struct {
 	// Finalize message for.
 	pendingBatchMtx sync.Mutex
 
+	// reservationMtx prevents a trader from attempting to have more than
+	// once active reservation at a time when creating new accounts. This is
+	// done to ensure an account picks up the correct reservation once its
+	// time to fund it.
+	reservationMtx sync.Mutex
+
 	wg   sync.WaitGroup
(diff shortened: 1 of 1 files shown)


account/manager.go
```diff
@@ -306,7 +312,11 @@ func (m *Manager) InitAccount(ctx context.Context, value btcutil.Amount,
 
 	// With our key obtained, we'll reserve an account with our auctioneer,
 	// who will provide us with their base key and our initial per-batch
-	// key.
+	// key. We'll make sure to acquire the reservation lock throughout the
+	// account funding process to ensure we use the same reservation, as
+	// only one can be active per trader LSAT.
+	m.reservationMtx.Lock()
```
Roasbeef, line 318: Why not lock above, for the entire method? Otherwise assuming we have 1 BTC, and two requests from in for 1 BTC accounts, then one of them will get stuck on coin selection, similar to what happened with Reza.

Roasbeef (changes requested):

zech564:
Yes

Roasbeef (approved):
LGTM 👾

PR #142, opened 2020-11-07
Allow distinct clearing price per lease duration bucket

guggero:
This PR contains the client part of the change to switch over to distinct lease duration buckets for batches.

UPDATE: Made this non-breaking. Clients with batch version `0` won't be able to submit orders with a lease duration != 2016.

UPDATE2: Depends on #193 which was split off to make the server changes easier to split into multiple PRs.


Roasbeef:
I think this is ready for review now?

guggero:
Yes, has been for a while.

poolrpc/auctioneer.proto
```diff
@@ -956,6 +956,32 @@ message ServerOrderStateResponse {
...
+    ACCEPTING_ORDERS indicates that we're accepting orders for this bucket, but
+    not yet clearing for this duration.
+    */
+    ACCEPTING_ORDERS = 2;
+
+    /*
+    CLEARING_MARKET indicates that we're accepting orders, and fully clearing
+    the market for this duration.
+    */
+    CLEARING_MARKET = 3;
```
halseth, line 990: Call this `market open`?

poolrpc/auctioneer.proto
```diff
@@ -441,48 +441,56 @@ message OrderMatchPrepare {
...
+    */
+    map<uint32, MatchedMarket> matched_markets = 1;
 
     /*
     A list of the user's own accounts that are being spent by the matched
     orders. The list contains the differences that would be applied by the
     server when executing the orders.
     */
-    repeated AccountDiff charged_accounts = 3;
+    repeated AccountDiff charged_accounts = 2;
```
halseth, line 458: I think we should avoid shifting the protos here. Feels like it could lead to some unexpected issues.

clientdb/batch_snapshot.go
```diff
@@ -426,13 +438,31 @@ func serializeLocalBatchSnapshot(w io.Writer, b *LocalBatchSnapshot) error {
 		}
 	}
 
+	// New batch versions have an additional map of duration->price that we
```
halseth, line 441: Great solution 👍

rpcserver.go
```diff
@@ -1810,10 +1810,19 @@ func (s *rpcServer) prepareLeasesResponse(ctx context.Context,
 					bidDuration = match.Order.(*order.Bid).LeaseDuration
 				}
 
+				// The clearing price is dependent on the lease
+				// duration, except for older batch versions
+				// where there was just one single duration.
+				duration := bidDuration
+				if batch.Version != order.DefaultVersion {
```
halseth, line 1828: I don't understand why this is needed. Won't the bid duration always match the duration for the batch it was matched in?

poolrpc/auctioneer.proto
```diff
@@ -965,21 +991,18 @@ message TermsResponse {
...
-    The set of lease durations the market is currently accepting. If a duration
-    is mapped to false, then this means the durations are being accepted, but
-    the market may be paused (not clearing batches). This might happen if a new
-    duration is added, and the auction wishes to build of sufficient market
-    thickness before starting to clear batches.
+    The set of lease durations the market is currently accepting and the state
+    the duration buckets currently are in.
     */
-    map<uint32, bool> lease_durations = 4;
+    map<uint32, DurationBucketState> lease_duration_buckets = 4;
```
halseth, line 1013: wouldn't it be pretty easy to not break the API by adding this as a new field?

halseth (changes requested):
First pass. Main question is whether we can avoid breaking the API.

poolrpc/auctioneer.proto
guggero: Yeah, made all changes non-breaking for older clients.

rpcserver.go
guggero: This is for legacy legacy orders where the durations were defined free style. When de-serializing those batch snapshots, the single clearing price is put into the map with the 2016 duration as the key. Otherwise we wouldn't know whether to look it up using the bid or ask duration (as they could differ before the uniform 2016 duration was introduced).

guggero:
I was able to make everything non-breaking!
Still need to do some manual tests to confirm, but I think this PR is ready for a second look.

clientdb/batch_snapshot.go
```diff
@@ -465,6 +495,40 @@ func deserializeLocalBatchSnapshot(r io.Reader) (*LocalBatchSnapshot, error) {
...
+	// additional data to read here, we can assume we have such an order.
+	err = ReadElements(r, &numPrices)
+	switch err {
+	// Successful read, continue with reading the prices.
+	case nil:
+
+	// If it wasn't found in the database, then we'll assume the default
+	// value.
+	case io.EOF, io.ErrUnexpectedEOF:
+		b.ClearingPrices[order.LegacyLeaseDurationBucket] = clearingPrice
```
wpaulino, line 514: Could avoid the use of `EOF` markers by instead checking if the clearing price is non-zero signaling an old batch.

rpcserver.go
```diff
@@ -239,6 +242,14 @@ func (s *rpcServer) consumePendingOpenChannels(
 			}
 
 			rpcLog.Errorf("Unable to read channel event: %v", err)
+
+			// If the lnd node shut down, there's no use continuing.
+			if err == io.EOF || err == io.ErrUnexpectedEOF ||
+				status.Code(err) == codes.Unavailable {
```
wpaulino, line 248: :+1:

wpaulino (commented):
Non-breaking changes look good on first pass!

poolrpc/trader.proto
```diff
@@ -871,7 +871,11 @@ message LeaseDurationRequest {
 }
 
 message LeaseDurationResponse {
-    map<uint32, bool> lease_durations = 1;
+    /*
+    The set of lease durations the market is currently accepting and the state
+    the duration buckets currently are in.
+    */
+    map<uint32, DurationBucketState> lease_duration_buckets = 1;
```
halseth, line 878: this breaks clients of the trader service.

poolrpc/auctioneer.proto
```diff
@@ -483,6 +498,12 @@ message OrderMatchPrepare {
     informed that a software update is required.
     */
     uint32 batch_version = 9;
+
+    /*
+    Maps the distinct lease duration markets to the orders that were matched
```
halseth, line 503: Nice! 👍

order/rpc_parse.go
```diff
@@ -274,7 +296,6 @@ func ParseRPCBatch(prepareMsg *poolrpc.OrderMatchPrepare) (*Batch,
 	}
 
 	// Convert clearing price, fee rate and rebate.
-	b.ClearingPrice = FixedRatePremium(prepareMsg.ClearingPriceRate)
```
halseth, line 277: should be removeed in later commit?

clientdb/batch_snapshot.go
```diff
@@ -382,8 +383,19 @@ func getSnapshotBuckets(tx *bbolt.Tx) (*bbolt.Bucket, *bbolt.Bucket,
 }
 
 func serializeLocalBatchSnapshot(w io.Writer, b *LocalBatchSnapshot) error {
+	// The previous batch versions had a single clearing price. We'll
+	// extract it from the map which will only contain one entry for that
+	// older version. For new batch versions we just store a zero price.
+	var clearingPrice order.FixedRatePremium
+	if len(b.ClearingPrices) == 1 {
```
halseth, line 390: keep in mind that this might have only one element for new version as well, if only one bucket had matches.

halseth (changes requested):
Great work on this PR! The latest version reads well to me. 

Also needs a rebase!

poolrpc/trader.proto
guggero: You're right, maybe someone is already using the RPC or REST interface. Deprecated the legacy format instead of replacing it.

clientdb/batch_snapshot.go
guggero: Good idea! I thought it might screw with our unit tests if we always write a zero price. But it turns out this works well.

clientdb/batch_snapshot.go
guggero: Removed, based on Wilmer's suggestion above.

clientdb/batch_snapshot_test.go
```diff
@@ -112,44 +108,91 @@ var testSnapshot = &LocalBatchSnapshot{
...
-			a.TraderKey.PubKey.Curve = nil
-			a.AuctioneerKey.Curve = nil
-			a.BatchKey.Curve = nil
-		}
+			post, err := deserializeLocalBatchSnapshot(&buf)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if !reflect.DeepEqual(pre, post) {
```
wpaulino, line 180: Nit: Is `require.Equal` any better here?

wpaulino (approved):

halseth (approved):
Awesome, this turned out very clean in the end 👌 

Only nit from me is that I think we should make deprecated fields more clear by removing all but the deprecated message from their doc.

poolrpc/auctioneer.proto
```diff
@@ -979,6 +1005,7 @@ message TermsResponse {
 
     /*
     The maximum order duration in blocks currently allowed by the auctioneer.
+    Deprecated, use explicit order duration from lease_duration_buckets.
```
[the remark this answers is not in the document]
guggero: Good idea, done :+1:

guggero:
Rebased and addressed all comments.

Even though this is ready to be merged now, I think we should wait until the server part is close to being merged as well.

clientdb/batch_snapshot.go
```diff
@@ -481,6 +504,34 @@ func deserializeLocalBatchSnapshot(r io.Reader) (*LocalBatchSnapshot, error) {
...
+		duration  uint32
+		price     order.FixedRatePremium
+	)
+	err = ReadElements(r, &numPrices)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := uint32(0); i < numPrices; i++ {
+		err = ReadElements(r, &duration, &price)
```
halseth, line 527: this is probably okay since these are just `uint32`, but triggers my spidey sense :P Could define these in the inner scope to make clear that each iteration is not overwriting the previous value.

halseth (approved):
Re-ACK ✅

clientdb/batch_snapshot.go
```diff
@@ -465,6 +488,34 @@ func deserializeLocalBatchSnapshot(r io.Reader) (*LocalBatchSnapshot, error) {
 		b.MatchedOrders[nonce] = append(b.MatchedOrders[nonce], m)
 	}
 
+	// Older batches had a single clearing price instead of a map. If we
+	// have a non-zero price, it means this is an old snapshot and we don't
+	// need to read any further.
+	if clearingPrice > 0 {
+		b.ClearingPrices[order.LegacyLeaseDurationBucket] = clearingPrice
+
```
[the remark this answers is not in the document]
guggero: Hmm, I thought I already replied to this, but apparently that comment got lost. I added the newline on purpose to make the return stand out a bit more since it's an early exit from the function.

clientdb/batch_snapshot_test.go
guggero: Yes, the diff seems to be much cleaner :+1:

guggero:
Replaced by #199.

Issue #145, opened 2020-11-08
Convertion between satoshi and units issue


Pantamis:
Thanks for making pool ! I am trying it in LiT now that onchain fees are lower.

I submit an ask order but I found a weird issue with satoshis and units convertion:

`$ pool --rpcserver=localhost:10009 orders submit ask 6500000 [trader key] ...`
`> [pool] unable to parse order params: minimum channel amount 0.0065 BTC must be a multiple of 0.001 BTC`

Notice that the amount 6500000 = 6 500 000 satoshis is displayed as 0.0065 BTC, which is a mistake (it must be 0.065 BTC).

However this order is accepted and amount in BTC are correct:

`$ pool --rpcserver=localhost:10009 orders submit ask 6000000`
`> Ask Amount: 0.06 BTC`
`...`

guggero:
Thank you for the report. Just to clarify, this isn't a conversion error. The `minimum channel amount` is the smallest amount you want to have matched in one match. You can set it by specifying `--min_chan_amt`. The default value is 10% of the full order amount, which is the 0.0065 BTC that you saw.
What should be fixed though is that those default 10% should always be rounded up to be a multiple of the order unit, so in your case 0.007.

PR #151, opened 2020-11-09
pool: round default min chan amt to nearest unit

wpaulino:
Fixes https://github.com/lightninglabs/pool/issues/145.

### DIFF
--- a/cmd/pool/order.go
+++ b/cmd/pool/order.go
@@ -105,31 +105,29 @@ func parseCommonParams(ctx *cli.Context, blockDuration uint32) (*poolrpc.Order, 
 	}
 
 	// If the minimum channel amount flag wasn't provided, use a default of
-	// 10%, but make sure it doesn't dip below the minimum allowed.
-	minOrderChanAmt := btcutil.Amount(order.BaseSupplyUnit)
+	// 10% and round to the nearest unit.
 	minChanAmt := btcutil.Amount(ctx.Uint64("min_chan_amt"))
 	if minChanAmt == 0 {
-		minChanAmt = btcutil.Amount(params.Amt) / 10
-		if minChanAmt < minOrderChanAmt {
-			minChanAmt = minOrderChanAmt
-		}
+		minChanAmt = order.RoundToNextSupplyUnit(
+			btcutil.Amount(params.Amt) / 10,
+		).ToSatoshis()
 	}
 
 	// Verify the minimum channel amount flag has been properly set.
 	switch {
-	case minChanAmt%minOrderChanAmt != 0:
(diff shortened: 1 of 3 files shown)


cmd/pool/order.go
```diff
@@ -82,6 +82,14 @@ func promptForConfirmation(msg string) bool {
 	}
 }
 
+// roundUp rounds up `n` to the nearest multiple.
+func roundUp(n, multiple btcutil.Amount) btcutil.Amount {
```
Roasbeef, line 86: Algo looks sound at a glance. If we want to clamp things down further, it shouldn't be too involved to create a `testing/quick` test to ensure the output of this function, given arbitrary inputs (within our range bounds) should always return something that's a multiple of our unit size. This can likely live where the unit is defined.

guggero: Could make this a method on the `SupplyUnit` itself for easy testing.

guggero (approved):
LGTM!

order/supplyunit.go
```diff
@@ -21,6 +21,12 @@ func NewSupplyFromSats(sats btcutil.Amount) SupplyUnit {
 	return SupplyUnit(uint64(sats) / uint64(BaseSupplyUnit))
 }
 
+// RoundToNextSupplyUnit computes and rounds to the next whole number of supply
```
guggero, line 24: Can we remove the TODO on L19 or is there still something more to do, @Roasbeef?

guggero (approved):
LGTM, nice test!

order/supplyunit.go
Roasbeef: Nope this should be good, was just a hunch to revisit to ensure things are consistent. I think we need to do a sweep w.r.t where we use either of these methods, we may want to unify and ensure we always round up.

order/supplyunit.go
```diff
@@ -10,7 +10,7 @@ type SupplyUnit uint64
 const (
 	// BaseSupplyUnit is the smallest channel that can be bought or sold in
 	// the system. These units are expressed in satoshis.
-	BaseSupplyUnit SupplyUnit = 100_000
+	BaseSupplyUnit btcutil.Amount = 100_000
```
Roasbeef, line 13: Why change? To cutdown on casting somewhere else?

Roasbeef (approved):
LGTM 🍥

order/supplyunit.go
wpaulino: It just seemed strange to define it as `SupplyUnit` as you could then invoke `ToSatoshis` on it resulting in `100000^2`.

halseth (approved):
ACK

Issue #149, opened 2020-11-09
Withdraw transaction not broadcasted


louneskmt:
Hi there!

I was using Lightning Pool with LiT v0.3.0, and I tried to withdraw some funds from my pool account. I got the following error: `[pool] rpc error: code = Unknown desc = account must be in StateOpen to bemodified`.

Meanwhile, I had some troubles with the connection to my node (I actually proxy my node onion url to localhost:10009 with socat, and make LiT connect to this last), so I restarted my node.

Now Pool seems to have taken in account the withdraw: my account `outpoint` and `latest_txid` have been updated, as well as the `available_balance` and `value` fields. The thing is that the transaction doesn't appear in any explorers, nor in my own node's, after almost two hours.

What can I do?

Let me know if you need more details.

lnd version: v0.11.1-beta 
Lightning Terminal version: v0.3.0

Roasbeef:
What state is the account actually in?

louneskmt:
> What state is the account actually in?

It is in `PENDING_UPDATE` state

Roasbeef:
That state means it's waiting on an account update to confirm (deposit, withdrawal, etc) On start up, do you see either `poold` or `lnd` try to rebroadcast that transaction?

louneskmt:
Here are the lnd logs when I restart LiT (and thus poold):
```
lnd | 2020-11-09 20:26:15.289 [INF] NTFN: New block epoch subscription
lnd | 2020-11-09 20:26:18.526 [INF] NTFN: New block epoch subscription
lnd | 2020-11-09 20:26:21.821 [INF] NTFN: New block epoch subscription
lnd | 2020-11-09 20:26:22.564 [INF] NTFN: New confirmation subscription: conf_id=2, txid=3cb7bb036d22e7c7b9d8e5243590fea3b48da89de3cc4b8a37b3bb797fcb6cd0, num_confs=3 height_hint=656194
lnd | 2020-11-09 20:26:24.928 [INF] DISC: Broadcasting 80 new announcements in 8 sub batches
lnd | 2020-11-09 20:26:50.187 [INF] CRTR: Processed channels=0 updates=43 nodes=4 in last 59.999744164s
```
And LiT startup logs:
```
2020-11-09 21:26:14.030 [INF] RPCS: Starting trader server
2020-11-09 21:26:16.042 [INF] LOOP: Starting event loop at height 656194
[...]

PR #152, opened 2020-11-10
auctioneer: include input sig scripts for account deposits

wpaulino:
This ensures that the auctioneer can arrive at the correct transaction such that it can continue to track the state of the account on the chain.

Fixes https://github.com/lightninglabs/pool/issues/149.


Roasbeef (approved):
LGTM ✨

halseth (approved):
LGTM, good catch!

Issue #146, opened 2020-11-08
Enforce a minimum amount to withdrawals


cryptagoras:
I believe there were some complications when requesting a withdrawal with the default 0 sats amount.
Not sure about server-side on the pool but I don't see any minimum checks on the client-side.

Roasbeef:
> I believe there were some complications when requesting a withdrawal with the default 0 sats amount.

As in you ran into an issue yourself when attempting such a withdarawl request?

cryptagoras:
> > I believe there were some complications when requesting a withdrawal with the default 0 sats amount.
> 
> As in you ran into an issue yourself when attempting such a withdarawl request?

Yes, my account is stuck at `PENDING_UPDATE` for a day or so, a new `latest_txid` was generated ( f08b3532de09b58bad1b27244d746a6f6c5971376dcb22b9a19b0ba50b032b6a ) but I can't find the TX anywhere. I was waiting for Monday to come and check on IRC first before reporting it here.

wpaulino:
Do you have any logs at the time you performed the withdrawal that could point to the issue?

cryptagoras:
Sure.
```
2020-11-08 10:43:07.676 [INF] RPCS: Closing acct=02beb8f19d6361f30e78ae81a3d22a96fbfccbc7def1c8573c0c0af280919fff1d
2020-11-08 10:43:08.024 [ERR] AUCT: [/poolrpc.ChannelAuctioneer/OrderState]: rpc error: code = Unknown desc = no order found
2020-11-08 10:43:08.024 [ERR] RPCS: [/poolrpc.Trader/CloseAccount]: rpc error: code = Unknown desc = no order found
2020-11-08 10:44:52.756 [INF] RPCS: Withdrawing from acct=02beb8f19d6361f30e78ae81a3d22a96fbfccbc7def1c8573c0c0af280919fff1d
2020-11-08 10:44:53.266 [ERR] RPCS: [/poolrpc.Trader/WithdrawAccount]: rpc error: code = Unknown desc = unmatched backend error: -26: dust (code 64)
2020-11-08 10:45:15.137 [INF] RPCS: Withdrawing from acct=02beb8f19d6361f30e78ae81a3d22a96fbfccbc7def1c8573c0c0af280919fff1d
2020-11-08 10:45:15.137 [ERR] RPCS: [/poolrpc.Trader/WithdrawAccount]: account must be in StateOpen to bemodified
```
[...]

PR #156, opened 2020-11-11
account: request signature before updating account due to modification

wpaulino:
This ensures that a trader cannot progress their account state without the server acknowledging their request. In the event where the server acknowledges the request and responds, and the trader shuts down before receiving the response, manual intervention will be required by the auctioneer to remove the pending modification. This serves as a stopgap to address the recent account modification issues until a better solution is in place (re-request signature for pending modification).

Motivated by https://github.com/lightninglabs/pool/issues/146 and https://github.com/lightninglabs/pool/issues/149.

### DIFF
--- a/account/manager.go
+++ b/account/manager.go
@@ -1262,20 +1262,21 @@ func (m *Manager) spendAccount(ctx context.Context, account *Account,
 		}))
 	}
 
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM ☄️

PR #158, opened 2020-11-11
order: account for last match consuming all units left in reserved value computation

### LINKED ISSUE #159: Absurdly high available_balance in account
When I list my accounts to check my balance, I get an absurdly high value for `available_balance`.

```
{
	"accounts": [
		{
			...
			"value": 2046329,
			"available_balance": 18446744073706606460,
			...
		}
	]
}
```

At the moment I have one open order for more than the `value` above that's still unfulfilled, so that also seems wrong and I'm guessing is related. My available balance looks like it's just short of 2^64, so I figure it's doing `value - open_orders` and coming up with a negative number and putting that into an unsigned int field. 

As an aside, I found the name `value` unintuitive and had to read the proto file to understand what it represented, perhaps a name of `balance` would be clearer?

wpaulino:
Fixes #159

### DIFF
--- a/order/interface_test.go
+++ b/order/interface_test.go
@@ -149,6 +149,32 @@ func TestOrderReservedValue(t *testing.T) {
 			},
 		},
 		{
+			name: "ask 10 units 5 min units match",
+			order: &Ask{
+				Kit: Kit{
+					State:            StateSubmitted,
+					UnitsUnfulfilled: 10,
+					FixedRate:        10_000_000,
+					MaxBatchFeeRate:  1000,
+					LeaseDuration:    144,
+					MinUnitsMatch:    5,
+				},
+			},
+		},
+		{
+			name: "bid 10 units 5 min units match",
+			order: &Bid{
+				Kit: Kit{
+					State:            StateSubmitted,
+					UnitsUnfulfilled: 10,
+					FixedRate:        10_000_000,
+					MaxBatchFeeRate:  1000,
+					LeaseDuration:    144,
+					MinUnitsMatch:    5,
+				},
+			},
+		},
+		{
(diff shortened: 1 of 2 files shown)


order/interface_test.go
```diff
@@ -190,10 +190,22 @@ func TestOrderReservedValue(t *testing.T) {
...
 			// For bids the taker pays the most fees if the min
-			// units get matched every block.
+			// units get matched every block. There's an edge case
+			// where if the units unfulfilled is not divisible by
+			// the min units match, then the last match will consume
+			// all the remaining units left.
 			numBlocks := int(o.UnitsUnfulfilled / o.MinUnitsMatch)
-			amt := o.MinUnitsMatch.ToSatoshis()
+			unitsRem := o.UnitsUnfulfilled % o.MinUnitsMatch
+			lastMatch := o.MinUnitsMatch
```
halseth, line 199: Oh, I like this `lastMatch` approach 👍  Maybe easier to read than the `rem` approach I took?

order/interface_test.go
```diff
@@ -190,10 +190,22 @@ func TestOrderReservedValue(t *testing.T) {
 			}
 
 			// For bids the taker pays the most fees if the min
-			// units get matched every block.
+			// units get matched every block. There's an edge case
```
halseth, line 193: Looks like we have test cases for the remainder case (bid and ask of size 10 with min size 4). Could also add a case for the non-remainder case for a `MinUnitsMatch` that is not the base unit?

halseth (approved):
Simple addition to the test cases, otherwis LGTM 👍

Roasbeef (approved):
LGTM 🤖

Issue #48, opened 2020-08-11
RPC to query for multiple batches/snapshots


jamaljsr:
Currently, the `llm auction snapshot` RPC returns only the latest batch. In order to get a list of past batches for the last 24 hours, my app would need to make 144 RPC requests. 

I'd like to be able to make one RPC request to return a list of batches over a specific time period or number of blocks. It would also be nice to be able to do some sort of paging, so that my app can continue making additional requests to go further back in time. I do not have any strong opinions on how this is implemented (blocks vs timestamps vs start_index/limit). Whatever is easiest to implement is cool with me.

PR #165, opened 2020-11-19
Add RPC to query multiple batches

guggero:
Fixes https://github.com/lightninglabs/pool/issues/48.

We add a new RPC:
- `BatchSnapshots` allows to query multiple batches in one request, a start batch ID and number of batches must be given.


poolrpc/trader.proto
```diff
@@ -201,6 +201,33 @@ service Trader {
...
+            get: "/v1/pool/batch/snapshots/{start_batch_id}/{end_batch_id}"
+        };
+    };
+
+    /* pool: `auction history`
+    BatchHistory returns a list of all batches that were ever created. The list
+    of batch IDs is sorted in ascending order, meaning the oldest batch is the
+    first in the list, while the most recent batch is the last entry.
+    */
+    rpc BatchHistory (BatchHistoryRequest) returns (BatchHistoryResponse) {
```
halseth, line 226: Why are both these RPCs needed?  I was thinking something like `{end_batch, num}` would be enough to paginate backwards through the batch history avoid multiple round trips.

guggero: Hmm, yeah, makes sense. Not sure why I came up with the other pattern...

cmd/pool/auction.go
```diff
@@ -154,6 +156,71 @@ func batchSnapshot(ctx *cli.Context) error {
 	return nil
 }
 
+var batchSnapshotsCommand = cli.Command{
+	Name:      "snapshots",
```
halseth, line 160: nit: could also just use the reuse old CLI command with default args num=1?

order/batch.go
```diff
@@ -393,3 +403,42 @@ type BatchStorer interface {
...
 }
+
+// DecrementingBatchIDs lists all possible batch IDs that can exist between a
+// start and end key. Start key must be the larger/higher key, decrementing it a
+// given number of times should finally result in the end key. If the keys
+// aren't related or are too far apart, we return a maximum number of IDs that
+// corresponds to our safety net parameter and the end key won't be contained in
+// the list. The returned list is in descending order, meaning the first entry
+// is the start key, the last entry is the end key.
+func DecrementingBatchIDs(startKey, endKey *btcec.PublicKey) []BatchID {
```
halseth, line 414: not needed anymore?

halseth (approved):
Looks good 👍

cmd/pool/auction.go
guggero: Good idea, done!

order/batch.go
guggero: It is, but only in the Prometheus collector on the server side. I think it's still nice to have this on the client side though.

cmd/pool/auction.go
```diff
@@ -138,18 +151,27 @@ func batchSnapshot(ctx *cli.Context) error {
...
+				StartBatchId: batchID,
+				NumBatchesBack: uint32(ctx.Uint64(
+					"num_batches",
+				)),
+			},
+		)
+	} else {
+		resp, err = client.BatchSnapshot(
+			ctxb, &poolrpc.BatchSnapshotRequest{
+				BatchId: batchID,
```
wpaulino, line 166: Took a look at the server and it requires a positive `num_batches_back` so we need to specify it here in case it's been set to 1.

guggero: We call the singular RPC here which doesn't have that parameter. But perhaps we can just deprecate that one and only use the new, multi batch RPC.

wpaulino: Ah didn't realize that. I thought the plan was to deprecate that one eventually anyway?

wpaulino (approved):

guggero:
I switched over to only use the `BatchSnapshots` client RPC and added a deprecation note to the proto file.

PR #178, opened 2020-11-24
Add debug CLI and wrap errors

guggero:
Adds a debug CLI that helps inspecting a local trader database. Also wraps some errors to give more information about the cause.

EDIT: The advantage of these commands is that they work without starting the trader daemon (so no interaction with the server) and also work on databases of other users, as no interaction with lnd is required either.
Perhaps it's really just something we'll use and it doesn't need to be in the main code base. But I've hidden the command group so most people won't notice this.

### DIFF
--- a/auctioneer/batch.go
+++ b/auctioneer/batch.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/btcsuite/btcd/wire"
@@ -41,7 +42,7 @@ func (c *Client) checkPendingBatch() error {
 		return nil
 	}
 	if err != nil {
-		return err
+		return fmt.Errorf("loading pending batch failed: %v", err)
 	}
 
 	finalizedTx, err := c.finalizedBatchTx(id)
@@ -53,7 +54,7 @@ func (c *Client) checkPendingBatch() error {
 		return nil
 	}
 	if err != nil {
-		return err
+		return fmt.Errorf("querying finalized batch TX failed: %v", err)
 	}
 
 	if tx.TxHash() != finalizedTx.TxHash() {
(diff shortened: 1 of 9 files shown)


auctioneer/batch.go
```diff
@@ -67,15 +67,15 @@ func (c *Client) checkPendingBatch() error {
 // finalizedBatchTx retrieves the finalized transaction of a batch according to
 // the auctioneer, i.e., the transaction that will be broadcast to the network.
 func (c *Client) finalizedBatchTx(id order.BatchID) (*wire.MsgTx, error) {
-	req := &poolrpc.RelevantBatchRequest{Id: id[:]}
-	batch, err := c.client.RelevantBatchSnapshot(context.Background(), req)
+	req := &poolrpc.BatchSnapshotRequest{BatchId: id[:]}
```
halseth, line 70: Hm, should we just remove the `RelevantBatchRequest` RPC?

halseth (changes requested):
I think it is better to split off the debug command :)

guggero:
Split the actual bug fix off into https://github.com/lightninglabs/pool/pull/179 and removed the `RelevantBatchSnapshot` RPC there.

cmd/pool/debug.go
```diff
@@ -0,0 +1,246 @@
...
+
+var debugCommands = []cli.Command{
+	{
+		Name:      "debug",
+		ShortName: "d",
+		Usage:     "Debug the local trader state.",
+		Hidden:    true,
+		Subcommands: []cli.Command{
+			dumpAccountsCommand,
+			dumpOrdersCommand,
```
wpaulino, line 27: Any reason to have these given that we have `pool accounts list` and `pool orders list`?

wpaulino (approved):

cmd/pool/debug.go
guggero: I should have given some more explanation, you're right. I did this in a hurry to get a fix for the bug out quickly.
The advantage of these commands is that they work without starting the trader daemon (so no interaction with the server) and also work on databases of other users, as no interaction with `lnd` is required either.
Perhaps it's really just something we'll use and it doesn't need to be in the main code base. But I've hidden the command group so most people won't notice this.

halseth (approved):
Looks useful, LGTM :)

Issue #176, opened 2020-11-24
poold crashed and fails to start


ratpoison4:
The crash after many hours of normal work and waiting for an order to be matched:

```
[INF] RPCS: Received new block notification: height=658351
[INF] RPCS: Received new block notification: height=658352
[INF] AUCT: Server is shutting down, will reconnect in 5s
[INF] AUCT: Successfully connected to auction server
[ERR] RPCS: Error in server stream: EOF
[ERR] AUCT: Error in stream, trying to reconnect: EOF
panic: close of closed channel

goroutine 58 [running]:
github.com/lightninglabs/pool/auctioneer.(*Client).closeStream(0xc0005a4100, 0x0, 0x0)
        /src/github.com/lightninglabs/pool/auctioneer/client.go:223 +0x194
github.com/lightninglabs/pool/auctioneer.(*Client).HandleServerShutdown(0xc0005a4100, 0x13a1ca0, 0xc000192050, 0xc00044e330, 0x1)
        /src/github.com/lightninglabs/pool/auctioneer/client.go:1035 +0xe5
github.com/lightninglabs/pool.(*rpcServer).serverHandler(0xc0005c02d0, 0xc0005a38c0, 0xc000160840)
        /src/github.com/lightninglabs/pool/rpcserver.go:321 +0x48e
created by github.com/lightninglabs/pool.(*rpcServer).Start
        /src/github.com/lightninglabs/pool/rpcserver.go:211 +0x9cf
```

Trying to start `poold` again fail:

```
$ poold
[INF] LLMD: Version: 0.3.3-alpha commit=
[INF] LNDC: Creating lnd connection to localhost:10009
[INF] LNDC: Connected to lnd
[INF] LNDC: lnd version: v0.11.1-beta, build tags 'autopilotrpc,signrpc,walletrpc,chainrpc,invoicesrpc,watchtowerrpc'
[INF] LNDC: Using network mainnet
[INF] LNDC: Waiting for lnd to be fully synced to its chain backend, this might take a while
[INF] LNDC: lnd is now fully synced to its chain backend
[INF] LLMD: Auction server address: pool.lightning.finance:12010
[INF] CLDB: Checking for schema update: latest_version=1, db_version=1
[INF] RPCS: Starting trader server
[INF] RPCS: Connected to lnd node [redacted] with pubkey [redacted]
[INF] ACCT: Watching spend of [redacted]:0 for account [redacted]
[INF] AUCT: Successfully connected to auction server
[...]

PR #179, opened 2020-11-24
Fix startup error because of wrong RPC

guggero:
This PR fixes https://github.com/lightninglabs/pool/issues/176 where several users reported the trader daemon crashing on startup.

Removed the unused `RelevantBatchSnaphost` RPC as well. @wpaulino wasn't this used for something else too? Did we remove that already? I thought this was vital for something else as well.

### DIFF
--- a/auctioneer/batch.go
+++ b/auctioneer/batch.go
@@ -66,14 +66,14 @@ func (c *Client) checkPendingBatch() error {
 // finalizedBatchTx retrieves the finalized transaction of a batch according to
 // the auctioneer, i.e., the transaction that will be broadcast to the network.
 func (c *Client) finalizedBatchTx(id order.BatchID) (*wire.MsgTx, error) {
-	req := &poolrpc.RelevantBatchRequest{Id: id[:]}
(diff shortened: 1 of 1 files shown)


poolrpc/auctioneer.proto
```diff
@@ -997,52 +996,6 @@ message TermsResponse {
     uint64 next_batch_clear_timestamp = 7;
 }
 
-message RelevantBatchRequest {
```
halseth, line 1000: I think it is okay removing this (which technically is a breaking change?) since it didn't work anyway.

halseth (approved):
Nice fix 👍

jamaljsr:
tACK 👍

wpaulino (approved):

poolrpc/auctioneer.proto
Roasbeef: FWIW, this'll end up breaking a portion of the server once we update it eventually. However, we'll need to still keep this in place for older clients, so I think we need to keep it after all until we no longer see any queries for the call.

Roasbeef (changes requested):

PR #181, opened 2020-11-24
log+watcher: avoid log of manual context canceled error

wpaulino:


guggero (approved):
Yesss, thank you for the fix :100:

halseth (approved):

PR #184, opened 2020-11-30
Clean up pending batch artifacts on startup

guggero:
Depends on #187, only the last 5 commits are new.

If a trader shuts down after being involved in a batch that resulted in a pending channel but is never finalized because other traders caused the batch to fail, they never get the chance to clean up those pending channels because the information necessary for that is only kept in memory.

We now also try to clean up pending channels and funding shims on startup if we detect an old, pending batch.
The big refactor in #187 was mainly necessary to do this in a clean way (concerning what component is responsible for what tasks).

For users with existing pending channels prior to this fix (they will never go away), we could add a manual `cleanup` command that:
* Gets all pending channels from `lnd` that have a thaw height
* Ask the auctioneer for all known batch transactions (easy once the cached batch query is in place)
* Abandon all pending channels that have a funding TXID that never made it into a batch.

But I think that should be done in an additional PR.

### DIFF
--- a/auctioneer/batch.go
+++ b/auctioneer/batch.go
@@ -25,7 +25,10 @@ type BatchSource interface {
 	// PendingBatchSnapshot retrieves the snapshot of the currently pending
 	// batch. If there isn't one, account.ErrNoPendingBatch is returned.
 	PendingBatchSnapshot() (*clientdb.LocalBatchSnapshot, error)
+}
 
+// BatchCleaner abstracts the cleaning up of a trader's pending batch.
+type BatchCleaner interface {
 	// DeletePendingBatch removes all references to the current pending
 	// batch without applying its staged updates to accounts and orders. If
 	// no pending batch exists, this acts as a no-op.
@@ -58,7 +61,7 @@ func (c *Client) checkPendingBatch() error {
 	}
 
 	if snapshot.BatchTX.TxHash() != finalizedTx.TxHash() {
-		return c.cfg.BatchSource.DeletePendingBatch()
+		return c.cfg.BatchCleaner.DeletePendingBatch()
 	}
 
 	return nil
--- a/auctioneer/batch.go
+++ b/auctioneer/batch.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/pool/account"
-	"github.com/lightninglabs/pool/order"
+	"github.com/lightninglabs/pool/clientdb"
 	"github.com/lightninglabs/pool/poolrpc"
 )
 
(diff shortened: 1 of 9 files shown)


rpcserver.go
```diff
@@ -269,6 +269,15 @@ func (s *rpcServer) Stop() error {
 	}
 
 	rpcLog.Info("Trader server stopping")
+
+	// TODO(guggero): Make sure this cannot lead to loss of channel if the
+	// trader shuts down after sending the sign message but before it can
+	// receive the finalize message.
+	if err := s.removePendingBatchArtifacts(); err != nil {
```
wpaulino, line 276: So we don't always want to do this here as we could shut down after signing and before receiving the finalize message and end up deleting our channels (assuming the batch finalizes without further changes). Instead, this should go in `checkPendingBatch` where the other batch recovery logic lives after detecting a different batch was finalized (different batch transaction hash).

halseth:
Super old PR that started to tackle this: https://github.com/lightningnetwork/lnd/pull/1755

auctioneer/batch.go
```diff
@@ -22,9 +22,11 @@ var (
 
 // BatchSource abstracts the source of a trader's pending batch.
 type BatchSource interface {
-	// PendingBatch retrieves the ID and transaction of the current pending
-	// batch. If one does not exist, account.ErrNoPendingBatch is returned.
-	PendingBatch() (order.BatchID, *wire.MsgTx, error)
+	// PendingBatchSnapshot retrieves the snapshot and raw transaction of
+	// the currently pending batch. If there isn't one,
+	// account.ErrNoPendingBatch is returned.
+	PendingBatchSnapshot() (*clientdb.LocalBatchSnapshot, *wire.MsgTx,
```
wpaulino, line 29: The batch transaction should already be in the snapshot so we can remove it from the return parameter.

server.go
```diff
@@ -67,6 +68,7 @@ type Server struct {
 
 	cfg             *Config
 	db              *clientdb.DB
+	fundingManager  *funding.Manager
```
wpaulino, line 71: Was there a reason for it (and the other managers) living in the RPC server instead of here similar to how it's done in lnd?

wpaulino (commented):
Looks pretty good!

auctioneer/batch.go
guggero: Right, done. Was also able to remove the pending batch TX key that was used for this.

server.go
guggero: I'm not sure. I think it just grew that way organically because only the RPC server needed access to them.

clientdb/batch_snapshot.go
```diff
@@ -308,6 +308,24 @@ func storePendingBatchSnapshot(tx *bbolt.Tx,
 	return topBucket.Put(batchSnapshotPendingKey, buf.Bytes())
 }
 
+// fetchPendingBatchSnapshot retrieves the currently pending batch snapshot from
+// the database or returns the account.ErrNoPendingBatch error if none exists.
+func fetchPendingBatchSnapshot(tx *bbolt.Tx) (*LocalBatchSnapshot,
```
halseth, line 313: nit: do we need the linebreak here?

halseth (approved):
LGTM ✅

wpaulino (approved):

clientdb/batch_snapshot.go
guggero: Nope, fixed :+1:

PR #185, opened 2020-12-02
account+rpc: allow expiration update for existing accounts

### LINKED ISSUE #8: accounts: renew option
We should give users an option to renew an account that's about to expire, or has already expired.

### LINKED ISSUE #39: Create account expiry field is vague
When creating an account using ordered flags (`llm accounts new [command options] amt expiry`), it's not clear if `expiry` is `expiry_height` or `expiry_blocks`. After testing, I see that it is `expiry_height`, but that's not immediately obvious from the help displayed. 

Minor nit, I think it would be better to have `expiry_blocks` as the ordered flag. I'm more likely to remember that 1008 blocks is around one week and multiply from there pretty easily in my head. With `expiry_height`, I have to first go find the current block height, then do that same math to add onto it. I'm lazy, so help me save a few seconds.

### LINKED ISSUE #160: accounts: default value for expiration
We should have a default value for `expiry_blocks` so users don't need to decide on an expiration date. With account renewal the expiration date becomes less important. 

```
$ pool accounts new --amt=50000000
Balance: 50,000,000 sats
Expiration: 4320 blocks (~30 days)
To change the expiration pass either expiry_blocks or expiry_height
Create account? (Y/n): 
{
   ...
}
```

Default value: `12960 (~90 days)`

wpaulino:
This PR allows traders to update their account expiration after it has been created through the RPC/CLI interface. It also serves as groundwork for when we eventually allow automatic account renewals through batch participation. As a result of this functionality, a default of 30 days has been set going forward for all new accounts created, and can be overwritten if necessary.

Fixes #8.
Fixes #39.
Fixes #160.


account/manager.go
```diff
@@ -1070,6 +1068,59 @@ func (m *Manager) WithdrawAccount(ctx context.Context,
...
 
+// UpdateAccountExpiry updates the expiration of an open/expired account. This
+// will always require a signature from the auctioneer, even after the account
+// has expired, to ensure the auctioneer is aware the account is being renewed.
+func (m *Manager) UpdateAccountExpiry(ctx context.Context,
+	traderKey *btcec.PublicKey, newExpiry uint32,
+	feeRate chainfee.SatPerKWeight, bestHeight uint32) (*Account,
+	*wire.MsgTx, error) {
+
+	// The account can only be modified in `StateOpen`.
```
halseth, line 1079: comment is wrong

cmd/pool/account.go
```diff
@@ -86,7 +86,9 @@ var newAccountCommand = cli.Command{
 		cli.Uint64Flag{
 			Name: accountExpiryRelative,
 			Usage: "the relative height (from the current chain " +
-				"height) that the account should expire at",
+				"height) that the account should expire at " +
+				"(default of 30 days equivalent in blocks)",
+			Value: defaultExpiryRelative,
```
halseth, line 95: where is this defined?

cmd/pool/account.go
```diff
@@ -62,6 +63,9 @@ const (
 
 	accountExpiryRelative = "expiry_blocks"
 
+	// Equivalent to approx. 30d in blocks.
+	defaultExpiryRelative = 30 * 144
```
halseth, line 67: ok, should probably be with previous commit

cmd/pool/account.go
```diff
@@ -247,6 +250,106 @@ func listAccounts(ctx *cli.Context) error {
 	return nil
 }
 
+var updateExpiryCommand = cli.Command{
+	Name:      "expire",
```
halseth, line 254: `expire` sounds like it will close the account. Should all it `renew` instead?

halseth (commented):
Cool, this will just work with the existing `ModifyAccount` call on the `auctioneer`?

Not many comments, great PR 👍

poolrpc/trader.proto
```diff
@@ -73,6 +73,17 @@ service Trader {
         };
     };
 
+    /* pool: `accounts expiry`
+    UpdateAccountExpiry modifies the expiration of an account. The expiration
+    can be decreased or increased.
+    */
+    rpc UpdateAccountExpiry (UpdateAccountExpiryRequest) returns (UpdateAccountExpiryResponse) {
```
Roasbeef, line 80: `RenewAccount`?

Roasbeef: I guess we could only expose this model on the CLI as an alternative.

cmd/pool/account.go
Roasbeef: +1 for `renew`

Roasbeef (changes requested):

account/manager.go
```diff
@@ -1200,20 +1251,17 @@ func (m *Manager) spendAccount(ctx context.Context, account *Account,
...
-		// This will require a new account expiration.
-		if !isClose {
-			return nil, nil, errors.New("modifications for expired " +
-				"accounts are not currently supported")
-		}
-
+	switch {
+	// Only force the expiry path once the account has expired and the
+	// account is being closed.
+	case witnessType == expiryWitness && isClose:
```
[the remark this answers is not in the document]
wpaulino: The account is already expired at this point. This is to ensure any account modifications after the account's been expired always go through the auctioneer so that it can learn about the changes being made.

account/manager.go
```diff
@@ -927,21 +973,34 @@ func (m *Manager) handleAccountExpiry(traderKey *btcec.PublicKey,
...
 	// If the account has already been closed or is in the process of doing
 	// so, there's no need to mark it as expired.
-	if account.State == StatePendingClosed || account.State == StateClosed {
+	case StatePendingClosed, StateClosed:
 		return nil
+
+	// If the account is waiting for a confirmation, use the expired state
+	// indicating so, unless its expiration was extended within the update.
+	case StatePendingUpdate, StatePendingBatch:
+		if account.Expiry > height {
```
Roasbeef, line 986: Ah since, it'll confirm and resume state transitions afterwards, makes sense.

Roasbeef (commented):
Latest version looking pretty good, my comment re allowing while pending batch still stands

account/manager.go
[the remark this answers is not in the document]
halseth: shouldn't `expiryWitness && !isClose`  be an invalid combination?

poolrpc/trader.proto
```diff
@@ -370,6 +380,29 @@ message DepositAccountResponse {
...
+message RenewAccountRequest {
+    // The key associated with the account to renew.
+    bytes account_key = 1;
+
+    oneof account_expiry {
+        // The new absolute expiration height of the account.
+        uint32 absolute_expiry = 2;
+
+        // The new relative expiration height of the account.
+        uint32 relative_expiry = 3;
```
halseth, line 392: I think it's enough to only support absolute heght.

halseth: Hm, or maybe it's useful for the client to not have to learn the current block height?

halseth (changes requested):

account/manager.go
```diff
@@ -1070,6 +1068,59 @@ func (m *Manager) WithdrawAccount(ctx context.Context,
...
+	traderKey *btcec.PublicKey, newExpiry uint32,
+	feeRate chainfee.SatPerKWeight, bestHeight uint32) (*Account,
+	*wire.MsgTx, error) {
+
+	// The account can only be modified in `StateOpen`.
+	account, err := m.cfg.Store.Account(traderKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	if account.State != StateOpen && account.State != StateExpired {
```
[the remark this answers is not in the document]
wpaulino: Done.

account/manager.go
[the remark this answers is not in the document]
wpaulino: Not necessarily, the multi-sig spend path can still be used after the expiration as long as the auctioneer allows it, which it will only do so in the event of a renewal.

account/manager.go
```diff
@@ -934,7 +934,9 @@ func (m *Manager) handleAccountSpend(traderKey *btcec.PublicKey,
 }
 
 // handleAccountExpiry marks an account as expired within the database.
-func (m *Manager) handleAccountExpiry(traderKey *btcec.PublicKey) error {
+func (m *Manager) handleAccountExpiry(traderKey *btcec.PublicKey,
+	height uint32) error {
```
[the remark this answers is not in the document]
wpaulino: It is in a later commit, this is just here to make the commit compile.

account/manager.go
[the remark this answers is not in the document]
wpaulino: I left it here as it only applies to these specific cases. An account won't have a different expiry than the one being notified in the other states.

poolrpc/trader.proto
wpaulino: Correct. This is already the case when opening new accounts as well. Communication between the trader and auctioneer use absolute block heights.

Roasbeef (approved):
LGTM 🍀

account/manager.go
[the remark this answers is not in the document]
halseth: Yeah, but if you use `expiryWitness` with `isClose == false` it won't go through the auctioneer, and the client is trying to not close the account? I don't understand

halseth (changes requested):

account/manager.go
```diff
@@ -1273,6 +1333,9 @@ func (m *Manager) spendAccount(ctx context.Context, account *Account,
 
 	case multiSigWitness:
 		spendPkg, err = m.createSpendTx(ctx, account, inputs, outputs, 0)
+
+	default:
+		err = fmt.Errorf("unhandled witness type: %v", witnessType)
```
halseth, line 1338: switch much more clear now 👍

account/manager.go
```diff
@@ -1713,24 +1776,22 @@ coinSelection:
...
-	newAccountOutput := &wire.TxOut{
-		Value:    int64(newAccountValue),
-		PkScript: newPkScript,
-	}
 	modifiers := []Modifier{
 		ValueModifier(newAccountValue),
 		IncrementBatchKey(),
 	}
+	if newAccountExpiry != nil {
+		modifiers = append(modifiers, ExpiryModifier(*newAccountExpiry))
```
halseth, line 1788: could also add this directly as a variadic argument to the method sigature

halseth (approved):
Great work, LGTM ✅

guggero:
Replaced by #199.

PR #187, opened 2020-12-07
refactor: move funding manager into own package, make it clean up pending batches

guggero:
To implement @wpaulino's suggestion in #184 (https://github.com/lightninglabs/pool/pull/184#discussion_r532975596) in a clean way, we first need to do some refactoring.

This PR moves the funding manager from the main package into its own `funding` package.
The pending channel cleanup methods that were implemented for the in-memory pending batches only are then refactored to be more generic and moved into the `funding` package.
This will allow us to also clean up pending batches from local database snapshots on daemon startup (which will be implemented in #184).

### DIFF
--- a/fundingmgr.go
+++ b/fundingmgr.go
@@ -90,11 +90,9 @@ type fundingMgr struct {
 	// pendingOpenChannels is a channel through which we'll receive
 	// notifications for pending open channels resulting from a successful
 	// batch.
-	pendingOpenChannels <-chan *lnrpc.ChannelEventUpdate_PendingOpenChannel
+	pendingOpenChannels chan *lnrpc.ChannelEventUpdate_PendingOpenChannel
 
 	batchStepTimeout time.Duration
-
-	quit chan struct{}
 }
 
 // deriveFundingShim generates the proper funding shim that should be used by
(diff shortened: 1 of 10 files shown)


order/batch.go
```diff
@@ -298,9 +298,12 @@ func AbandonCanceledChannels(matchedOrders map[Nonce][]*MatchedOrder,
 				},
 			)
 			if err != nil {
-				log.Warnf("Unable to abandon channel "+
-					"(channel_point=%v:%d) for order=%v",
-					txHash, idx, ourOrderNonce)
+				log.Infof("Cleaning up incomplete/replaced " +
```
wpaulino: :+1:

wpaulino: Consider debug instead?

fundingmgr.go
```diff
@@ -87,14 +87,7 @@ type fundingMgr struct {
 	// node doesn't already have channels with.
 	newNodesOnly bool
 
-	// pendingOpenChannels is a channel through which we'll receive
-	// notifications for pending open channels resulting from a successful
-	// batch.
-	pendingOpenChannels <-chan *lnrpc.ChannelEventUpdate_PendingOpenChannel
```
wpaulino, line 93: Is there a reason this can't continue to live within the funding manager since that's where we open channels?

order/batch.go
halseth: Should we check the error message here somehow? To know that we got an expected error, not a critical one

halseth (approved):
👍

order/batch.go
guggero: Wasn't sure if it would ever be helpful in case do run into problem and this fails for an actual reason. But I guess we'll ask people to run with `debuglevel=debug` in case of a problem anyway, so I changed this to `Debugf`.

fundingmgr.go
guggero: I wasn't sure how exactly to do it, this is mostly needed for #184. But I think it does make sense to make the funding manager the owner over the update chan :+1:

order/batch.go
halseth: we could string match (or something better?) on the expected error before logging or exiting in case of a critical one.

wpaulino (approved):

PR #190, opened 2020-12-09
Added how to cancel orders.

BLdotN:
I kind of had a hard time finding how to cancel an order. When i finally found out, i thought it would be nice to add it here :)

### DIFF
--- a/docs/orders.md
+++ b/docs/orders.md
@@ -280,6 +280,14 @@ To reduce the value of this worst-case estimation, there are two possible flags 
   value too low might result in orders not being executed at all or only when
 
   the mempools are almost empty.
+  
+## Cancel orders
+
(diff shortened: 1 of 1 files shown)


docs/orders.md
```diff
@@ -280,6 +280,14 @@ To reduce the value of this worst-case estimation, there are two possible flags
   value too low might result in orders not being executed at all or only when
 
   the mempools are almost empty.
+  
+## Cancel orders
```
Roasbeef, line 284: "Cancelling Orders" instead?

BLdotN: Sure! Tnx for the feedback.

Roasbeef (approved):
LGTM 💈

PR #193, opened 2020-12-16
Add batch creation timestamp, remove order max duration, rename batch version fields

guggero:
Split off from #142 to allow storing and returning the batch creation timestamp on the server side.

**This should be in the same release as #142 for all serialization assumptions to work!** Merely splitting this off to make the review of the server part easier by splitting into smaller PRs.

We don't store the batch confirmation timestamp yet because that requires us to look into updating a batch snapshot state and invalidating the server caches.


halseth (approved):
Easy change, LGTM! So we should not merge this til #142 since the timestamp won't be set?

guggero:
Yes, ideally, we just merge #142 (which contains these commits as well and is reviewed already) once the server PRs are close to being merged.

wpaulino (approved):

guggero:
All commits are also contained in #142. Since both are approved now, I'm closing this first part so we can merge #142 in one piece.

PR #197, opened 2021-01-04
account: use latest TX from DB instead of querying lnd

guggero:
To make the account manager less dependent on the wallet state of lnd,
we use the latest TX state that we now store for the account instead of
the lnd ListTransactions RPC where possible.
This fixes account recovery for lnd nodes that were restored from the
seed and don't remember/detect all Pool transactions automatically.

This should fix the error `2020-12-24 14:46:33.574 [ERR] RPCS: error storing recovered account: account funding TX with output 002033ab6debxxxx not found` that was reported to me on Slack.

### DIFF
--- a/account/manager.go
+++ b/account/manager.go
@@ -431,16 +431,20 @@ func (m *Manager) resumeAccount(ctx context.Context, account *Account, // nolint
 		return fmt.Errorf("unable to construct account output: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


wpaulino (approved):

Roasbeef (approved):
LGTM 🐊

PR #199, opened 2021-01-07
Fix lnd v0.12.0-beta compatibility, combine #142 and #185

### LINKED ISSUE #8: accounts: renew option

### LINKED ISSUE #39: Create account expiry field is vague

### LINKED ISSUE #160: accounts: default value for expiration

guggero:
This PR fixes compatibility with `lnd v0.12.0-beta.rc3` and later.

It also contains the full commit list of #142 and #185. As soon as the server is deployed to mainnet, this PR can be merged and the other two can be closed.

Fixes #8.
Fixes #39.
Fixes #160.


wpaulino (approved):

rpcserver.go
```diff
@@ -1586,7 +1586,16 @@ func (s *rpcServer) sendSignBatch(batch *order.Batch, sigs order.BatchSignature,
 		switch chanInfo.Version {
 		case chanbackup.TweaklessCommitVersion:
 			channelType = poolrpc.ChannelType_TWEAKLESS
-		case chanbackup.AnchorsCommitVersion:
+
+		// The AnchorsCommitVersion was never widely deployed (at least
```
halseth, line 1590: 👍

halseth (approved):

PR #201, opened 2021-01-14
auctioneerrpc: move auctioneer proto files into own sub-module

guggero:
This PR prepares the server side proto files to be owned by the server repo by splitting them off into their own `auctioneerrpc` package that is then declared as a sub-module.

To remove a proto based import, we also extract the REST annotations into its own file to streamline the client proto somewhat.


wpaulino (approved):

auctioneerrpc/auctioneer.proto
```diff
@@ -1,8 +1,8 @@
 syntax = "proto3";
 
-package poolrpc;
+package auctioneerrpc;
```
halseth, line 3: is this a breaking change?

guggero: Argh, yes it is... I think we'll need to serve everything under both paths for a while. Same proto, different gRPC package. Good catch, totally forgot about the package being in the URI.

poolrpc/gen_protos.sh
```diff
@@ -6,19 +6,17 @@ set -e
...
 
-done
-
-# Only generate the REST proxy and definitions for the client component.
-protoc -I/usr/local/include -I. \
-       -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
-       --grpc-gateway_out=logtostderr=true,paths=source_relative:. \
-       trader.proto
+  # Generate the REST reverse proxy.
+  protoc -I/usr/local/include -I. \
```
Roasbeef: Odd indentation here, is this intended to be a part of the for loop? If so should be indented to make it a bit easier to read imo.

auctioneerrpc/auctioneer.proto
Roasbeef: Breaking change from an RPC PoV since this module name is used in the path to the actual HTTP endpoint? 

What do we need to do in order to serve both paths on the server end of things?

go.mod
```diff
@@ -15,15 +15,17 @@ require (
...
 	github.com/urfave/cli v1.20.0
 	go.etcd.io/bbolt v1.3.5-0.20200615073812-232d8fc87f50
 	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
-	google.golang.org/genproto v0.0.0-20200513103714-09dca8ec2884
 	google.golang.org/grpc v1.29.1
 	gopkg.in/macaroon-bakery.v2 v2.0.1
 	gopkg.in/macaroon.v2 v2.1.0
 )
+
+replace github.com/lightninglabs/pool/auctioneerrpc => ./auctioneerrpc
```
Roasbeef, line 31: 👍

Roasbeef (changes requested):
Looks pretty good, excited to have this land so we can pick up the pace of dev on teh server side a bit. One comment about confusing indentation, and another about what we need to do on the server end if we're about to break the entire HTTP path with this change.

Roasbeef:
Do we need to modify the code that auto-generates the current set of docs due to this?

auctioneerrpc/auctioneer.proto
guggero: Okay, I figured that we can just go ahead and move the _golang_ package and use it under the `auctioneerrpc` package name but keep the RPC name as it was. That way this is non breaking as on the RPC level nothing changes.
I confirmed that this doesn't break old clients.

poolrpc/gen_protos.sh
guggero: Was a mix of spaces and tabs, fixed.

guggero:
I was able to make it non breaking by just keeping the old RPC name but changing the golang name.

> Do we need to modify the code that auto-generates the current set of docs due to this?

Yes, probably. I'll look into it after this is merged :+1:

auctioneerrpc/auctioneer.proto
```diff
@@ -1,8 +1,12 @@
 syntax = "proto3";
 
+// We can't rename this to auctioneerrpc, otherwise it would be a breaking
```
halseth, line 3: 👍

auctioneerrpc/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/lightninglabs/pool/auctioneerrpc
```
halseth, line 1: Commit message could include an explanation for why this change is needed.

halseth (approved):
LGTM 👍

PR #210, opened 2021-01-26
Fix multiple REST issues, allow only one macaroon to be used for lnd connection

### LINKED ISSUE #196: Don't require all macaroons
Analogous to https://github.com/lightninglabs/loop/issues/299

Please provide an option to supply just one macaroon instead of the whole directory.

### LINKED ISSUE #204: REST Api call /v1/pool/batch/snapshots should work without parameters
The /v1/pool/batch/snapshots has two parameters, start_batch_id and num_batches_back. Without the parameters, the call doesn't work:
`/v1/pool/batch/snapshots` -> `Not Found`
`/v1/pool/batch/snapshots/` -> `Not Found`
`/v1/pool/batch/snapshots//` -> JSON formatted error: `type mismatch, parameter: num_batches_back`
`/v1/pool/batch/snapshots//1` -> Returns last batch snapshot.

This is a non-standard URL pattern. A more standard pattern would be to have `num_batches_back` as a query parameter and have the path parameter  `start_batch_id` optional.

`/v1/pool/batch/snapshots` -> return array with last batch
`/v1/pool/batch/snapshots/` -> return array with last batch
`/v1/pool/batch/snapshots?num_batches_back=5` -> return array with last 5 batches
`/v1/pool/batch/snapshots/deadbeef` -> return array with batch deadbeef
`/v1/pool/batch/snapshots/deadbeef?num_batches_back=5` -> return array with 5 batches starting from batch deadbeef

https://github.com/lightninglabs/pool/blob/94db020f7380a52feee7ddbfdbc37b3a7a52a8dc/poolrpc/trader.proto#L227

### LINKED ISSUE #208: REST API  /v1/pool/orders state field when ORDER_SUBMITTED is missing
The state field for newly submitted orders is missing. Through the pool cli, these are shown as ORDER_SUBMITTED.

Secondary: the default for the rest call is active_only = false, for the pool clli, the default is true.

### LINKED ISSUE #209: Expose current block height through RPC and REST
Current block height is not exposed. This is useful information for automatically renewing an account.

It is shown in the RPCS logging, with `Received new block notification` message.

guggero:
Fixes #208.
Fixes #209.
Fixes #196.
Fixes #204.

Contains a number of smaller fixes that should help developers and users alike.


cmd/pool/main.go
```diff
@@ -126,14 +127,37 @@ func main() {
...
+	app.Commands = append(app.Commands, stopDaemonCommand)
 
 	err := app.Run(os.Args)
 	if err != nil {
 		fatal(err)
 	}
 }
 
+var stopDaemonCommand = cli.Command{
+	Name:  "stopdaemon",
```
halseth, line 141: just `stop` ?

config.go
```diff
@@ -75,9 +80,20 @@ var (
...
 
 type LndConfig struct {
-	Host        string `long:"host" description:"lnd instance rpc address"`
-	MacaroonDir string `long:"macaroondir" description:"Path to the directory containing all the required lnd macaroons"`
-	TLSPath     string `long:"tlspath" description:"Path to lnd tls certificate"`
+	Host string `long:"host" description:"lnd instance rpc address"`
+
+	// MacaroonDir is the directory that contains all the macaroon files
+	// required for the remote connection.
+	MacaroonDir string `long:"macaroondir" description:"DEPRECATED: Use macaroonpath. The directory containing all lnd macaroons to use for the remote connection."`
```
halseth, line 87: maybe enough with "DEPRECATED: Use macaroonpath."

config.go
```diff
@@ -239,6 +255,31 @@ func Validate(cfg *Config) error {
 		return err
 	}
 
+	// Make sure only one of the macaroon options is used.
+	switch {
+	case cfg.Lnd.MacaroonPath != "" && cfg.Lnd.MacaroonDir != "":
+		return fmt.Errorf("use either --lnd.macaroondir or --lnd." +
```
halseth, line 261: since `macaroondir` is deprecated, should say `use macaroonpath`

server.go
```diff
@@ -127,8 +128,9 @@ func (s *Server) Start() error {
 	// TODO(roasbeef): more granular macaroons, can ask user to make just
 	// what we need
 	s.lndClient, err = lndclient.NewBasicClient(
-		s.cfg.Lnd.Host, s.cfg.Lnd.TLSPath, s.cfg.Lnd.MacaroonDir,
-		s.cfg.Network,
+		s.cfg.Lnd.Host, s.cfg.Lnd.TLSPath,
+		path.Dir(s.cfg.Lnd.MacaroonPath), s.cfg.Network,
+		lndclient.MacFilename(defaultLndMacaroon),
```
halseth, line 133: does this work if admin macaroon is not used?

halseth (changes requested):

halseth:
Great PR and additions :=

server.go
guggero: Good catch! It would indeed not have worked. Fixed it by using `path.Base()` instead.

poolrpc/trader.swagger.json
```diff
@@ -424,6 +424,88 @@
...
+            "type": "integer",
+            "format": "int64"
+          }
+        ],
+        "tags": [
+          "Trader"
+        ]
+      }
+    },
+    "/v1/pool/batch/snapshots/{start_batch_id}": {
```
mutatrum, line 450: Does this work with `/v1/pool/batch/snapshots/` as well, without `start_batch_id`? It states the parameter is required (line 473), but with or without trailing slash should work the same IMO.

mutatrum:
Nice additions!

Two more things for getinfo, but these might be out of scope for this:

- poold version
- auctioneer connection state

poolrpc/trader.swagger.json
guggero: Yes, that's what line 409 is for. There are three URL patterns now:
```
/v1/pool/batch/snapshots
/v1/pool/batch/snapshots/{start_batch_id}
/v1/pool/batch/snapshots/{start_batch_id}/{num_batches_back}
```

And you can use the GET parameters too:
```
/v1/pool/batch/snapshots?num_batches_back=10
/v1/pool/batch/snapshots/{start_batch_id}?num_batches_back=10
```

guggero:
Thanks for the review, @mutatrum!

> Two more things for getinfo, but these might be out of scope for this:

Great ideas, added both of them.

order/manager.go
```diff
@@ -308,6 +309,16 @@ func (m *Manager) BatchFinalize(batchID BatchID) error {
 	return nil
 }
 
+// OurNodePubkey returns our lnd node's public identity key or an error if the
+// manager wasn't fully started yet.
+func (m *Manager) OurNodePubkey() ([33]byte, error) {
+	if m.ourNodeInfo == nil {
```
halseth, line 315: is this racy?

auctioneer/client.go
```diff
@@ -781,6 +781,12 @@ func (c *Client) wait(backoff time.Duration) error {
 	}
 }
 
+// IsSubscribed returns true if at least one account is in an active state and
+// the subscription stream to the server was established successfully.
+func (c *Client) IsSubscribed() bool {
+	return c.serverStream != nil
```
halseth, line 787: need mutex?

server.go
```diff
@@ -127,8 +128,9 @@ func (s *Server) Start() error {
 	// TODO(roasbeef): more granular macaroons, can ask user to make just
 	// what we need
 	s.lndClient, err = lndclient.NewBasicClient(
-		s.cfg.Lnd.Host, s.cfg.Lnd.TLSPath, s.cfg.Lnd.MacaroonDir,
-		s.cfg.Network,
+		s.cfg.Lnd.Host, s.cfg.Lnd.TLSPath,
+		path.Dir(s.cfg.Lnd.MacaroonPath), s.cfg.Network,
```
halseth, line 132: probably a `lndclient` comment really, but shouldn't we really be passing in the full macaroon path here also, and remove the dir?

halseth (changes requested):
Two comments about potential race conditions, otherwise LGTM 👍

poolrpc/trader.swagger.json
mutatrum: My comment was on `/v1/pool/batch/snapshots/`, with a trailing slash and without start_batch_id query param. It looks like that's not included in the pattern. But it's a minor thing.

order/manager.go
guggero: Yes, good point. Added an atomic `isStarted` flag to the manager.

auctioneer/client.go
guggero: Ah yes. There even was a mutex already. Fixed another potentially unguarded access by narrowing the lock: 8a308dd8

server.go
guggero: True. We should try to get rid of the base client anyway. I put this on my TODO list.

order/manager.go
```diff
@@ -83,25 +86,28 @@ func NewManager(cfg *ManagerConfig) *Manager {
 
 // Start starts all concurrent tasks the manager is responsible for.
 func (m *Manager) Start() error {
+	if atomic.LoadUint32(&m.isStarted) == 1 {
```
halseth, line 89: isn't this already handled by the below `sync.Once`?

order/manager.go
```diff
@@ -308,6 +316,16 @@ func (m *Manager) BatchFinalize(batchID BatchID) error {
 	return nil
 }
 
+// OurNodePubkey returns our lnd node's public identity key or an error if the
+// manager wasn't fully started yet.
+func (m *Manager) OurNodePubkey() ([33]byte, error) {
+	if atomic.LoadUint32(&m.isStarted) != 1 {
```
halseth, line 322: Hm, okay I guess it's okay using it for this purpose 😛

poolrpc/trader.proto
```diff
@@ -7,6 +7,17 @@ package poolrpc;
 option go_package = "github.com/lightninglabs/pool/poolrpc";
 
 service Trader {
+    /* pool: `getinfo`
+    GetInfo returns general information about the state of the Pool trader
+    daemon.
+    */
+    rpc GetInfo (GetInfoRequest) returns (GetInfoResponse);
+
+    /* pool: `stopdaemon`
```
halseth, line 16: nit: stop

halseth (approved):

server.go
```diff
@@ -195,11 +195,23 @@ func (s *Server) Start() error {
...
 		}
 
+		// The default JSON marshaler of the REST proxy only sets
+		// OrigName to true, which instructs it to use the same field
+		// names as specified in the proto file and not switch to camel
+		// case. What we also want is that the marshaler prints all
+		// values, even if they are falsey.
+		customMarshalerOption := proxy.WithMarshalerOption(
+			proxy.MIMEWildcard, &proxy.JSONPb{
+				OrigName:     true,
```
Roasbeef, line 207: 👍

auctioneer/client.go
```diff
@@ -492,10 +492,14 @@ func (c *Client) connectAndAuthenticate(ctx context.Context,
 		return sub, true, nil
 	}
 
+	// Guard the read access only. We can't use defer for the unlock because
+	// both the connectServerStream and SendAuctionMessage need to hold the
+	// mutex as well.
 	c.streamMutex.Lock()
-	defer c.streamMutex.Unlock()
+	needToConnect := c.serverStream == nil
+	c.streamMutex.Unlock()
```
Roasbeef, line 500: Don't we also need the lock for `subscribedAccts` below?

server.go
Roasbeef: The `lndclient.MacFilename` functinoal arg was initially meant for this purpose...but seems that knowledge wasn't properly passed down to those that modify the library most frequently.

Roasbeef (approved):
Solid set of changes! Quite a few issue will be closed by this PR which is great, ofc. 

One smol question re ensuring proper mutex usage, but we should be good after that's cleared up. 

LGTM 🍄

auctioneer/client.go
guggero: Yes, added an extra mutex.

PR #211, opened 2021-01-27
Add user agent and initiator string to account open and order submission calls

guggero:
With the upcoming Pool UI we want to be able to see very rough statistics on what software is used for opening accounts and submitting orders. This will currently only distinguish between the standalone Pool daemon (`poold`) and the LiT integrated daemon (`litd`) for the static user agent and then add the initiator `pool-cli` for calls issued with the `pool` binary.


version.go
```diff
@@ -52,13 +54,11 @@ func Version() string {
...
-// which are not valid according to the semantic versioning guidelines for
-// pre-release version and build metadata strings.  In particular they MUST
-// only contain characters in semanticAlphabet.
-func normalizeVerString(str string) string {
+// which are not valid according to the given alphabet.
+func normalizeVerString(str, alphabet string) string {
 	var result bytes.Buffer
 	for _, r := range str {
-		if strings.ContainsRune(semanticAlphabet, r) {
+		if strings.ContainsRune(alphabet, r) {
```
halseth, line 147: isn't it better to just return an error if this is not true?

halseth (approved):
Cool approach 😀

version.go
guggero: Perhaps, yes. But that'd require us to thread through the error in a few places and bug the user with an error for something that's just a nice to have (in case of the initiator). Or do you mean for the semantic version as well?

halseth: Was thinking just in general. If an invalid user agent or semantic is set, we will fail to start. But perhaps I'm missing some case this would be disruptive for.

auctioneerrpc/auctioneer.proto
```diff
@@ -116,6 +125,15 @@ message ServerSubmitOrderRequest {
...
     }
+
+    // The user agent string that identifies the software running on the user's
+    // side. This can be changed in the user's client software but it _SHOULD_
+    // conform to the following pattern:
+    //    Agent-Name/semver-version(/additional-info)
+    // Examples:
+    //    poold/v0.4.2-beta/commit=3b635821,initiator=pool-cli
+    //    litd/v0.4.0-alpha/commit=326d754,initiator=lit-ui
+    string user_agent = 3;
```
carlaKC, line 136: Also note here that it's limited to 255?

carlaKC (approved):
LGTM 🌮

version.go
guggero: Okay, I added a check for the version and agent name on startup and also when overwriting the agent name.
I think we really don't care enough about the initiator being silently truncated for an error to be threaded through so I didn't change that behavior.

guggero:
Added a version bump commit as well so we can include a tagged version in LiT.

version.go
```diff
@@ -30,19 +32,26 @@ const (
 	appPreRelease = "alpha"
 )
 
+func init() {
```
halseth, line 62: hehe, probably not worth adding an init just for this case 🙈

version.go
halseth: I take it back, better to do what you had initially :)

version.go
```diff
@@ -7,58 +7,173 @@ package pool
...
+// agentName stores the name of the software that is added as the first part of
+// the user agent string. This defaults to the value "poold" when being run as
+// a standalone component but can be overwritten by LiT for example when poold
+// is integrated into the UI.
+var agentName = defaultAgentName
+
+// SetAgentName sets the static part of the user agent string, identifying the
+// type of binary that is running. This function panics if the agent name
+// contains characters outside of the allowed semantic alphabet.
+func SetAgentName(newAgentName string) {
```
halseth, line 56: not used?

halseth (changes requested):

version.go
guggero: Okay, I reverted the change with the `init()` function but left the check in for when the agent name is overwritten externally.

version.go
guggero: Yes, it'll be used in LiT to overwrite the default name. Updated the comment to make it clearer.

halseth (approved):

PR #217, opened 2021-02-04
sidecar: add ticket with TLV encoders

guggero:
This is a first preview of the sidecar ticket type and its TLV and string encoders.

> Ticket is a struct holding all the information for establishing/buying a
> sidecar channel. It is meant to be used in a PSBT like manner where each
> participant incrementally adds their information according to their role and
> the current step.

The ticket will replace all new parameters/flags described in https://github.com/guggero/pool/blob/sidecar/docs/sidecar_channels.md.
Instead the ticket will be passed between the participants and enriched with more information in each step.

### DIFF
--- a/sidecar/codec.go
+++ b/sidecar/codec.go
@@ -1,0 +1,96 @@
+package sidecar
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+)
+
+const (
+	// checksumLen is the number of bytes we add as a checksum. We'll take
+	// this many bytes out of the full SHA256 hash and add it to the string
+	// encoded version of the ticket as an additional integrity check.
+	checksumLen = 4
+)
+
+var (
+	// defaultStringEncoding is the default encoding algorithm we use to
+	// encode the gzipped binary data as a string.
+	defaultStringEncoding = base64.URLEncoding
+
+	// sidecarPrefix is a specially prepared prefix that will spell out
+	// "sidecar" when encoded as base64. It will cause all our sidecar
+	// tickets to be easily recognizable with a human readable part.
+	sidecarPrefix = []byte{0xb2, 0x27, 0x5e, 0x71, 0xaa, 0xc0}
+)
+
+// EncodeToString serializes and encodes the ticket as an URL safe string that
+// contains a human readable prefix and a checksum.
+func EncodeToString(t *Ticket) (string, error) {
+	var buf bytes.Buffer
+
+	// Add a prefix to make the ticket easily recognizable by human eyes.
+	if _, err := buf.Write(sidecarPrefix); err != nil {
+		return "", err
+	}
+
+	// Serialize the raw ticket itself.
+	if err := SerializeTicket(&buf, t); err != nil {
+		return "", err
+	}
+
+	// Create a checksum (SHA256) of all the bytes so far (prefix + raw
+	// ticket) and add 4 bytes of that to the string encoded version. This
+	// will prevent the ticket from appearing valid even if parts of it were
+	// not copy/pasted correctly.
+	hash := sha256.New()
+	_, _ = hash.Write(buf.Bytes())
+	checksum := hash.Sum(nil)[0:4]
+	if _, err := buf.Write(checksum); err != nil {
+		return "", err
+	}
+
+	return defaultStringEncoding.EncodeToString(buf.Bytes()), nil
+}
+
+// DecodeString decodes and then deserializes the given sidecar ticket string.
+func DecodeString(s string) (*Ticket, error) {
+	rawBytes, err := defaultStringEncoding.DecodeString(s)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(rawBytes) < len(sidecarPrefix)+checksumLen {
+		return nil, fmt.Errorf("not a sidecar ticket, invalid length")
+	}
+
+	totalLength := len(rawBytes)
+	prefixLength := len(sidecarPrefix)
+	payloadLength := totalLength - prefixLength - checksumLen
+
+	prefix := rawBytes[0:prefixLength]
+	payload := rawBytes[prefixLength : prefixLength+payloadLength]
+	checksum := rawBytes[totalLength-checksumLen : totalLength]
+
+	// Make sure the prefix matches our static prefix.
+	if !bytes.Equal(prefix, sidecarPrefix) {
+		return nil, fmt.Errorf("not a sicecar ticket, invalid prefix "+
+			"%x", prefix)
+	}
+
+	// Calculate the SHA256 sum of the prefix and payload and compare it to
+	// the checksum we also expect to be in the full string serialized
+	// ticket.
+	hash := sha256.New()
+	_, _ = hash.Write(prefix)
+	_, _ = hash.Write(payload)
+	calculatedChecksum := hash.Sum(nil)[0:4]
+	if !bytes.Equal(checksum, calculatedChecksum) {
+		return nil, fmt.Errorf("invalid sidecar ticket, checksum " +
+			"mismatch")
+	}
+
+	// Everything's fine, let's decode the actual payload.
+	return DeserializeTicket(bytes.NewReader(payload))
+}
--- a/sidecar/codec_test.go
+++ b/sidecar/codec_test.go
@@ -1,0 +1,68 @@
+package sidecar
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	hardcodedTicket = "sidecarAAQgHBgUEAwIBAAIBYwMBAgp5CwgAAAAAAAADCQwIAA" +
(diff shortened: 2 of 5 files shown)


sidecar/interface.go
```diff
@@ -0,0 +1,230 @@
...
+	// StateCanceled is the state a ticket is in after the sidecar channel
+	// bid order was canceled by the taker.
+	StateCanceled State = 6
+)
+
+// Offer is a struct holding the information that a sidecar channel provider is
+// committing to when offering to buy a channel for the recipient. The sidecar
+// channel flow is initiated by the provider creating a ticket and adding its
+// signed offer to it.
+type Offer struct {
```
Roasbeef, line 65: Likely don't need in this version, but I imagine we'll also eventually have the following fields:
  * pubkey of the node that the channel may be opened to 
  * type of channel

sidecar/interface.go
```diff
@@ -0,0 +1,230 @@
...
+	PendingChannelID [32]byte
+}
+
+// Ticket is a struct holding all the information for establishing/buying a
+// sidecar channel. It is meant to be used in a PSBT like manner where each
+// participant incrementally adds their information according to their role and
+// the current step.
+type Ticket struct {
+	// ID is a pseudorandom identifier of the ticket.
+	ID [8]byte
```
Roasbeef, line 128: Any reason 8 and not 32?

sidecar/tlv.go
```diff
@@ -0,0 +1,377 @@
...
+		))
+	}
+	if ticket.Execution != nil {
+		records = append(records, tlv.MakeStaticRecord(
+			executionType, &ticket.Execution, executionLength,
+			EExecution, DExecution,
+		))
+	}
+
+	tlvStream, err := tlv.NewStream(records...)
```
Roasbeef, line 73: I thought the idea was to use _two_ logical streams to allow their contents to be signed independently?

Roasbeef:
Pondering if we should add a static prefix for the encoded string versions just so it's a bit more identifiable for human users...

sidecar/tlv.go
```diff
@@ -0,0 +1,377 @@
...
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcutil"
+	"github.com/lightninglabs/pool/order"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+const (
+	idType        tlv.Type = 1
```
halseth, line 14: any given reason for using odd numbers here? Since it is a custom type.

sidecar/tlv.go
```diff
@@ -0,0 +1,377 @@
...
+		*vPtr = v
+
+		return tlv.DBytes32(r, &v.PendingChannelID, buf, 32)
+	}
+	return tlv.NewTypeForEncodingErr(val, "*Execution")
+}
+
+// EPubKeyN is a nil safe Encoder for the btcec.PublicKey type. An error is
+// returned if val is not a **btcec.PublicKey.
+func EPubKeyN(w io.Writer, val interface{}, buf *[8]byte) error {
```
halseth, line 254: kinda feels like a lot of boiler plate here. Shouldn't most of the logic we need (writing bytes to a TLV stream) already be defined somewhere it can be reused?

sidecar/codec.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"bytes"
+	"encoding/base64"
+)
+
+const (
+	// defaultEncodeBufSize is the default buffer size we use when encoding
+	// a ticket.
+	defaultEncodeBufSize = 512
```
halseth, line 11: Is this really needed?

halseth (commented):
Interesting approach!

sidecar/tlv.go
guggero: Discussed this offline. We did mention using two streams initially but during implementation I noticed that the pattern of having a distinct `Digest()` method and then embedding the signature inside the TLV stream itself makes things easier. Mostly for also storing a ticket in the DB as we can just serialize the whole stream and put it into `etcd` or `bbolt`.

sidecar/interface.go
guggero: I'll keep that in mind. With the extensibility of the TLV stream we should be able to easily add those in a later version.
It got me thinking though... Should the individual fields of the sub structs (`Offer`, `Receiver` and so on) also be encoded as TLV records? If I understood the TLV library correctly you can have records of a stream missing and can add new records. But within one record (for example looking at `DOffer`), all the fields that are read with `tlv.DUint64` still need to be present or you run into an EOF.

Could there be an issue with encoding TLV records inside TLV records or should I go ahead and refactor the sub structs to also use records instead of just fields?

sidecar/interface.go
guggero: Mostly to save some space. This doesn't need to be cryptographically strong IMO, just some unique identifier to tell them apart at the UI level.

sidecar/tlv.go
guggero: No, I just copied the pattern I found in https://github.com/lightningnetwork/lnd/blob/master/channeldb/invoices.go#L129.
I probably need to read up on the implication of even/odd types. Or does this only apply to the LN gossip wire protocol?

sidecar/tlv.go
guggero: There are pre-defined functions for many things available, yes. But there wasn't one for conditionally serializing a pubkey depending on whether it was nil or not. But I think we should instead use individual TLV records everywhere then the nilness problem would go away.

sidecar/codec.go
guggero: No, left over from a previous design. Removed.

guggero:
> Pondering if we should add a static prefix for the encoded string versions just so it's a bit more identifiable for human users...

Great idea. I did some base64 magic and added a prefix that will spell out "sidecar" when encoded as base64 :grin: 

The other idea I implemented in a first draft was to put everything through gzip to reduce the size. But I then discarded it because that adds another library dependency for anyone trying to implement this in another language.
And I think it doesn't really matter how long the string is. Or do we care about putting it into a tweet for example?

guggero:
I also added a checksum to the string encoded version as an additional integrity check.

sidecar/tlv.go
guggero: Ah, I think I get it now. If a type is _unknown_ and is odd, the tlv library will just ignore it. But since we know all our types here, we don't need to use odd types. Fixed.

sidecar/interface.go
guggero: I went ahead and also used tlv records for the sub structs.
The nice side effect of this is that this also makes it easier to parse the ticket in other languages (for example in JS for a "light" implementation of the sidecar receiver client) as the tlv parsing can just be done recursively.

sidecar/tlv.go
```diff
@@ -0,0 +1,386 @@
...
+		tlv.MakePrimitiveRecord(pushAmtType, &pushAmt),
+		tlv.MakePrimitiveRecord(signPubKeyType, &o.SignPubKey),
+		tlv.MakeStaticRecord(
+			sigOfferDigestType, &o.SigOfferDigest, 64, ESig, DSig,
+		),
+	); err != nil {
+		return o, err
+	}
+
+	o.Capacity = btcutil.Amount(capacity)
```
halseth, line 217: probably doesn't matter, but check type was parsed before setting it for consistency?

halseth: never mind, later I see we decode directly into the structs, which is kinda nice way to shorten the code.

sidecar/codec_test.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"math/big"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	hardcodedTicket = "sidecarAAQgHBgUEAwIBAAIBYwMBAgp5CwgAAAAAAAADCQwIAA" +
```
halseth, line 12: 😀

halseth (approved):
This iteration looks very good!

Need to see the bigger picture, but the serilazation code now became less verbose and easy to follow IMO. LGTM ✅

sidecar/interface.go
Roasbeef: Hmm, so then in this case the issuer is the one responsible for uniqueness?

Roasbeef: I guess in the end, the nonce of the order is the thing that really binds it all.

sidecar/codec_test.go
Roasbeef: base64 is kinda ugly imo, but hey, it encodes lol

Roasbeef: also we can also change it later if really needed

Roasbeef (approved):
LGTM 🚡

sidecar/codec_test.go
guggero: Yeah, it's also not optimal because of the special characters (that's why I chose URL safe encoding here). Which encoding did you have in mind? bech32? I chose base64 for two reasons: A bit shorter than bech32 (double the alphabet size) and because PSBT uses the same (trying to convey the idea that this is similar in its concept as a transport format).

PR #222, opened 2021-02-22
sidecar channels 2/3: Offer, register and submit sidecar orders

guggero:
Depends on #232, only last 10 commits are new.

This is the first part of what's needed to execute sidecar orders.
There are 4 steps in the sidecar protocol, this PR covers the first 3 (see `doc/sidecar_channels.md` for more information):
1. Provider: Create and sign an offer for a sidecar channel (`pool sidecar offer`)
2. Recipient: Register an offer and add node information (`pool sidecar register`)
3. Provider: Submit a bid order with the sidecar ticket (`pool orders submit bid ... --sidecar_ticket ...`)

The 4th and last step is the execution of the sidecar order itself and will be added in a separate PR.


funding/manager.go
```diff
@@ -653,6 +668,116 @@ func (m *Manager) RemovePendingBatchArtifacts(
...
+// EncodeOrderedTicket signs the order digest of a sidecar ticket and returns it
+// in the string encoded version.
+func (m *Manager) EncodeOrderedTicket(ctx context.Context, t *sidecar.Ticket,
+	nonce order.Nonce) (string, error) {
+
+	// Since this information will just be returned to the user and printed
+	// we can just return an empty string if there is no ticket for an
+	// order or a ticket is in an invalid state.
+	if t == nil || t.State != sidecar.StateRegistered {
+		return "", nil
```
halseth, line 749: return error instead?

clientdb/order.go
```diff
@@ -317,6 +341,24 @@ func storeOrderMinUnitsMatchTX(rootBucket *bbolt.Bucket, nonce order.Nonce,
 	return orderBucket.Put(orderMinUnitsMatchKey, buf.Bytes())
 }
 
+// storeOrderTlvTX saves the order's additional information as a tlv stream
+// within the order's root bucket.
+func storeOrderTlvTX(rootBucket *bbolt.Bucket, nonce order.Nonce,
```
halseth, line 358: don't have to pass nonce as separate param

clientdb/order.go
```diff
@@ -621,11 +633,73 @@ func DeserializeOrder(nonce order.Nonce, r io.Reader) (
...
+		return err
+	}
+
+	// Now check what records were actually parsed from the stream and
+	// assign any parsed fields to our order.
+	switch castOrder := o.(type) {
+	case *order.Ask:
+
+	case *order.Bid:
+		if t, ok := parsedTypes[sidecarTicketType]; ok && t == nil {
```
halseth, line 661: Maybe I misread the documentation here, but do we need to check `t == nil`? Seems like `t != nil && ok` would be an error.

order/manager.go
```diff
@@ -326,6 +366,78 @@ func (m *Manager) OurNodePubkey() ([33]byte, error) {
 	return m.ourNodeInfo.IdentityPubkey, nil
 }
 
+// validateTicketForOrder makes sure that the given sidecar ticket is in the
+// correct state for being used in a bid order and has all the necessary
+// information set. We also check that our node initially offered to lease this
+// channel by checking the embedded signature.
+func (m *Manager) validateTicketForOrder(ctx context.Context, t *sidecar.Ticket,
```
halseth, line 373: wasn't a similar method created for the recipient, could we reuse the code?

cmd/pool/order.go
```diff
@@ -257,7 +263,14 @@ func ordersSubmitAsk(ctx *cli.Context) error { // nolint: dupl
...
 	ask := &poolrpc.Ask{
 		LeaseDurationBlocks: uint32(ctx.Uint64("lease_duration_blocks")),
-		Version:             uint32(order.VersionLeaseDurationBuckets),
+		Version:             uint32(order.VersionSidecarChannel),
+	}
+
+	// Because there is no disadvantage of being matched against bids with
+	// push amounts (other than perhaps a bit more complicated accounting
+	// formulas), we make them opt-out.
+	if ctx.Bool("disallow_push_amt") {
```
halseth, line 272: hm, is there really any compelling reason that we know of to make this optional?

halseth (commented):
Did a high-level pass, looks good IMO!

funding/manager.go
Roasbeef: Hmm, depends on the intended usage of this method...

funding/manager.go
```diff
@@ -653,6 +668,116 @@ func (m *Manager) RemovePendingBatchArtifacts(
...
+		return nil, fmt.Errorf("error storing sidecar ticket: %v", err)
+	}
+
+	return ticket, nil
+}
+
+// EncodeOrderedTicket signs the order digest of a sidecar ticket and returns it
+// in the string encoded version.
+func (m *Manager) EncodeOrderedTicket(ctx context.Context, t *sidecar.Ticket,
+	nonce order.Nonce) (string, error) {
```
Roasbeef, line 743: Isn't the nonce already encoded in the ticket itself?

Roasbeef: Also this is the _bid_ nonce right?

poolrpc/trader.proto
```diff
@@ -966,3 +987,45 @@ message StopDaemonRequest {
...
+    */
+    string ticket = 1;
+}
+
+message RegisterSidecarRequest {
+    /*
+    The sidecar ticket to register and add the node and channel funding
+    information to. The ticket must be in the state "offered".
+    */
+    string ticket = 1;
```
Roasbeef, line 1038: Should this be using `SidecarTicket` itself vs the plain string version of it, as we might add some new fields over time.

clientdb/order.go
```diff
@@ -42,11 +42,16 @@ var (
...
+
+	// orderTlvKey is a key within the order bucket for additional, tlv
+	// encoded data.
+	orderTlvKey = []byte("order-tlv")
 )
 
 type extraOrderData struct {
 	minNodeTier   order.NodeTier
 	minUnitsMatch order.SupplyUnit
+	tlvData       []byte
```
Roasbeef, line 66: Ah very cool, so now in the future we'll map any new fields added to the orders on the proto level to his new field?

Roasbeef: Also I don't think a migration is necessary, given that we're using a new key here vs modifying the existing values of an order.

clientdb/order.go
Roasbeef: So if the type is found and it's known, it'll be in the map but with a nil value (`ok` check is important here). If it's found and the type is _unknown_ then it'll be in the map, but the value will be the raw bytes of the type.

order/interface.go
```diff
@@ -43,6 +44,16 @@ const (
 	// to use lease durations outside of the default/legacy 2016 block
 	// duration.
 	VersionLeaseDurationBuckets Version = 2
+
+	// VersionSidecarChannel is the order version that added sidecar
```
Roasbeef, line 48: Hmm, so is ths idea that the bid order itself will contain a sidecar ticket rather than open encoding some of the fields? As an example, a `push_amt` field could've been added to the bid. However, I think this is fine as if someone wants to create a balanced channel (so they're the receiver and offerer of the sidecar channel), we can have most of the interaction be hidden behind the RPC API. 

This brings to light that we should also allow that self balanced channel use case to be accessible over the RPC API via just the regular `SubmitOrder(bid)` call. The only changes we'd need here would be do specify the push amt manually (we should call it something else, like `self_chan_balance`) or something, and then update the acceptance logic to ensure that the push amt is set within the channel itself. This might either require a channel acceptor hook, or this information might already be available (or can be added) within the pending channels call itself.

auctioneerrpc/auctioneer.proto
```diff
@@ -887,10 +887,18 @@ message ServerBid {
...
     */
     NodeTier min_node_tier = 5;
+
+    /*
+    If this bid order is meant to lease a channel for another node (which is
+    dubbed a "sidecar channel") then this ticket contains all information
+    required for setting up that sidecar channel. The ticket is expected to be
+    the raw bytes of the ticket tlv stream without any prefix or checksum.
+    */
+    bytes sidecar_ticket = 6;
```
Roasbeef, line 901: See my comment re also allowing some of this information to be specified w/o the raw encoding of the ticket. Alternatively, we can make a help er function to construct the ticket, but with the same information of the node creating it as both the offerer and the acceptor.

order/manager.go
```diff
@@ -142,30 +145,72 @@ func (m *Manager) PrepareOrder(ctx context.Context, order Order,
...
-		return nil, fmt.Errorf("unable to parse node uris: %v", err)
+	params := &ServerOrderParams{}
+	bid, isBid := order.(*Bid)
+	isSidecar := isBid && bid.SidecarTicket != nil
+
+	// Using a sidecar ticket with a bid order means we won't be receiving
+	// the channel ourselves and are instead leasing a channel for another
+	// node. Therefore we need to add the other node's identity and multisig
+	// public key in the bid order we send to the auctioneer.
+	if isSidecar {
```
Roasbeef, line 156: See comment re the use case of people wanting balanced chanenls from the getgo, w/o no third party involved. We've seen many people on reddit and those telegram groups are doing a sort of "trust fall" exercise, which we can make more trust minimized by allowing them to use our existing execution engine for this purpose.

auctioneerrpc/auctioneer.proto
Roasbeef: Also it could be that we accept the sidecar ticket on the level of the client protos, but instead send out the individual fields between the client and the actual server.

cmd/pool/order.go
Roasbeef: IMO a maker shouldn't really care about this, given that with the way it ends up being executed, their funds aren't used for the push amount itself. However maybe there's soem routing related reason they wouldn't want to have a balanced channel in the first place? Not sure...

Roasbeef (changes requested):
Excellent work! Getting closer to finally realizing the long time dream of sidecar channels. 

The only major omission I see in this PR, is the inability for a taker to have the channel created to them be balanced by using some of their own funds int he process w/o having to go through the whole sidecar dance with themselves. Perhaps all we need as some helper methods to allow them to have `poold` itself handle the sidecar channel creation (as they play the role of both the offerer and the acceptor) .

cmd/pool/order.go
guggero: My concern here was mainly the accounting of the maker. Their Pool account balance wouldn't decrease by the expected amount (`capacity - premium`) and instead some remote channel balance would suddenly appear in their `lnd` balance. I'll look at this again in the execution PR. Maybe it's enough if we just correctly set the `push_amt` in the matched order so the maker knows what exactly went down. Going to remove the commit for now, easy to add again later on.

clientdb/sidecar.go
```diff
@@ -0,0 +1,139 @@
...
+	"go.etcd.io/bbolt"
+)
+
+var (
+	// ErrNoSidecar is the error returned if no sidecar with the given
+	// multisig pubkey exists in the store.
+	ErrNoSidecar = errors.New("no sidecar found")
+
+	// sidecarsBucketKey is a bucket that contains all sidecars that are
+	// currently pending or completed. This bucket is keyed by the multisig
```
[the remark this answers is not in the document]
guggero: Right, if for whatever reason a provider decides to not use random IDs then it could lead to collisions. They need to be unique on the asker's/provider's side. But if a recipient gets tickets from multiple providers, it _could_ create a collision. Added the provider's signing pubkey to the key as well.

funding/manager.go
guggero: I removed this method completely, wasn't necessary anymore after some refactoring.

funding/manager.go
guggero: Ended up solving this differently.

poolrpc/trader.proto
guggero: Wouldn't we end up adding those new fields to the ticket _itself_, as in the TLV encoded part? Or what new fields do you mean here?

order/interface.go
guggero: I ended up creating an independent PR for the `self_chan_balance`. At first I wasn't sure whether the feature was interesting outside of sidecar orders. Since it apparently clearly is, I added that PR first and created a 3 part split of the PRs.
The new user flow allows `self_chan_balance` to be used outside of sidecar channels (by just specifying the `--self_chan_balance` when submitting a bid order). When using a sidecar ticket the ticket itself will just dictate (=overwrite) that self channel balance when submitting a bid order.

sidecar_acceptor.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("error deriving multisig key: %v", err)
+	}
+
+	// We also need our node identity key in the parsed format.
+	nodeKey, err := btcec.ParsePubKey(nodePubKey[:], btcec.S256())
+	if err != nil {
+		return nil, fmt.Errorf("error parsing node pubkey: %v", err)
+	}
+
+	ticket.State = sidecar.StateRegistered
```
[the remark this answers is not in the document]
halseth: This is still not fixed?

order/interface.go
```diff
@@ -570,6 +589,25 @@ func (b *Bid) Digest() ([sha256.Size]byte, error) {
 			return result, err
 		}
 
+	case VersionSidecarChannel:
+		var buf bytes.Buffer
+		if b.SidecarTicket != nil {
+			err := sidecar.SerializeTicket(&buf, b.SidecarTicket)
```
halseth, line 595: nit: can also just write directly to `msg` if done at the end

halseth (approved):
Great PR! LGTM pending dependencies ✅

auctioneerrpc/auctioneer.proto
guggero: I've done this in the latest iteration :+1:

sidecar_acceptor.go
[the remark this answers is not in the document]
guggero: Ah yes, indeed. I fixed a similar issue in another place but forgot this one.

order/interface.go
guggero: Good idea, done.

cmd/pool/order.go
guggero: This is solved with the self chan balance now being additive. Therefore the maker's account balance doesn't change at all compared to orders without a push amount.

clientdb/sidecar.go
```diff
@@ -0,0 +1,173 @@
...
+	// currently pending or completed. This bucket is keyed by the ticket ID
+	// and offer signing pubkey of a sidecar.
+	sidecarsBucketKey = []byte("sidecars")
+)
+
+const (
+	// sidecarKeyLen is the length of a sidecar ticket's key. It is the
+	// length of the sidecar ID (8 bytes) plus the length of a compressed
+	// public key (33 bytes).
+	sidecarKeyLen = 8 + 33
```
Roasbeef, line 28: 👍 

Makes things a bit longer in practice, but nice to not have to worry about collisions. 

As an aside, I wonder if we should also add an _expiry_ to the tickets as well. Can be done easily later since the entire thing is a TLV record.

order/interface.go
```diff
@@ -49,6 +49,16 @@ const (
...
+
+	// VersionSidecarChannel is the order version that added sidecar
+	// channels for bid orders. Only orders with this version are allowed
+	// to set the sidecar ticket field on bid orders. Since sidecar orders
+	// also add the feature of push amounts on the leased channels, this
+	// affects makers as well. Makers that don't want to support leasing out
+	// channels with a push amount (because it might screw up their
+	// accounting or whatever) can opt out by explicitly submitting their
+	// ask orders with a version previous to this one.
+	VersionSidecarChannel Version = 4
```
Roasbeef, line 61: Non-blocking as we can change this before we actually deploy it, but should we consider just combining these versions? Or perhaps on the server-side there's a use for having the versions be distinct?

order/manager.go
```diff
@@ -142,30 +152,75 @@ func (m *Manager) PrepareOrder(ctx context.Context, order Order,
...
+		)
+		copy(
+			params.MultiSigKey[:],
+			ticket.Recipient.MultiSigPubKey.SerializeCompressed(),
+		)
+
+		// Most sidecar recipients won't be public nodes and therefore
+		// don't have connectable addresses.
+		//
+		// TODO(guggero): Add addresses to the Receiver part of the
```
Roasbeef, line 189: Yeah the only edge case here is if the maker on the other side is running _only_ over Tor. On the server-side if this becomes an issue, we can start to add information w.r.t preferred addresses for the acceptor. This way they can avoid nodes that are only Tor as that'll just result in failed execution.

order/manager.go
```diff
@@ -334,6 +384,59 @@ func (m *Manager) OurNodePubkey() ([33]byte, error) {
...
+			"information")
+	}
+
+	// Make sure the offer is valid and actually came from us.
+	o := t.Offer
+	if err := sidecar.VerifyOffer(ctx, t, m.cfg.Signer); err != nil {
+		return fmt.Errorf("error verifying sidecar offer: %v", err)
+	}
+
+	ourNodeKeyRaw, err := m.OurNodePubkey()
```
Roasbeef, line 413: I can see how this is useful on the ask side (similar to chan ann's we have two sigs for the on-chain output as well as the node itself), but what does this add from the PoV of the bidder here? Shouldn't this be done with the public key of the account submitting the bid? However in that scenario the order digest is already being signed which _includes_ the side car ticket itself.

Roasbeef: I guess it locks things down a bit, since it's possible for the node creating the offer to be different than the node creating the bid itself.

Roasbeef (approved):
LGTM 🕸

Only one lingering question re the authentication properties we're aimng for by also having the party submitting the bid also sign the sidecar ticket with its _node_ ID.

Roasbeef:
Needs a rebase!

guggero:
Rebased. But this should only be merged as a package, together with 3/3.

guggero:
Replaced by #242 that extracts all refactor and non-functional commits.

PR #224, opened 2021-02-24
Provide default value for lnd macaroon path, fix shutdown panic in LiT

guggero:
Fixes two issues I noticed when testing the latest version of Pool in combination with LiT.


halseth (approved):
Nice 👍

guggero:
I also added a version bump commit so we can use a tagged version in LiT. Going to merge as soon as Travis is green since the changes in the PR are more or less trivial.

PR #228, opened 2021-03-02
cmd/poold: pprof listen on localhost

Crypt-iQ:
Otherwise, debugging information is exposed if the port is reachable.

### DIFF
--- a/cmd/poold/main.go
+++ b/cmd/poold/main.go
@@ -77,7 +77,7 @@ func start() error {
 		}
 
 		go func() {
-			listenAddr := net.JoinHostPort("", config.Profile)
+			listenAddr := net.JoinHostPort("127.0.0.1", config.Profile)
 			profileRedirect := http.RedirectHandler("/debug/pprof",
 				http.StatusSeeOther)
 			http.Handle("/", profileRedirect)


cmd/poold/main.go
```diff
@@ -77,7 +77,7 @@ func start() error {
 		}
 
 		go func() {
-			listenAddr := net.JoinHostPort("", config.Profile)
+			listenAddr := net.JoinHostPort("127.0.0.1", config.Profile)
```
halseth, line 80: there are very valid use case for allowing profiling on a externally reachable port though.  Could we make the `config.Profile` option optionally set this IP?

halseth (changes requested):

guggero (changes requested):
:+1: for making this configurable. How about this: We change the `config.Profile` to a string and parse it. If it's just a number (e.g. to not break existing users that have the flag set), we use `127.0.0.1` as the host. If it's a `host:port` pair, use that directly.
Same for the other repos where this was proposed.

Misuwp (approved):

guggero:
Replaced by https://github.com/lightninglabs/pool/pull/391.

Issue #226, opened 2021-03-01
Fix account documentation


gked:
The [documentation](https://pool.lightning.engineering/accounts#closing-an-account), describing account closure is incorrect. The example omits required option  `--sat_per_vbyte.`
![image](https://user-images.githubusercontent.com/12704474/109450683-5e8fc080-7a00-11eb-8cf7-3e8df5307959.png)

Please update your example from
`*pool accounts close --trader_key=0288096be9917f8ebdfc6eb2701635fe658f4eae1e0274dcce41418b3fb5145732`

to

` pool accounts close --trader_key=0288096be9917f8ebdfc6eb2701635fe658f4eae1e0274dcce41418b3fb5145732 --sat_per_vbyte 0`

gked:
@guggero looks like you are open to get help on this. I can create a pull request.

Roasbeef:
Fixed by #229

PR #229, opened 2021-03-04
Updated for a account close example

gked:
Fixing  #226
The example showing how to close an account was incorrect. Fixed it by adding a required argument `--sat_per_vbyte`


docs/accounts.md
```diff
@@ -110,6 +110,5 @@ Incrementally _withdrawing_ from an account is also supported. The command is si
 Finally, if you wish to send _all_ your funds elsewhere, it's possible to close your account out before the main expiration period. We can close out the account we created above with the following command:
 
 ```text
-🏔 pool accounts close --trader_key=0288096be9917f8ebdfc6eb2701635fe658f4eae1e0274dcce41418b3fb5145732
+🏔 pool accounts close --trader_key=0288096be9917f8ebdfc6eb2701635fe658f4eae1e0274dcce41418b3fb5145732 --sat_per_vbyte 0
```
guggero, line 113: nit: maybe use another value than 0 as that will not work? Perhaps 10 or something like that (not too low as that wouldn't work with the current mempool).

guggero (changes requested):
Thanks for the PR!

Just need a different value than 0 then we're good to merge this.

docs/accounts.md
gked: sure thing. I can update to the value I used when closing my account few days ago.

halseth (changes requested):
Thanls for the PR! Can you squash the commits down to one? Otherwise this LGTM 🤠

guggero:
I'll "merge and squash" instead to get this in.

guggero (approved):

PR #232, opened 2021-03-09
sidecar channels 1/3: Allow self channel balance on bid orders

guggero:
During review of #222 it turned out that the push amount (or new: self channel balance) on its own is a useful feature to have.

To simplify the user flow for using the self channel balance for both normal bids as well as sidecar channels, we add this as a standalone feature first and let the sidecar ticket just pre-determine the self channel balance. Under the hood this feature will work independent of sidecar channels.

Reviewers: This PR contains some of the commits from #222. Sorry for the inconvenience, but I think this way it's easiest to follow. The PRs will also be split in 3 parts now:
 - 1/3: This PR: Implements just the self channel balance feature.
 - 2/3: #222: Adds submission of sidecar orders.
 - 3/3: #232: Adds execution of sidecar orders.


order/interface.go
```diff
@@ -43,6 +44,11 @@ const (
 	// to use lease durations outside of the default/legacy 2016 block
 	// duration.
 	VersionLeaseDurationBuckets Version = 2
+
+	// VersionSelfChanBalance is the order version that added use of the
+	// self channel balance field. Only orders with this version are allowed
+	// to use the self channel balance field.
+	VersionSelfChanBalance Version = 3
```
Roasbeef, line 51: Is the plan to keep this version for sidecars as well, or have that be another version so 4 in this case?

order/interface.go
```diff
@@ -568,6 +592,28 @@ func (b *Bid) ReservedValue(feeSchedule terms.FeeSchedule) btcutil.Amount {
...
+			"order version")
+	}
+
+	if err := sidecar.CheckOfferParams(
+		b.Amt, b.SelfChanBalance, BaseSupplyUnit,
+	); err != nil {
+		return fmt.Errorf("invalid self chan balance: %v", err)
+	}
+
+	if b.Units != b.MinUnitsMatch {
```
Roasbeef, line 610: 👍

order/interface.go
Roasbeef: Also comment reminder to make this the default elsewhere in the codebase.

clientdb/order.go
```diff
@@ -6,11 +6,23 @@ import (
...
 )
 
+const (
+	// selfChanBalanceType is the tlv type we use to store the self channel
+	// balance on bid orders.
+	//
+	// NOTE: When adding new tlv types for extra order data, make sure the
+	// types don't overlap between bid and ask orders! Always increase the
+	// type by one for each new one.
+	selfChanBalanceType tlv.Type = 1
```
Roasbeef, line 23: Re this comment, can we prefix the _name_ of the type with `ask` or `bid`, then be able to safely have both type streams share the same values if we need to?

cmd/pool/order.go
```diff
@@ -401,6 +406,12 @@ var ordersSubmitBidCommand = cli.Command{
 			Name:  "force",
 			Usage: "skip order placement confirmation",
 		},
+		cli.Uint64Flag{
+			Name: "self_chan_balance",
+			Usage: "give the channel leased by this bid order an " +
+				"initial balance by shifting additional funds " +
+				"from our account into the channel",
```
Roasbeef, line 413: Maybe worth mentioning this can be used to open 50/50 balanced channels? 

We'll also need to ensure that the method to ensure enough balance is reserved also adds this component as well.

order/manager_test.go
```diff
@@ -94,3 +95,171 @@ func TestValidateOrderAccountIsolation(t *testing.T) {
 		t.Fatalf("order A failed validation: %v", err)
 	}
 }
+
+// TestValidateOrder makes sure that the order manager properly validates
```
Roasbeef, line 99: 👍

order/batch_verifier_test.go
```diff
@@ -406,9 +417,10 @@ func TestBatchVerifier(t *testing.T) {
 				{
 					// balance - bid1Premium - bid2Premium -
 					// bid1ExecFee - bid2ExecFee - chainFees
+					// - bid2SelfChanBalance
 					// 500_000 - 1000 - 1000 -
-					// 1_110 - 1_110 - 186
-					Value:    495_594,
+					// 1_110 - 1_110 - 186 - 50
+					Value:    495_544,
```
Roasbeef, line 423: Btw...I dread whenever I need to change these fields 🤣

funding/manager.go
```diff
@@ -453,6 +457,9 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
 				NodePubkey:         matchedOrder.NodeKey[:],
 				LocalFundingAmount: chanAmt,
 				FundingShim:        fundingShim,
+				PushSat: int64(
+					matchedOrderBid.SelfChanBalance,
```
Roasbeef, line 475: 👍

funding/manager.go
```diff
@@ -453,6 +457,9 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
 				NodePubkey:         matchedOrder.NodeKey[:],
 				LocalFundingAmount: chanAmt,
 				FundingShim:        fundingShim,
+				PushSat: int64(
```
Roasbeef, line 474: Hmm, seems we should also change the way we communicate the channel size, no? As an example:
 * I want to buy a channel of 1 BTC inbound. 
 * I also want to commit 0.5 BTC of my own coins so I'm able to send and forward immediately using the channel. 
 * Therefore, I want the remote party to open a 1.5 BTC channel, as I'll be pushing over 0.5 BTC during execution into their account. 

AFAICT rn with this PR, we'll open a 1 BTC channel with 0.5 being pushed over, so same total amount, but I should only really pay the other party for that 0.5 BTC they commit.

funding/manager.go
```diff
@@ -104,14 +104,19 @@ type Manager struct {
 	// BatchStepTimeout is the timeout the manager uses when executing a
 	// single batch step.
 	BatchStepTimeout time.Duration
+
+	// NotifyShimCreated is a function that should be called whenever a
+	// funding shim is created for a bid order where we expect an incoming
+	// channel at any moment.
+	NotifyShimCreated func(ourBid *order.Bid, pendingChanID [32]byte)
```
Roasbeef, line 111: Alternatively, we could've extended the funding shim to also verify this information just as it does the pubkey/amt, etc. One other upcoming use case for this would be once we make channel types explicit in the protocol, and allow users to purchase precise channel types in Pool.

channel_acceptor.go
```diff
@@ -0,0 +1,134 @@
...
+		return &lndclient.AcceptorResponse{Accept: true}, nil
+	}
+
+	// The push amount in the acceptor request is in milli sats, we need to
+	// convert it first.
+	pushAmtSat := lnwire.MilliSatoshi(req.PushAmt).ToSatoshis()
+
+	// Push amount must be exactly what we expect. Otherwise the asker could
+	// be trying to cheat.
+	if expectedChanBid.SelfChanBalance != pushAmtSat {
```
Roasbeef, line 125: The upside of doing it here though as that we don't need to update `lnd` each time there's some additional constraint we want to assert.

Roasbeef (changes requested):
Excellent work! This change itself will add a lot of value to Pool as users can use it to fully bootstrap their node to the network given they can both purchase inbound and also establish outbound with nodes they want to have channels open with. However, depending on the scenario (and the node's tier, etc), the user may prefer to just submit an Ask and a Bid distinctly. 

The one thing that jumped out as me is if we should treat the`self_balance` as being additive, or implicitly included in the execution flow. See my comment in-line re the example case.

halseth (commented):
Did high level pass, nothing major. This looks like it could be a very handy feature! And implementation not bad at all :)

funding/manager.go
guggero: Ah, I hadn't considered this. But it totally makes sense that the self channel balance should be additive. I'm going to fix up the PR with that logic in mind.

cmd/pool/order.go
guggero: Added the text about 50/50 balanced to the description.
Because we add the self channel balance to the bid's `ReservedValue` method, it is properly included in all balance checks.

funding/manager.go
guggero: Yes, I think we should do this in 0.13 and then also add a bunch more of our RPCs we use to `lndclient` so we can also get rid of the `baseClient` variable we have in there. That way, after 0.13, we can do a "cleanup" PR for Pool and the server to simplify some things.

channel_acceptor.go
guggero: Yeah, it allows us to move forward with this now. Can always "upgrade" this once we have everything we need in the funding shim itself.

order/interface.go
guggero: The only place we currently actively set the value is in the CLI. I've updated the version there.

guggero:
I've updated this PR (and the server) to make the self channel balance _additive_. Ready for another round of review.

funding/manager.go
```diff
@@ -189,7 +193,7 @@ func (m *Manager) deriveFundingShim(ourOrder order.Order,
 	// shim, and register it so we use the proper funding key when we
 	// receive the marker's incoming funding request.
 	chanPointShim := &lnrpc.ChanPointShim{
-		Amt:       int64(chanSize),
+		Amt:       int64(chanSize + selfChanBalance),
```
Roasbeef, line 191: 👍

funding/manager.go
```diff
@@ -448,11 +456,15 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
 			//
 			// TODO(roasbeef): sat per byte from order?
 			//  * also other params to set as well
-			chanAmt := int64(matchedOrder.UnitsFilled.ToSatoshis())
+			chanAmt := matchedOrder.UnitsFilled.ToSatoshis()
+			chanAmt += matchedOrderBid.SelfChanBalance
```
Roasbeef, line 474: 👍

Roasbeef (approved):
LGTM 🦎

server.go
```diff
@@ -469,6 +472,9 @@ func (s *Server) setupClient() error {
 		),
 	}
 
+	s.channelAcceptor = NewChannelAcceptor(s.lndServices.Client)
+	s.fundingManager.NotifyShimCreated = s.channelAcceptor.ShimRegistered
```
halseth, line 476: set it directly when creating the struct?

halseth (changes requested):
Main question is whether we really need the pushAmt/min_size restriction, otherwise the code looks good

server.go
guggero: Done :+1:

order/interface.go
```diff
@@ -568,6 +592,28 @@ func (b *Bid) ReservedValue(feeSchedule terms.FeeSchedule) btcutil.Amount {
...
+		return fmt.Errorf("to use self chan balance the min units " +
```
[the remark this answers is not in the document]
guggero: Discussed this offline and agreed to postpone this to later. I created a feature request here: https://github.com/lightninglabs/pool/issues/238

order/tradingfees.go
```diff
@@ -179,6 +179,10 @@ func takerDelta(feeSchedule terms.FeeSchedule,
 	balanceDelta := -satsPremium
 	takerFeesPaid := satsPremium
 
+	// If there is a self channel balance, it must also be paid out of the
+	// taker account.
+	balanceDelta -= selfChanBalance
```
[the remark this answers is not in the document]
guggero: Good point, we definitely need to think about that. I think since they still pay the chain fee part (and execution base fee) it should be okay to do without an additional fee? Or we just need to bump the base fee itself.
Anyway, I think for a first iteration it's probably good to limit the `self_chan_balance` to the order amount itself.

guggero:
> Main question is whether we really need the pushAmt/min_size restriction, otherwise the code looks good

Discussed this offline and agreed to postpone this to later. I created a feature request here: #238

Roasbeef:
@halseth this g2g from your PoV?

Roasbeef:
Assuming yes since you've approved the 2/2?

halseth (approved):
Yeah, all good 👍

PR #235, opened 2021-03-16
sidecar channels 3/3: Accept sidecar channels

guggero:
Depends on #242, only last 14 (small!) commits are new.

This is the third part of what's needed to execute sidecar orders, the actual _acceptance_ of sidecar orders in the auction server stream.


funding/manager.go
```diff
@@ -152,13 +152,32 @@ func (m *Manager) deriveFundingShim(ourOrder order.Order,
...
-		thawHeight = ourOrder.(*order.Bid).LeaseDuration
-		selfChanBalance = ourOrder.(*order.Bid).SelfChanBalance
+		thawHeight = ourOrderBid.LeaseDuration
+		selfChanBalance = ourOrderBid.SelfChanBalance
+
+		// If our order is a sidecar channel, we already have the
+		// multisig key in the ticket and don't need to re-derive it
+		// again here.
+		ticket := ourOrderBid.SidecarTicket
+		if ourOrderIsSidecar && ticket.Recipient != nil {
```
halseth, line 172: can we move this down to an `else` clause with the other option instead?

funding/manager.go
```diff
@@ -601,6 +614,16 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
 		}
 	}
 
+	// If the only order that we were involved in is a sidecar channel and
+	// we are the provider, not the ultimate recipient, then there's nothing
+	// to wait for since the channel is going to be opened to another node.
+	if len(chanPoints) == 0 && someOrdersAreSidecars {
```
halseth, line 629: Isn't it enough to check `len(chanPoints) == 0`?

sidecar_acceptor.go
```diff
@@ -0,0 +1,490 @@
...
+	// update its state in the database and start expecting a channel for it
+	// now.
+	t.State = sidecar.StateExpectingChannel
+	if err := a.store.UpdateSidecar(t); err != nil {
+		return fmt.Errorf("error updating sidecar: %v", err)
+	}
+
+	a.pendingSidecarOrders[nonce] = t
+
+	return a.client.SubscribeAccountUpdates(ctx, &keychain.KeyDescriptor{
```
halseth, line 256: I'm a bit confused here. Is `ExpectChannel` meant to be long running, or is it called and immediately returns?

halseth (changes requested):
High-level pass!

funding/manager.go
```diff
@@ -597,15 +615,18 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
...
+		case <-ctx.Done():
 			// One or more of our peers timed out. At this point the
 			// chanPoints map should only contain channels that
 			// failed/timed out. So we can partially reject all of
 			// them.
 			const timeoutErrStr = "timed out waiting for pending " +
 				"open channel notification"
-			for chanPoint, nonce := range chanPoints {
-				partialReject(nonce, timeoutErrStr, chanPoint)
+			for _, nonce := range chanPoints {
```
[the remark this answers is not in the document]
guggero: Yes, it should be safe. We waited for all goroutines to complete with the `eg.Wait()` in `BatchChannelSetup`. We technically didn't have to use the `partialReject` closure here before, that was just for convenience.

funding/manager.go
```diff
@@ -582,6 +582,70 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
...
+	log.Infof("Batch(%x): opening channels for %v matched orders",
+		batch.ID[:], len(batch.MatchedOrders))
+
+	// For each ask order of ours that's matched, we'll make a new funding
+	// flow, blocking until they all progress to the final state.
+	batchTxHash := batch.BatchTX.TxHash()
+	for ourOrderNonce, matchedOrders := range batch.MatchedOrders {
+		ourOrder, err := getOrder(ourOrderNonce)
+		if err != nil {
+			log.Debugf("Skipping channel funding for order %v "+
```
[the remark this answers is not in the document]
guggero: Hmm, if we want to only allow either sidecar or normal orders to be active at the same time, this should even be a hard error. Fixed.

sidecar_acceptor.go
```diff
@@ -131,3 +193,268 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
...
 	return nil
 }
+
+// ExpectChannel informs the acceptor that a new bid order was submitted for the
+// given sidecar ticket. We subscribe to auction events using the multisig key
+// we gave out when we registered the ticket.
+func (a *SidecarAcceptor) ExpectChannel(ctx context.Context,
+	t *sidecar.Ticket) error {
+
+	a.Lock()
```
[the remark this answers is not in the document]
guggero: We do read and write to the `a.pendingSidecarOrders` map which is also accessed in other goroutines. I reduced the scope a bit to not include the `SubscribeAccountUpdates` call.

sidecar_acceptor.go
```diff
@@ -131,3 +193,268 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
...
+	})
+	if err != nil {
+		log.Errorf("Error sending accept msg: %v", err)
+		return a.sendRejectBatch(batch, err)
+	}
+
+	// We know we're involved in a batch, so let's store it for the next
+	// step.
+	a.Lock()
+	a.pendingBatch = batch
```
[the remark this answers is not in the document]
guggero: We don't really have a state machine in that sense. Only one event loop that processes the messages from the server. But perhaps a bit of refactoring would be good here, going to do that.

sidecar_acceptor.go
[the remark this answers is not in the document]
guggero: Actually, it turned out that a separate mutex was probably the cleaner solution.

sidecar_acceptor.go
guggero: Yes, you are right, the name `SubscribeAccountUpdates` isn't conveying its purpose correctly. I renamed the method and also added a comment here.

funding/manager.go
```diff
@@ -582,6 +582,70 @@ func (m *Manager) BatchChannelSetup(batch *order.Batch,
...
+			if err != nil {
+				return nil, err
+			}
+			chanPoint := wire.OutPoint{
+				Hash: batchTxHash,
+				Index: fundingShim.GetChanPointShim().ChanPoint.
+					OutputIndex,
+			}
+
+			chanPoints[chanPoint] = matchedOrder.Order.Nonce()
```
Roasbeef, line 696: Where is `registerFundingShim` called? If it isn't executed then we won't be expecting the proper parameters and may end up creating a channel that will fail during funding due to a multi-sig key mismatch.

sidecar_acceptor.go
```diff
@@ -131,3 +193,268 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
...
+			return a.sendRejectBatch(batch, err)
+		}
+	}
+
+	// Before we accept the batch, we'll finish preparations on our end
+	// which include applying any order match predicates, connecting out to
+	// peers, and registering funding shim. We don't do a full batch
+	// validation since we don't have any information about the account
+	// that's being used to pay for the sidecar channel.
+	err = a.fundingManager.PrepChannelFunding(
```
Roasbeef, line 313: Ahh I see it's called here indirectly, and we only needed to derive the funding shims back then in order to know what to expect.

sidecar_acceptor.go
```diff
@@ -131,3 +193,268 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
...
+// matchPrepare handles an incoming OrderMatchPrepare message from the server.
+// Since we're only on the receiving end of a sidecar channel (which is always
+// a bid order) the tasks are simplified compared to normal bid order execution.
+func (a *SidecarAcceptor) matchPrepare(
+	msg *auctioneerrpc.OrderMatchPrepare) error {
+
+	// Parse and formally validate what we got from the server.
+	log.Tracef("Received prepare msg from server, batch_id=%x: %v",
+		msg.BatchId, spew.Sdump(msg))
+	batch, err := order.ParseRPCBatch(msg)
```
Roasbeef, line 283: Seems like a bit more cod here can be consolidated with the main (full client) flow for processing a batch. Don't consider it non-blocking, as of right now, still need to get through the rest of the diff.

funding/manager.go
Roasbeef: So you mean turn this into a `switch` statement effectively?

funding/manager.go
```diff
@@ -152,13 +152,32 @@ func (m *Manager) deriveFundingShim(ourOrder order.Order,
...
+
+		// If our order is a sidecar channel, we already have the
+		// multisig key in the ticket and don't need to re-derive it
+		// again here.
+		ticket := ourOrderBid.SidecarTicket
+		if ourOrderIsSidecar && ticket.Recipient != nil {
+			ourMultiSigKey = &keychain.KeyDescriptor{
+				KeyLocator: keychain.KeyLocator{
+					Family: keychain.KeyFamilyMultiSig,
+					Index:  ticket.Recipient.MultiSigKeyIndex,
```
Roasbeef, line 176: Does the locator information actually need to be set if we just have their raw pubkey? Same goes for the index within the ticket itself.

funding/manager.go
Roasbeef: Yeah I think so, I guess this makes it slightly more explicit? Can't think of any other case where we _wouldn't_ have any chan points at this point though.

funding/manager.go
guggero: Extracted just this `if` to make it a bit more readable. A switch wouldn't be optimal as we'd have some duplication.

auctioneer/client.go
```diff
@@ -103,6 +103,11 @@ type Config struct {
 	// GenUserAgent is a function that generates a complete user agent
 	// string given the incoming request context.
 	GenUserAgent func(context.Context) string
+
+	// ConnectSidecar is a flag indicating that instead of connecting to the
+	// default SubscribeBatchAuction RPC the client should connect to
+	// SubscribeSidecar for getting batch updates.
+	ConnectSidecar bool
```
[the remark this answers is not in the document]
guggero: Ah, right. I kind of forgot about that use case. I now made the code fully functional for "parallel" operations. So a "full" Pool client can have sidecar orders (as recipient) and normal orders, even in the same batch. I'll add integration tests for that case as well.

sidecar_acceptor.go
guggero: I wanted to keep this separate so the custom flow can be understood more easily instead of adding even more cases to the code in the normal RPC server.

auctioneer/client.go
```diff
@@ -892,7 +897,11 @@ func (c *Client) connectServerStream(initialBackoff time.Duration,
 	// Now that we know the connection itself is established, we also re-
 	// connect the long-lived stream.
 	ctx, c.streamCancel = context.WithCancel(ctxb)
-	c.serverStream, err = c.client.SubscribeBatchAuction(ctx)
+	if c.cfg.ConnectSidecar {
+		c.serverStream, err = c.client.SubscribeSidecar(ctx)
```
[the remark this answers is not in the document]
guggero: No, it should work directly since the sidecar connection client is a secondary instance with its own connection. This is an interesting test case for an integration test as well :+1:

funding/manager.go
guggero: Unfortunately, we need both the public key as well as the locator (index) for the 3-way authentication in the auctioneer client. The `SignMessage` RPC only takes a locator so we need to know the index. And in fact, looking at the use of `ChanPointShim` in `lnd`, it looks like we only ever use the locator part there as well and ignore the pubkey itself.

funding/manager.go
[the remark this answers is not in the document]
guggero: Scratch the comment above, I wrote that before making it possible to have both sidecar and non-sidecar orders at the same time. Changed it back to `Debugf` and added a comment why an error here is acceptable.

guggero:
I've addressed all comments and also added some more unit tests where it made sense.

halseth (commented):
Code looks good! Want to do a bit of manual testing before finally approving, but I think we are close 👌

log.go
```diff
@@ -20,12 +20,13 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-const Subsystem = "LLMD"
+const Subsystem = "POOL"
```
Roasbeef, line 23: lol

auctioneer/client.go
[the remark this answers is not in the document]
Roasbeef: Dope!

funding/manager.go
```diff
@@ -736,6 +736,76 @@ func (m *Manager) BatchChannelSetup(
...
+	var (
+		chanPoints     = make(map[wire.OutPoint]order.Nonce)
+		fundingRejects = make(map[order.Nonce]*auctioneerrpc.OrderReject)
+	)
+
+	// We need to make sure our whole process doesn't take too long overall
+	// so we create a context that is valid for the whole funding step and
+	// use that everywhere.
+	setupCtx, cancel := context.WithTimeout(
+		context.Background(), m.cfg.BatchStepTimeout,
```
Roasbeef, line 807: Just leaving a note here, but curious to see if we'll need to either increase this value, or use a special value for sidecars particularly if we intend for this to be used in a mobile setting....

server.go
```diff
@@ -499,12 +499,16 @@ func (s *Server) setupClient() error {
 		},
 	}
 
-	// Create the acceptors for receiving sidecar channels.
+	// Create the acceptors for receiving sidecar channels. We need to
+	// create a copy of the auctioneer client configuration because the
+	// acceptor is going to overwrite some of its values.
+	clientCfgCopy := *clientCfg
```
Roasbeef, line 505: 👍

sidecar_acceptor.go
Roasbeef: Gotcha yeah great decision IMO, albeit there's a _bit_ of duplication here, the overall logic is much easier to follow.

Roasbeef (approved):
LGTM 🏎'

The final PR partitioning as well as commit structure made this new diff really easy to review: 👍

PR #237, opened 2021-03-17
Add market info to getinfo call

guggero:
With the server now exposing this new RPC, we want to add the output of
it to the getinfo call to give users an idea what the current market
situation looks like.


halseth (approved):
Travis is failing, otherwise this LGTM 🍏

wpaulino (approved):

PR #245, opened 2021-04-21
poolrpc: fix command example that broke API docs

guggero:
The command example for the `sidecar expectchannel` was incorrect which
lead to the API docs generator erroring out.


Roasbeef (approved):
LGTM 🍍

PR #246, opened 2021-04-23
rpcserver: fix Python gRPC client ALPN property error

PierreRochard:
There is an open issue for Python gRPC clients that is
currently being debugged https://github.com/grpc/grpc/issues/23172

It can be mitigated server-side by specifying h2 in the metadata header.

Tested on macOS and Linux. 

Unsure if this would have any negative side effects on other gRPC client langs.

### DIFF
--- a/config.go
+++ b/config.go
@@ -344,6 +344,7 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
 	}
 
 	tlsCfg := cert.TLSConfFromCert(certData)
+	tlsCfg.NextProtos = []string{"h2"}
 	restCreds, err := credentials.NewClientTLSFromFile(
 		cfg.TLSCertPath, "",
 	)


config.go
```diff
@@ -344,6 +344,7 @@ func getTLSConfig(cfg *Config) (*tls.Config, *credentials.TransportCredentials,
 	}
 
 	tlsCfg := cert.TLSConfFromCert(certData)
+	tlsCfg.NextProtos = []string{"h2"}
```
Roasbeef, line 347: Are you testing w/ `poold` or `litd`? I'm thinking we might need to also port this over to `lnd` as well to patch it for all clients (as `litd` uses the lnd server as a proxy essentially).

PierreRochard: Only 'poold', with the exact same python code and libraries I was not able to reproduce with 'lnd'.

guggero: Interesting. We've had ALPN issues in the past with node.js clients, but this seems to be something different. If you have a bit of time to investigate, could you confirm my hunch that this has something to do with the version of the `google.golang.org/grpc` library? It should be as simple as bumping `google.golang.org/grpc` to `v1.29.1` in `go.mod` of `lnd` and check if you run into the same problem with the exact python code and libraries you used for Pool.

guggero (approved):
LGTM, thanks for the fix!

Roasbeef (approved):
LGTM 🍬

config.go
PierreRochard: I thought the same, but even with the version bump I was unable to reproduce on `lnd` 🤔

guggero: Weird... So it must be another library then. But thanks for trying that! I assume the go version you compiled Pool and lnd with was also the same?

PierreRochard: Yes, same version of go

Issue #177, opened 2020-11-24
poold hangs if litd is already running


halseth:
Should probably exit with an error.

PR #247, opened 2021-04-26
macaroons+clientdb: add timeout to DB open

guggero:
Fixes https://github.com/lightninglabs/pool/issues/177.

To make sure we don't just block for forever if another Pool daemon
process is already running, we add a timeout and error out if obtaining
the unique file lock fails after 5 seconds.

### DIFF
--- a/clientdb/db.go
+++ b/clientdb/db.go
@@ -2,8 +2,10 @@ package clientdb
 
 import (
 	"encoding/binary"
+	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"go.etcd.io/bbolt"
 )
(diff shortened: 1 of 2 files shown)


clientdb/db.go
```diff
@@ -15,6 +17,13 @@ const (
 	// dbFilePermission is the default permission the client database file
 	// is created with.
 	dbFilePermission = 0600
+
+	// DefaultPoolDBTimeout is the default maximum time we wait for the
+	// Pool bbolt database to be opened. If the database is already opened
+	// by another process, the unique lock cannot be obtained. With the
+	// timeout we error out after the given time instead of just blocking
+	// for forever.
+	DefaultPoolDBTimeout = 5 * time.Second
```
wpaulino, line 26: Nit: no need to export.

wpaulino (approved):

clientdb/db.go
guggero: I unified it with the macaroon database timeout and removed `macDatabaseOpenTimeout`, so it now needs to be exported.

halseth (approved):
👍

PR #250, opened 2021-04-29
multi: implement automated negotiation for sidecar channels

Roasbeef:
In this PR, we add automated negotiation for sidecar channel using a newly added server-side network pipe abstraction. A new flag is added to the offer portion of a sidecar channel. If set, then the provider is to specify the entire bid up front in order to allow us to execute the automated negotiation portion. Along the way, we also start to sign the sidecar ticket using the provider's trader pubkey rather than their node's public key. This allows us to use the ticket as an authentication mechanism for the recipient to open a new network pipe to send information to the provider. 

Two new state machines are added to the sidecar acceptor which will drive the sidecar ticket to completion in a synchronous manner. Upon restart, we re-launch these state machines to ensure that the ticket will always eventually be completed. 

Lingering items:
- [x] Add a reliable sender abstraction to ensure proper fault-tolerance of the new acceptor state machines. 
- [x] Ensure the acceptor state machines always terminate after completion or cancellation of the sidecar funding.


auctioneer/client.go
```diff
@@ -1318,6 +1321,108 @@ func (c *Client) MarketInfo(ctx context.Context) (
...
+	return err
+}
+
+// InitAccountCipherBox attempts to initialize a new CipherBox using the
+// account key as an authentication mechanism.
+func (c *Client) InitAccountCipherBox(ctx context.Context, sid [64]byte,
+	acctKey *keychain.KeyDescriptor) error {
+
+	streamSig, err := c.cfg.Signer.SignMessage(
+		ctx, sid[:], acctKey.KeyLocator,
```
halseth, line 1357: should we add a prefix so we are not just signing arbitrary data here?

order/manager.go
```diff
@@ -36,13 +35,6 @@ var (
 	// implement the same batch verification version as the server.
 	ErrVersionMismatch = fmt.Errorf("version %d mismatches server version",
 		CurrentBatchVersion)
-
-	// nodeIdentityKeyLoc is the key locator from which the identity key of
-	// an lnd node is derived.
-	nodeIdentityKeyLoc = keychain.KeyLocator{
```
halseth, line 42: any reason we opted for the node key to begin with?

sidecar_acceptor.go
```diff
@@ -212,7 +212,7 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
 		ctx, int32(keychain.KeyFamilyMultiSig),
 	)
 	if err != nil {
-		return fmt.Errorf("error deriving multisig key: %v", err)
+		return nil, fmt.Errorf("error deriving multisig key: %v", err)
 	}
 
 	ticket.State = sidecar.StateRegistered
```
halseth, line 271: should we actually make a copy rather then mutating this?

funding/manager.go
```diff
@@ -1022,8 +1025,14 @@ func (m *Manager) OfferSidecar(ctx context.Context, capacity,
...
 
-	// Let's now store and return the ticket with the signed offer.
-	err = m.cfg.DB.AddSidecar(ticket)
+	// Let's now store and return the ticket with the signed offer. If a
+	// bid was specified, then we'll commit that as well so we can submit
+	// it to the auctioneer later once we communicate with the recipient.
+	if bid == nil {
+		err = m.cfg.DB.AddSidecar(ticket)
+	} else {
+		err = m.cfg.DB.AddSidecarWithBid(ticket, bid)
```
halseth, line 1034: hm, could we unify this to _always_ require a bid with the sidecar?

sidecar_acceptor.go
```diff
@@ -273,6 +275,409 @@ func (a *SidecarAcceptor) ExpectChannel(ctx context.Context,
...
+	}
+
+	return nil
+}
+
+// AutoAcceptSidecar signals to the acceptor that the recipient of a potential
+// sidecar channel request automated acceptance of the sidecar channel. We'll
+// use the cipher box of the provider of the ticket (and a new one we'll create
+// for the reply side) to finalize negotiation, resulting in a
+func (a *SidecarAcceptor) AutoAcceptSidecar(ctx context.Context,
```
halseth, line 396: nice 😀

halseth (changes requested):
Very cool, this will be a big deal for the usability of sidecar channels 😎

auctioneer/client.go
Roasbeef: IIRC this call ends up adding the prefix underneath by default.

order/manager.go
Roasbeef: Not sure, one other added benefit of this change is that a ticket is no longer bound to a particular _node_, but rather an account in the system.

sidecar_acceptor.go
Roasbeef: Yeah I think I like that path, I made a change elsewhere to start to copy and modify instead of mutating. We've had some bugs on the server-side related to mutation w/ pointer and IMO copying makes the control flow more clear generally.

funding/manager.go
Roasbeef: Yeah I tried that initially, but the trouble I ran into was a circular dependnacy: we can't add a bid to the main interface w/o introducing one. So it would need some bigger overhaul to pull off...

guggero: Yeah, had some similar troubles too, that's why the order nonce is `[32]byte` and not `order.Nonce`. The closest we can come is to add a new `Bid` struct to the sidecar that uses primitive types only and then in the `order` package have a `NewBidFromTicket` function or something.

cmd/pool/order.go
```diff
@@ -49,6 +49,57 @@ var ordersCommands = []cli.Command{
 	},
 }
 
+// baseBidFlags is the set of flags that are common to any command that may
+// need to accept a bid such as the main bid submission method as when a user
+// attempts to offer a sidecar ticket.
+var baseBidFlags = []cli.Flag{
```
guggero, line 55: :+1:

guggero (commented):
Very nice design!
Did a speedy first pass, will review properly next week.

@Roasbeef did you already start working on itests that go through the actual auto negotiation? Otherwise I could look into that, if you want.

auctioneer/client.go
halseth: Ah, true

cmd/pool/sidecar.go
```diff
@@ -67,64 +59,107 @@ func sidecarOffer(ctx *cli.Context) error {
...
 	}
+
 	defer cleanup()
 
+	// Give the user a chance to confirm the order details as this is
+	// binding once submitted.
+	if !ctx.Bool("force") {
+		if err := printOrderDetails(
+			client, btcutil.Amount(bid.Details.Amt),
+			order.SupplyUnit(bid.Details.MinUnitsMatch).ToSatoshis(),
```
guggero, line 142: In the non-auto case this will result in an `integer divide by zero` error because the `MinUnitsMatch` is not set. Probably should only print the order details in the `auto` case.

cmd/pool/sidecar.go
```diff
@@ -31,31 +33,21 @@ var sidecarOfferCommand = cli.Command{
...
-				"band, not part of the sidecar protocol",
-		},
-		cli.Uint64Flag{
-			Name: "lease_duration_blocks",
-			Usage: "the number of blocks the resulting leased " +
-				"channel should be open for",
-			Value: uint64(order.LegacyLeaseDurationBucket),
-		},
-	},
+	),
```
guggero, line 50: `sharedFlags` needs to be added here as well, otherwise it'll complain about the `max batch fee rate must be at least 1 sat/vByte`.

auctioneerrpc/hashmail.proto
```diff
@@ -0,0 +1,118 @@
...
+package poolrpc;
+
+option go_package = "github.com/lightninglabs/pool/auctioneerrpc";
+
+// HashMail exposes a simple synchronous network stream that can be used for
+// various types of synchronization and coordination. The service allows
+// authenticated users to create a simplex stream call a cipher box. Once the
+// stream is created, any user that knows of the stream ID can read/write from
+// the stream, but only a single user can be on either side at a time.
+service HashMail {
```
guggero, line 16: Reminder to add all those RPCs to the auth whitelist in aperture, otherwise a recipient with no channels won't be able to pay for an LSAT.

guggero (commented):
Amazing work on this!

Main comment is about simplifying the persistence in the `clientdb` package a bit. Apart from that it's mostly nits and smaller stuff.

Would be great if we could add an example for the "auto sidecar" route in the `docs/sidecar_channels.md`.

guggero:
Hmm, I ran into a few problems testing this locally with aperture thrown in the mix.

When trying to register the ticket, I get `server closed the stream without sending trailers` on the client side (still need to find out what causes this, same problem we have for some LiT users).
On the server side I see:

`2021-05-04 13:11:52.184 [DBG] SRVR: Stream creation validation failed (id=a2f291561097154db78b8071a11204593152a6c8b9665d8afb76ec95b29fe10f93a8d9eab8ab8e274cf70ba47d2465e463581f029e9d6933f53c633a5247d850): signature not valid for public key 03478ccbbf845f7ad48c2ff598dee53ea20a9b15784363ffba806904572750ec67
2021-05-04 13:11:52.184 [ERR] RPCS: [/poolrpc.HashMail/NewCipherBox]: signature not valid for public key 03478ccbbf845f7ad48c2ff598dee53ea20a9b15784363ffba806904572750ec67`

Could there be something out of sync w.r.t what's being signed in the ticket? The key `03478ccbb...` is the provider's account key.

guggero:
> 2021-05-04 13:11:52.184 [DBG] SRVR: Stream creation validation failed (id=a2f291561097154db78b8071a11204593152a6c8b9665d8afb76ec95b29fe10f93a8d9eab8ab8e274cf70ba47d2465e463581f029e9d6933f53c633a5247d850): signature not valid for public key 03478ccbbf845f7ad48c2ff598dee53ea20a9b15784363ffba806904572750ec67 2021-05-04 13:11:52.184 [ERR] RPCS: [/poolrpc.HashMail/NewCipherBox]: signature not valid for public key 03478ccbbf845f7ad48c2ff598dee53ea20a9b15784363ffba806904572750ec67

> Could there be something out of sync w.r.t what's being signed in the ticket? The key 03478ccbb... is the provider's account key.

Sorry, scratch that! I was using the wrong commit on the server side :see_no_evil: Was able to manually create an auto sidecar through the CLI after changing some stuff in the CLI code (see inline comments).

funding/manager.go
Roasbeef: Hmm, is that really a better option though? As is we don't have a routine that serializes a bid into a _single_ byte stream. Instead, due to historical reasons, we store them strewn amongst a few sub buckets, etc. Therefore it may be the case that the way we store the bid serialized within a sidecar starts to get out of sync with the way we store it on disk. With what's implemented right now, we get the benefit of re-using the exact set of storage routines both in this new sub-bucket and elsewhere.

sidecar/interface.go
```diff
@@ -114,6 +114,10 @@ type Offer struct {
 	// SigOfferDigest is a signature over the offer digest, signed with the
 	// private key that corresponds to the SignPubKey above.
 	SigOfferDigest *btcec.Signature
+
+	// Auto determines if the provider requires that the ticket be
+	// completed using an automated negotiation sequence.
+	Auto bool
```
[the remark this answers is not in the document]
Roasbeef: See my comment here: https://github.com/lightninglabs/pool/pull/250#discussion_r627882233 

IMO we'll end up storing the bid anyway, so we might as well try to re-use as much of the storage routines used elsewhere so we ensure that we have the same high-level schema for storing bids everywhere.

clientdb/sidecar.go
```diff
@@ -70,6 +81,57 @@ func (db *DB) AddSidecar(ticket *sidecar.Ticket) error {
...
+			sidecarBidIndex,
+		)
+		if err != nil {
+			return err
+		}
+
+		// Next, we'll store the nonce of the bid so we can retrieve it
+		// easily in the future.
+		ticketNonce := bid.Nonce()
+		err = bidIndexBucket.Put(sidecarKey, ticketNonce[:])
```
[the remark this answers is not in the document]
Roasbeef: Yeah you're totally right, switching over to just store it there instead.

clientdb/sidecar.go
```diff
@@ -171,3 +285,64 @@ func readSidecar(sourceBucket *bbolt.Bucket, id []byte) (*sidecar.Ticket,
...
+	var w bytes.Buffer
+	if err := SerializeOrder(bid, &w); err != nil {
+		return err
+	}
+
+	err := storeOrderTX(bidBucket, ticketNonce, w.Bytes(), nil)
+	if err != nil {
+		return err
+	}
+	err = storeOrderMinUnitsMatchTX(
```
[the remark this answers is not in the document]
Roasbeef: lolol yeah I had that same itch...if we ended up doing that, then IMO it would def make sense to just store the fully serialized order in the ticket, as that at that point there'd be no key/bucket structure we'd want to mirror

sidecar_acceptor.go
```diff
@@ -223,10 +223,10 @@ func (a *SidecarAcceptor) RegisterSidecar(ctx context.Context,
 		MultiSigKeyIndex: keyDesc.Index,
 	}
 	if err := a.cfg.SidecarDB.AddSidecar(ticket); err != nil {
-		return fmt.Errorf("error storing sidecar: %v", err)
+		return nil, fmt.Errorf("error storing sidecar: %v", err)
 	}
 
-	return nil
+	return ticket, nil
```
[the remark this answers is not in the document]
Roasbeef: Hmm must've been dropped in a rebase, modified to just accept a copy and return a pointer to the new version.

sidecar_acceptor.go
```diff
@@ -274,6 +276,421 @@ func (a *SidecarAcceptor) ExpectChannel(ctx context.Context,
...
+					continue
+				}
+
+				log.Infof("Sending registered ticket (id=%x) "+
+					"to provider (trader_key=%x), stream_id=%x",
+					ticket.ID[:],
+					ticket.Offer.SignPubKey.SerializeCompressed(),
+					providerStreamID[:])
+
+				strTicket, err := sidecar.EncodeToString(ticket)
```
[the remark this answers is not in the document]
Roasbeef: Hmm good point, switched over to just sending over the raw byte string.

sidecar_acceptor.go
```diff
@@ -274,6 +276,421 @@ func (a *SidecarAcceptor) ExpectChannel(ctx context.Context,
...
+	}()
+
+	return nil
+}
+
+// submitSideCarOrder attempts to submit a new bid that's bound to a finalized
+// sidecar ticket that's in the registered phase. If this method returns
+// successfully, then the ticket will have transitioned to the
+// sidecar.StateOrdered state.
+func (a *SidecarAcceptor) submitSideCarOrder(ctx context.Context,
```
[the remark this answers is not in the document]
Roasbeef: done

sidecar_acceptor.go
```diff
@@ -274,6 +276,421 @@ func (a *SidecarAcceptor) ExpectChannel(ctx context.Context,
...
+	// new messages from the recipient.
+	streamID, err := a.deriveProviderStreamID(ticket)
+	if err != nil {
+		return fmt.Errorf("unable to convert offer sig: %w", err)
+	}
+
+	log.Infof("Creating mailbox for ticket=%x, w/ stream_id=%x",
+		ticket.ID[:], streamID[:])
+
+	// TODO(roasbeef): make new context?
```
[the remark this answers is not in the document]
Roasbeef: Was left over from a debugging checkpoint: I was re-using the context from the initial RPC call, but once that call returned it was cancelled causing any of the calls within this method to fail. Resolved here by making a new comment, so can safely remove it!

poolrpc/trader.proto
```diff
@@ -1063,6 +1063,13 @@ message SidecarTicket {
     encoded byte form.
     */
     string ticket = 1;
+
+    /*
+    If the ticket was created using automated negotiation, then the order nonce
+    may also be returned along side the ticket so allow observing the progress
+    of the negotiation via the order state.
+    */
+    bytes order_nonce = 2;
```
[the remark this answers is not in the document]
Roasbeef: Ahh very true, had this in place to enable the itests, removed.

sidecar_acceptor.go
```diff
@@ -368,11 +368,11 @@ func (a *SidecarAcceptor) deriveRecipientStreamID(ticket *sidecar.Ticket) [64]by
 func validateOfferedTicket(ctx context.Context, t *sidecar.Ticket,
 	signer lndclient.SignerClient, db sidecar.Store) error {
 
-	// The ticket should be in the offered state at
-	// this point (has the bid information).
-	if t.State != sidecar.StateOffered {
+	// The ticket should be in the ordered state at this point (has the bid
+	// information).
+	if t.State != sidecar.StateOrdered {
```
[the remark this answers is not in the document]
Roasbeef: done

Roasbeef:
Thanks for the review y'all! Pushed up a new version addressing the latest comments, final thing I want to do is make the state machines a bit more robust (catching proper errors we return on the server-side, making sure it terminates, sending a "resend message" wrapped message on restart, etc).

auctioneer/client.go
```diff
@@ -1318,6 +1321,108 @@ func (c *Client) MarketInfo(ctx context.Context) (
...
+				StreamSig: streamSig,
+			},
+		},
+	}
+	_, err = c.hashMailClient.NewCipherBox(ctx, streamInit)
+	return err
+}
+
+// SendCipherBoxMsg attempts to the passed message into the cipher box
+// identified by the passed stream ID. This message will be on-blocking as long
```
halseth, line 1380: typo: non-blocking

clientdb/sidecar.go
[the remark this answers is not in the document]
halseth: 😅

PR #253, opened 2021-05-05
poolrpc: namespace URI of LSAT REST call

guggero:
When trying to enable a single unified REST proxy in LiT, we noticed
that one URI collides between Loop and Pool (/v1/lsat/tokens).
We change the URI in Pool to fix the collision.

This is a potentially breaking change! Though we do not use REST in LiT
and are not aware of other projects using Pool's REST API yet so it
should impact a small number of users (or nobody really in the best
case).

Related: https://github.com/lightninglabs/lightning-terminal/pull/216


jamaljsr (approved):
LGTM 👍

halseth (approved):

PR #259, opened 2021-05-19
sidecar: switch sidecar encoding from base68 to base58

Roasbeef:
In this commit, we change the encoding of the sidecar ticket from base68
to base58 as it doesn't contain any special characters which makes copy
pasting a string easier.

An example encoding of the new string looks something like:
```
sidecar13HtPsD1UiB2EA4ZfUbnVejG1Pu3p1GW2ydq65NZBuQgTaD9LRESHtKasXwNrQPYqkwkMyJNFiNoBig4PBQQKoKeXz5jHWXZ3pMZNX5JW2FXyf6zKTPWthRQczNQuf2pM8ppjTqHK6tYTEReZk7pFynYJoHnAeBtix28k2PiRSz5qVVmHpWHWFgVjxihEmcD1T1V1g5LwzVSRUyCJ11ajaE4nHdnU1ZR3HdduxXnyyefacfXWAts3tytr7RnWzRbnVPFSx7eVgeinknYvQ3njtiStpQpSacoBH4y1KckVpRAo8N8KuWWmAgYBNpxZfK8RsgViJ26kTpbfuyyHy1uPkAsGtv25iz5pNcopXyaXMnZGyDDWrDf9ZJEgU6wzm4VxwrPfNN4ASR6wLGWAKy1SsCMVFi5Bx6khN4ReSmJVrALWXrbgM52j7evHVRYXa1LZCJTBccCvaRXNYQqxJvnUnLnHW4N4THv4j1y94wmo2SafRfKapxppVW
```


sidecar/codec_test.go
```diff
@@ -9,15 +9,7 @@ import (
...
-	hardcodedTicket = "sidecarAAQgHBgUEAwIBAAIBYwMBAgp_CwgAAAAAAAADCQwIAA" +
-		"AAAAAAA3gNBAAAB-AOIQLVLm5gAB7Vh7eEvz8Y3CkF2DsvNuSWJj8oOxp1iS" +
-		"h5xw9AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACwAAAAAAAAAAA" +
-		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFhRMFSEC1S5uYAAe1Ye3hL8_GNwpBd" +
-		"g7LzbkliY_KDsadYkoeccWIQMZ1hb2o3OyT-XUX7KWS-pAVBloIPQe8aCTqc" +
-		"jiNOV89xcEAAAAAB5kHyALFiEsAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" +
-		"AAACBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGMAAAAAAAAAAA" +
-		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAISgiKSBjWE0AAAAAAAAAAAAAAAAAAA" +
-		"AAAAAAAAAAAAAAAAAAAHcU6Y4="
+	hardcodedTicket = "sidecar13HtPsD1UiB2EA4ZfUbnVejG1Pu3p1GW2ydq65NZBuQgTaD9LRESHtKasXwNrQPYqkwkMyJNFiNoBig4PBQQKoKeXz5jHWXZ3pMZNX5JW2FXyf6zKTPWthRQczNQuf2pM8ppjTqHK6tYTEReZk7pFynYJoHnAeBtix28k2PiRSz5qVVmHpWHWFgVjxihEmcD1T1V1g5LwzVSRUyCJ11ajaE4nHdnU1ZR3HdduxXnyyefacfXWAts3tytr7RnWzRbnVPFSx7eVgeinknYvQ3njtiStpQpSacoBH4y1KckVpRAo8N8KuWWmAgYBNpxZfK8RsgViJ26kTpbfuyyHy1uPkAsGtv25iz5pNcopXyaXMnZGyDDWrDf9ZJEgU6wzm4VxwrPfNN4ASR6wLGWAKy1SsCMVFi5Bx6khN4ReSmJVrALWXrbgM52j7evHVRYXa1LZCJTBccCvaRXNYQqxJvnUnLnHW4N4THv4j1y94wmo2SafRfKav7o2D9"
```
guggero, line 12: nit: wrap to give better idea of how long it actually is?

guggero (approved):
Nice, LGTM :100:

sidecar/codec.go
```diff
@@ -15,77 +16,107 @@ const (
...
-	// encode the gzipped binary data as a string.
-	defaultStringEncoding = base64.URLEncoding
-
-	// sidecarPrefix is a specially prepared prefix that will spell out
-	// "sidecar" when encoded as base64. It will cause all our sidecar
-	// tickets to be easily recognizable with a human readable part.
-	sidecarPrefix = []byte{0xb2, 0x27, 0x5e, 0x71, 0xaa, 0xc0}
+	// sidecarPrefix will serve as our "human readable" prefix to make the
+	// strings easily identifiable.
+	sidecarPrefix = "sidecar"
```
wpaulino, line 21: Nit: pre-existing, but looks like this can be const?

sidecar/codec.go
```diff
@@ -15,77 +16,107 @@ const (
...
-	// ticket) and add 4 bytes of that to the string encoded version. This
-	// will prevent the ticket from appearing valid even if parts of it were
-	// not copy/pasted correctly.
-	hash := sha256.New()
-	_, _ = hash.Write(buf.Bytes())
-	checksum := hash.Sum(nil)[0:4]
-	if _, err := buf.Write(checksum); err != nil {
+	// We'll now hash our full checksum buffer to generate the checksum
+	// that we'll tack onto the end of the ticket.
+	checksum := sha256.Sum256(checksumBuf.Bytes())
```
wpaulino, line 51: The base58 encoding for addresses does a double hash, any reason to not follow that here?

sidecar/codec.go
```diff
@@ -15,77 +16,107 @@ const (
...
+	//
+	// Similar to Bitcoin, we'll add a version so the prefix of the string
+	// is always the same.
+	if _, err := encodeBuf.Write(encodingVersion); err != nil {
+		return "", err
+	}
+	if err := SerializeTicket(&encodeBuf, t); err != nil {
+		return "", err
+	}
+	if _, err := encodeBuf.Write(checksum[:checksumLen]); err != nil {
```
wpaulino, line 64: Looks like this could all be replaced with `base58.CheckEncode`?

sidecar/codec.go
```diff
@@ -15,77 +16,107 @@ const (
...
+	// We'll ensure that the sidecar prefix matches exactly (of the string
+	// version), as otherwise the checksum check won't pass anyway.
+	encodedPrefix := s[:prefixLength]
+	if encodedPrefix != sidecarPrefix {
+		return nil, fmt.Errorf("not a sidecar ticket, invalid prefix "+
+			"%x", encodedPrefix)
+	}
+
+	// Now that we know the prefix passes, we'll move onto verifying the
+	// checksum.
```
wpaulino, line 102: Similarly, we could use `base58.CheckDecode` here.

sidecar/codec.go
```diff
@@ -15,77 +16,107 @@ const (
...
 
-	if len(rawBytes) < len(sidecarPrefix)+checksumLen {
-		return nil, fmt.Errorf("not a sidecar ticket, invalid length")
+	// First, we'll decode the set of raw bytes from the base58 encoding,
+	// snipping off our custom prefix before decoding.
+	rawBytes := base58.Decode(s[prefixLength:])
+
+	// We'll also ensure that the length passes basic sanity checks to not
+	// trip up the slicing logic below.
+	expectedLen := len(sidecarPrefix) + len(encodingVersion) + checksumLen
```
wpaulino, line 87: The sidecar prefix isn't encoded in the payload so it shouldn't be part of this length?

sidecar/codec.go
Roasbeef: Can't use that here since we don't base58 encode the prefix itself. The prior version of this selected characters such that their encoding using base68 would _always_ be "sidecar". In this case, we just append the prefix to the encoding, then include it in the checksum.

sidecar/codec.go
Roasbeef: See above comment.

sidecar/codec.go
Roasbeef: 1 hash seems good enough 🤷🏾‍♂️

Don't need to copy over a Bitcoin-ism IMO if we don't really need it

sidecar/codec.go
Roasbeef: Nice catch, removed. Added a quickcheck test as well to make sure we don't crash w/ arbitrary string input.

sidecar/codec_test.go
Roasbeef: wrapped!

PR #264, opened 2021-05-28
sidecar: add simple exponential back off for mailbox reads

Roasbeef:
In this commit, we fix an existing issue that would cause sidecar
negotiation to halt if either the server restarted, or if the main
connection was torn down. To address this, we add a simple abstraction
that implements stateful exponential back-off that'll wait for a period
of time before trying to read from the mailbox again.

### DIFF
--- a/backoff.go
+++ b/backoff.go
@@ -1,0 +1,42 @@
+package pool
+
+import "time"
+
+const (
+	// maxBackOff is the maximum amount of that that we'll wait to retry to
+	// w/e operation we're wanting to back off.
+	maxBackOff = time.Minute
+
+	// minBackOff is the smallest amount of time we'll wait until we retry
+	// the given operation.
+	minBackOff = time.Millisecond * 100
+)
+
+// backOffer implements a simple exponential back off scheme using a channel
+// and a timer.
+type backOffer struct {
+	backOffPeriod time.Duration
+}
+
+// backOff returns a channel that should be used to wait for the current back
+// off period before doubling the period for the subsequent call to this
+// method.
+func (b *backOffer) backOff(logID string) <-chan time.Time {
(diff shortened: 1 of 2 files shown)


backoff.go
```diff
@@ -0,0 +1,40 @@
...
+// off period before doubling the period for the subsequent call to this
+// method. A quit channel should be specified to allow the back off to exit if
+// the main daemon is shutting down.
+func (b *backOffer) backOff(logID string) <-chan time.Time {
+	currentBackoff := b.backOffPeriod
+	if currentBackoff == 0 {
+		currentBackoff = minBackOff
+	}
+
+	b.backOffPeriod *= 2
```
wpaulino, line 31: Nit: on the first run, this will wait for 200ms instead of the initial 100ms.

backoff.go
```diff
@@ -0,0 +1,40 @@
...
+
+// backOffer implements a simple exponential back off scheme using a channel
+// and a timer.
+type backOffer struct {
+	backOffPeriod time.Duration
+}
+
+// backOff returns a channel that should be used to wait for the current back
+// off period before doubling the period for the subsequent call to this
+// method. A quit channel should be specified to allow the back off to exit if
```
wpaulino, line 23: Nit: comment about quit channel can be removed?

wpaulino (approved):

guggero (approved):
LGTM barring @wpaulino's comments :tada:

backoff.go
Roasbeef: Yep, was left over from an earlier design.

backoff.go
Roasbeef: Removed that local variable, and used multiplication by zero to handle the base case.

PR #268, opened 2021-06-21
multi: add matched order nonce to lease

carlaKC:
Leases need to be cross-referenced with order events to get the timestamp that the lease was finalized at. This is relevant for operators who wish to record their pool lease purchases as expenses in a formal accounting system.

For example, given lease: 
```
{
   "channel_point": "dfed6140efa0257552c400d065572f44f399b5c390adb072e7acab2d3c632006:1",
   "channel_amt_sat": 2000000,
   "channel_duration_blocks": 2016,
   "channel_lease_expiry": 2166,
   "channel_node_key": "02581d130e9613cb4801daf2090b24631690c79112373f6454406516a4414b0711",
   "channel_node_tier": "TIER_0",
   "premium_sat": 999996,
   "clearing_rate_price": 248015,
   "order_fixed_rate": 4960,
   "execution_fee_sat": 2001,
   "chain_fee_sat": 8162,
   "order_nonce": "acd2dc6dd3ea6b75de016abd853af6c868e47805aacba9317c884cbf3c256579",
   "purchased": false,
   "self_chan_balance": 0,
   "sidecar_channel": false
 }
```

We match it with our finalized order event: 
```
 {
   "timestamp_ns": "1624263859826491000",
   "event_str": "OrderMatch(finalized)",
   "matched": {
      "match_state": "FINALIZED",
      "units_filled": 20,
      "matched_order": "bb1f009ff07a7e18c9f19213cf8328602f2639970014414f061f8bde0f0a5589", <-
      "reject_reason": "NONE"
   }
}
```

However, if we have multiple matches for a single order, there is no way to match a specific lease to its order finalized event because we do not surface the nonce of the order that we matched with (except for matching by amount, which can be unreliable if we have duplicate amounts). This PR addresses this issue by surfacing the matched order nonce in `Leases` so that they can be directly matched with order events:

```
"order_nonce": "acd2dc6dd3ea6b75de016abd853af6c868e47805aacba9317c884cbf3c256579",
"matched_order_nonce": "bb1f009ff07a7e18c9f19213cf8328602f2639970014414f061f8bde0f0a5589", <-
```


rpcserver.go
```diff
@@ -2476,7 +2479,6 @@ func rpcOrderStateToDBState(state auctioneerrpc.OrderState) (order.State,
 	switch state {
 	case auctioneerrpc.OrderState_ORDER_SUBMITTED:
 		return order.StateSubmitted, nil
-
```
wpaulino, line 2479: Nit: is this intentional?

wpaulino (approved):

rpcserver.go
carlaKC: nope, removed!

guggero (approved):
Very useful, LGTM :100:

PR #270, opened 2021-06-25
auctioneer+rpc: force a reconnection attempt on unknown error

Roasbeef:
As is, we only attempt to reconnect if we receive a nice io.EOF error
from the gRPC client, which is usually triggered when the server
shutsdown. However in the wild, possibly due to a load balancer and/or
reverse proxy bug, we encounter another error which as is, doesn't
trigger the usual reconnection logic.

In this commit, we now log any unknown error, but then also attempt to
reconnect again by sending the normal "server shutdown" message.

As is, if this doesn't work after the max number of retries, we won't
retry any longer, but perhaps we should still attempt to retry in the
background.

// EDIT(guggero):

It turns out that on startup the auctioneer can get overwhelmed if
hundreds of traders try to authenticate all at once. So some of them run
into a timeout (which sounds like it's the trader's fault but the
timeout covers the whole 3-way handshake and is actually delayed by the
server, not the client) which wasn't handled correctly.

We now inspect the correct error if the authentication fails and try
reconnecting.

### DIFF
--- a/auctioneer/client.go
+++ b/auctioneer/client.go
@@ -50,6 +50,10 @@ var (
 	// signals it's going to shut down.
 	ErrServerShutdown = errors.New("server shutting down")
 
+	// ErrServerErrored is the error that is returned if the auction server
+	// sends back an error instead of a proper message.
+	ErrServerErrored = errors.New("server sent unexpected error")
+
 	// ErrClientShutdown is the error that is returned if the trader client
 	// itself is shutting down.
 	ErrClientShutdown = errors.New("client shutting down")
@@ -577,6 +581,27 @@ func (c *Client) connectAndAuthenticate(ctx context.Context,
 	c.subscribedAcctsMtx.Unlock()
 	err := sub.authenticate(ctx)
 	if err != nil {
+		log.Errorf("Authentication failed for account %x: %v",
+			acctPubKey[:], err)
+
+		// The error that's returned from authenticate() might just be
+		// an error that occurred when sending on the stream. If the
+		// server is timing us out (or is shutting down) before we can
+		// send the final message, this might not be the _cause_ of the
+		// problem but just a follow-up symptom. So we need to look if
+		// there's anything on the tempErrChan (which we write to when
+		// getting unexpected messages or errors from the server) that's
+		// more conclusive.
+		select {
+		case err := <-tempErrChan:
+			// Ah, so it's the server shutting down, so let's re-
+			// try our connection.
+			if err == ErrServerErrored {
+				return sub, false, c.HandleServerShutdown(nil)
+			}
+
+		default:
+		}
 		return sub, false, err
 	}
 
@@ -968,9 +993,13 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
 				return
 			}
 
-			// Any other error we want to report back.
+			log.Errorf("Server connection error received: %v", err)
+
+			// For any other error type, we'll attempt to trigger
+			// the reconnect logic so we'll always try to connect
+			// to the server in the background.
 			select {
-			case c.errChanSwitch.ErrChan() <- err:
+			case c.errChanSwitch.ErrChan() <- ErrServerErrored:
 			case <-c.quit:
 			}
 			return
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -256,8 +256,14 @@ func (s *rpcServer) serverHandler(blockChan chan int32, blockErrChan chan error)
 					rpcLog.Errorf("Error closing stream: %v",
 						err)
 				}
+
+				// TODO(roasbeef): re-trigger reconnection
```
Roasbeef, line 260: I think what we want to do here instead is create a new `ErrUnexpectedServerShutdown` (the non `io.EOF` case), and then re trigger a reconnection.

guggero:
I tried to reproduce the "trader not reconnecting" bug in our itests.
But the reconnect works, no matter the version of aperture that's being used. So there might be something infrastructure related there as well and we should just test it out on our testing infrastructure so see what really prevents the traders from reconnecting.

Roasbeef:
> So there might be something infrastructure related there as well and we should just test it out on our testing infrastructure so see what really prevents the traders from reconnecting.

Interesting, yeah I think the move here is to try to crank up logging on a modified client, and reproduce it in the context of our normal staging environment.

guggero:
Finally saw the error in the wild:
```
2021-07-21 07:53:07.605 [INF] AUCT: Successfully connected to auction server
2021-07-21 07:53:17.790 [ERR] AUCT: [/poolrpc.ChannelAuctioneer/SubscribeBatchAuction]: rpc error: code = Unknown desc = no subscription received before timeout
2021-07-21 07:53:17.790 [TRC] AUCT: Received msg=(*auctioneerrpc.ServerAuctionMessage)(nil), err=rpc error: code = Unknown desc = no subscription received before timeout from server
```

Seems after a server restart all traders try to re-connect at the same time, putting the server in a rough spot. so the server times out and the client falls out of the retry mechanism.

auctioneer/client.go
```diff
@@ -577,6 +577,27 @@ func (c *Client) connectAndAuthenticate(ctx context.Context,
 	c.subscribedAcctsMtx.Unlock()
 	err := sub.authenticate(ctx)
 	if err != nil {
+		log.Errorf("Authentication failed for account %x: %v",
+			acctPubKey[:], err)
+
+		// The error that's returned from authenticate() might just be
```
Roasbeef, line 583: Ah ok, so before this, we would just return early here, never read on the channel and then ultimately give up until a restart?

auctioneer/client.go
```diff
@@ -968,9 +989,13 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
...
 
-			// Any other error we want to report back.
+			log.Errorf("Server connection error received: %v", err)
+
+			// For any other error type, we'll attempt to trigger
+			// the reconnect logic so we'll always try to connect
+			// to the server in the background.
 			select {
-			case c.errChanSwitch.ErrChan() <- err:
+			case c.errChanSwitch.ErrChan() <- ErrServerShutdown:
```
Roasbeef, line 998: This was from the original commit, right? Do we still want this? 

IIRC, the error assertion elsewhere in the codebase would cause us to actually _not_ reconnect (if this error is sent) since we assumed that a reconnection was already in place.

auctioneer/client.go
```diff
@@ -538,7 +538,7 @@ func (c *Client) connectAndAuthenticate(ctx context.Context,
 	c.streamMutex.Unlock()
 
 	if needToConnect {
-		err := c.connectServerStream(0, initialConnectRetries)
+		err := c.connectServerStream(0, reconnectRetries)
```
Roasbeef, line 541: 👍

auctioneer/client.go
guggero: Yes, exactly. It would just hang there. In fact if the connect was issued on startup, we'd never fully complete the server's `Start()` method and therefore hang in a state where we don't listen for shutdown signals yet. So worst case you'd even need to kill the daemon.

auctioneer/client.go
guggero: Ah, you're right. Created a specific error for this.

guggero (approved):

PR #271, opened 2021-07-01
Bump default LSAT max routing fee and make it configurable

guggero:
Attempts to fix the `no_route` error people are running into because the 10 satoshi default routing fee limit for the 1k sat LSAT payment is too restrictive in some situations.
We bump the default to 50 sats and also make the value configurable.


Roasbeef (approved):
LGTM 🍠

PR #272, opened 2021-07-19
When a macaroon string is passed in, don't create new macaroons

orbitalturtle:
When using the Lightning Terminal in stateless remote mode (described here: https://github.com/lightninglabs/lightning-terminal/issues/160), Pool shouldn't create a new macaroon. This is a small change to stop macaroons from being created when macaroon data is passed to Pool directly.

Relies on: https://github.com/lightninglabs/lndclient/pull/51

### DIFF
--- a/server.go
+++ b/server.go
@@ -132,7 +132,7 @@ func (s *Server) Start() error {
 	// what we need
 	s.lndClient, err = lndclient.NewBasicClient(
 		s.cfg.Lnd.Host, s.cfg.Lnd.TLSPath,
-		path.Dir(s.cfg.Lnd.MacaroonPath), s.cfg.Network,
+		path.Dir(s.cfg.Lnd.MacaroonPath), "", "", s.cfg.Network,
 		lndclient.MacFilename(path.Base(s.cfg.Lnd.MacaroonPath)),
 	)
 	if err != nil {
@@ -305,7 +305,7 @@ func (s *Server) Start() error {
 // StartAsSubserver is an alternative start method where the RPC server does not
 // create its own gRPC server but registers on an existing one.
 func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
-	lndGrpc *lndclient.GrpcLndServices) error {
+	lndGrpc *lndclient.GrpcLndServices, macData string) error {
 
 	if atomic.AddInt32(&s.started, 1) != 1 {
 		return fmt.Errorf("trader can only be started once")
@@ -330,12 +330,17 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
 		}
 	}()
 
-	// Start the macaroon service and let it create its default macaroon in
-	// case it doesn't exist yet.
-	if err := s.startMacaroonService(); err != nil {
-		return err
+	// If a macaroon string was passed in from lndclient, then we already 
+	// have a macaroon for Pool and don't need to start the macaroon
+	// service.
+	if macData == "" {
+		// Start the macaroon service and let it create its default 
+		// macaroon in case it doesn't exist yet.
+		if err := s.startMacaroonService(); err != nil {
+			return err
+		}
+		shutdownFuncs["macaroon"] = s.stopMacaroonService
 	}
-	shutdownFuncs["macaroon"] = s.stopMacaroonService
 
 	// Setup the auctioneer client and interceptor.
 	err := s.setupClient()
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -330,12 +330,17 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
...
-	// case it doesn't exist yet.
-	if err := s.startMacaroonService(); err != nil {
-		return err
+	// If a macaroon string was passed in from lndclient, then we already 
+	// have a macaroon for Pool and don't need to start the macaroon
+	// service.
+	if macData == "" {
+		// Start the macaroon service and let it create its default 
+		// macaroon in case it doesn't exist yet.
+		if err := s.startMacaroonService(); err != nil {
```
guggero, line 339: We can't just not start the macaroon service here. Otherwise we'll run into a nil pointer error in `ValidateMacaroon` if we're actually trying to validate an incoming one. Instead I think we need an additional parameter that specifies no macaroons should be created. Same goes for the other PRs in Loop and Faraday.

guggero (changes requested):

server.go
orbitalturtle: Got it, my bad! Made that change when you get a chance to take a look. If this looks ok I'll make similar changes in the faraday and loop PRs as well.

macaroons.go
```diff
@@ -140,11 +140,11 @@ var (
 		}},
 	}
 
-	// allPermissions is the list of all existing permissions that exist
+	// AllPermissions is the list of all existing permissions that exist
 	// for poold's RPC. The default macaroon that is created on startup
 	// contains all these permissions and is therefore equivalent to lnd's
 	// admin.macaroon but for pool.
-	allPermissions = []bakery.Op{{
+	AllPermissions = []bakery.Op{{
```
guggero, line 147: Why/where is this needed? Should be its own commit with a short explanation in the commit message.

macaroons.go
```diff
@@ -206,8 +206,11 @@ func (s *Server) startMacaroonService() error {
 		return fmt.Errorf("unable to unlock macaroon DB: %v", err)
 	}
 
-	// Create macaroon files for pool CLI to use if they don't exist.
-	if !lnrpc.FileExists(s.cfg.MacaroonPath) {
+	// If a macaroon string was passed in from lndclient, then we already
+	// have a macaroon for Pool and don't need to create macaroon files.
+	// Otherwise, Create macaroon files for pool CLI to use if they don't
+	// exist.
+	if macData == "" && !lnrpc.FileExists(s.cfg.MacaroonPath) {
```
guggero, line 213: If we don't actually look at the data, couldn't this just be a boolean called "noMacaroonCreation"?

guggero (commented):
Definitely looks better this way.

macaroons.go
orbitalturtle: good call

orbitalturtle:
Thanks for the review @guggero ! Made those changes.

guggero (commented):
LGTM pending the dependent PRs in `lndclient` and `lnd`.

The `lnd` PR is very close and can be merged as soon as the three remaining issues are addressed. Then we can update the `lndclient` and this PR to use the actual versions in `go.mod`.

orbitalturtle:
@guggero great, I updated lnd and lndclient with the actual versions. I also attempted to take care of all the changes required to bump those up... I'm especially not sure if I updated macaroons.NewService method correctly.

guggero:
@orbitalturtle I forgot to mention this to you, sorry. There is a branch I pushed to the main Pool repo called `lnd-14` that you can rebase on to get all the things in you need to make this work. Your PR will need to wait for the 0.14 lnd compatibility anyway, so can just make that a hard dependency.

orbitalturtle:
@guggero Got it, no problem, Just updated.

guggero (approved):
LGTM, thanks!

server.go
```diff
@@ -142,7 +142,7 @@ func (s *Server) Start() error {
 
 	// Start the macaroon service and let it create its default macaroon in
 	// case it doesn't exist yet.
```
positiveblue, line 144: This comment gets outdated after `noMacaroonCreation` is set to `false`.

server.go
```diff
@@ -335,7 +335,7 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
...
```
positiveblue, line 337: nit: change the comment so it reflects that `it creates its default macaroon in case it doesn't exist yet and noMacaroonCreation is true`

positiveblue (approved):

guggero:
Can be rebased onto master now. Sorry for the indirect/unintended close.

guggero:
I'm going to take over those PRs so they can hopefully make it into the lnd 0.14.0 release of LiT. I hope you don't mind!

Issue #240, opened 2021-03-25
orders: allow matching with special requirements (anchor commitments etc.)


cryptagoras:
As a seller/maker one issue that varies the experience of selling a channel lease is if the buyer will stay online or for whatever reason making me force close the channel which usually results in non-trivial economic damage.

Anchor channels provide a serious improvement on these cases, so as a seller I would like to be able to specify that I'm selling only anchor channel leases.

This will provide a greater peace of mind and allow me to provide better offers without having to account for potentially costly force-closings.

(I know this is probably too niche and hopefully anchors will get adopted faster enough, but maybe keep it in mind :clinking_glasses:  )

Roasbeef:
Def, this is in the works as an explicit order type. Needs a small change to lnd that makes channel types _explicit_: https://github.com/lightningnetwork/lnd/pull/5071

guggero:
Going through the older issues, I noticed this was fixed by https://github.com/lightninglabs/pool/pull/300.

PR #273, opened 2021-07-27
Introduce new channel type order extension and support script enforced channel leases

wpaulino:
Documentation still WIP.

Depends on https://github.com/lightninglabs/lndclient/pull/59 and https://github.com/lightningnetwork/lnd/pull/5549.
Fixes https://github.com/lightninglabs/pool/issues/240.


Roasbeef:
> ChannelTypePeerDependent: This exists to remain backwards-compatible. The type of a channel resulting from a matched order will vary based on the node features shared between the asker and bidder.

Nice, this'll be useful for when we add the new zero conf order type eventually.

auctioneerrpc/auctioneer.proto
```diff
@@ -843,7 +861,7 @@ message ServerOrder {
     /*
     The type of the channel that should be opened.
     */
-    uint32 chan_type = 12;
+    OrderChannelType channel_type = 12;
```
Roasbeef, line 872: This was never actually used, so it isn't technically a breaking change right?

order/rpc_parse.go
```diff
@@ -62,6 +62,25 @@ func ParseRPCOrder(version, leaseDuration uint32,
 	}
 	kit.MinUnitsMatch = SupplyUnit(details.MinUnitsMatch)
 
+	switch details.ChannelType {
+	// Default value, trader didn't specify a channel type.
+	case auctioneerrpc.OrderChannelType_ORDER_CHANNEL_TYPE_UNKNOWN:
+		// TODO: Switch to script enforcement by default once we can
```
Roasbeef, line 68: Make sense, I wonder if there's something generic we can slot in _now_ to make this a bit more foolproof (harder for us to forget about it).

auctioneerrpc/auctioneer.proto
```diff
@@ -564,12 +571,6 @@ message OrderMatchFinalize {
     The final transaction ID of the published batch transaction.
     */
     bytes batch_txid = 2;
-
-    /*
-    The current block height at the time the batch transaction was published to
-    the network.
-    */
-    uint32 height_hint = 3;
```
Roasbeef, line 572: Is this a breaking change? Off top, can't think of anything that relies on this field as is.

order/batch_verifier.go
```diff
@@ -77,6 +82,13 @@ func (v *batchVerifier) Verify(batch *Batch) error {
 		return ErrVersionMismatch
 	}
 
+	// Reject the patch if we're too far in the past or future compared to
```
Roasbeef, line 85: patch -> batch

funding/manager.go
```diff
@@ -315,6 +315,15 @@ func (m *Manager) deriveFundingShim(ourOrder order.Order,
 		selfChanBalance = matchedOrder.Order.(*order.Bid).SelfChanBalance
 	}
 
+	// If either order requires script enforcement, our thaw height needs to
+	// be an absolute height instead of a relative one.
+	switch {
+	case ourOrder.Details().ChannelType == order.ChannelTypeScriptEnforced:
+		fallthrough
+	case matchedOrder.Order.Details().ChannelType == order.ChannelTypeScriptEnforced:
+		thawHeight += batchHeightHint
```
Roasbeef, line 324: Shouldn't this instead add the height hint fuzz factor? Otherwise we end up doubling (?) what we think the final lease expiration height should be.

channel_acceptor.go
```diff
@@ -139,5 +140,38 @@ func (s *ChannelAcceptor) acceptChannel(_ context.Context,
 		}, nil
 	}
 
+	switch expectedChanBid.ChannelType {
+	// The bid doesn't have specific requirements for the channel type.
+	case order.ChannelTypePeerDependent:
+
```
Roasbeef, line 146: Style nit: add a `break` here to make it more explicit that we're not falling through (one of the quirk's w/ Go's `switch` statements IMO).

Roasbeef (changes requested):

auctioneerrpc/auctioneer.proto
wpaulino: Correct, and IIUC it also isn't a breaking change according to the proto3 language guide:

> Enumerator constants must be in the range of a 32-bit integer.

> enum is compatible with int32, uint32, int64, and uint64 in terms of wire format (note that values will be truncated if they don't fit). However be aware that client code may treat them differently when the message is deserialized: for example, unrecognized proto3 enum types will be preserved in the message, but how this is represented when the message is deserialized is language-dependent. Int fields always just preserve their value.

auctioneerrpc/auctioneer.proto
wpaulino: Yes, it wasn't being used by traders at all. I think rather than removing it though, we need to mark it as reserved in case we decide to extend this proto message in the future?

funding/manager.go
wpaulino: The `thawHeight` will continue to be relative, unless a script enforced channel is to be opened, in which case we convert it to an absolute height as required by adding the proposed batch height.

wpaulino:
This now depends on https://github.com/lightninglabs/pool/pull/275.

order/rpc_parse.go
guggero: We could look at lnd's feature flags on startup (similar as we do for finding out if a trader is tor only) and use the script enforced type for ask orders if the feature is turned on?

auctioneerrpc/auctioneer.proto
```diff
@@ -545,6 +545,13 @@ message OrderMatchPrepare {
     within and the discovered market clearing price.
     */
     map<uint32, MatchedMarket> matched_markets = 10;
+
+    /*
+    The earliest absolute height in the chain in which the batch transaction can
+    be found within. This will be used by traders to base off their absolute
+    channel lease maturity height.
+    */
+    uint32 batch_height_hint = 11;
```
guggero, line 561: We should use a new version tag for `auctioneerrpc` (for example `v1.0.3` in this case) in the main `go.mod` file when we change something. That way we only need to push a tag once we merge the PR and don't need yet another PR to bump the version _after_ pushing the tag. Saves a round trip and doesn't make anything invalid (I think) since we have a replace everywhere.

cmd/pool/order.go
```diff
@@ -230,6 +245,18 @@ func parseCommonParams(ctx *cli.Context, blockDuration uint32) (*poolrpc.Order,
 
 	params.RateFixed = rateFixed
 
+	// Determine the appropriate channel type that should be opened upon an
+	// order match.
+	channelType := ctx.String("channel_type")
+	switch channelType {
+	case "legacy":
```
guggero, line 252: nit: use constants here as well.

guggero (commented):
Did a high-level pass, still getting through the other dependent PRs. Great work though, diff reads very well!

guggero:
Can be rebased now that #275 is in.

order/rpc_parse.go
wpaulino: @Roasbeef also had an idea of coming out with a new lease duration market where all orders signal script enforcement.

auctioneerrpc/auctioneer.proto
wpaulino: Ah, good idea! Done.

PR #275, opened 2021-07-29
poolrpc: use docker to compile protos, update lnd to master with grpc-gateway upgrade

guggero:
To make pool work with the latest lnd version inside of LiT, we need
to upgrade the grpc-gateway library to the same v2 version here too.


wpaulino (approved):

Roasbeef (approved):
LGTM 🧪

PR #279, opened 2021-08-16
gomod: bump to lndclient 13-7

carlaKC:
This commit bumps the compile time version of pool to a version of lnd where we have addressed our etcd modules workaround so that code that has already removed the workaround can import pool.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,10 +13,10 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.5.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/lightninglabs/aperture v0.1.6-beta
-	github.com/lightninglabs/lndclient v0.12.0-8
-	github.com/lightninglabs/pool/auctioneerrpc v1.0.2
+	github.com/lightninglabs/lndclient v0.13.0-7
+	github.com/lightninglabs/pool/auctioneerrpc v1.0.3
 	github.com/lightninglabs/protobuf-hex-display v1.4.3-hex-display
-	github.com/lightningnetwork/lnd v0.13.0-beta.rc5.0.20210728112744-ebabda671786
+	github.com/lightningnetwork/lnd v0.13.0-beta.rc5.0.20210802115842-44971f0c46c9
 	github.com/lightningnetwork/lnd/cert v1.0.3
 	github.com/stretchr/testify v1.7.0
 	github.com/urfave/cli v1.20.0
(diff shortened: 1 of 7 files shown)


carlaKC:
This commit bumps our compile time lndclient, but not minimum version. The danger here is that we now have some APIs available at compile time, but our minimum runtime version does not have those APIs. 
* This is already the case, because we were previously pinned to 0.12, but have minimum version 0.11
* Added a pull request template to nudge devs

Did a quick runthrough of the apis pool depends on. Afaik we don't need to bump minimum version since these are all available in 0.11, but we may want to do so anyway? cc @guggero 
`server`
* s.server.lndServices.WalletKit.DeriveKey

`lsat.NewInterceptor`
* i.lnd.Client.PayInvoice
* i.lnd.Router.TrackPayment

`NewChannelAcceptor`
* s.lightning.ChannelAcceptor

`funding.NewManager`
*  m.cfg.WalletKit.DeriveKey
*  client.GetInfo(ctx)
* baseClient.AbandonChannel
* m.cfg.LightningClient.Connect
* m.cfg.LightningClient.ListChannels
* m.cfg.LightningClient.PendingChannels
* signer.SignMessage
* baseClient.FundingStateStep

`&auctioneer.Config`
* signer.SignMessage

`NewSidecarAcceptor`
* signer.VerifyMessage
*  a.cfg.Wallet.DeriveNextKey

guggero:
Hmm, I think we'll want people to update to 0.13 as soon as possible anyway. So a soft bump to 0.12 shouldn't be a problem and we should do it.

auctioneerrpc/gen_protos_docker.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
guggero, line 1: nit: those two files aren't actually used as we compile the `auctioneerrpc` package from the `poolrpc`'s `gen_protos.sh` (see bottom of the file).

carlaKC: Ah yeah, I think I added these when I was figuring out where it's generated + accidental commit :') removed

go.mod
```diff
@@ -13,10 +13,10 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.5.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/lightninglabs/aperture v0.1.6-beta
-	github.com/lightninglabs/lndclient v0.12.0-8
-	github.com/lightninglabs/pool/auctioneerrpc v1.0.2
+	github.com/lightninglabs/lndclient v0.13.0-7
+	github.com/lightninglabs/pool/auctioneerrpc v1.0.3
```
guggero, line 17: :pray:

guggero (approved):
Very nice, LGTM :100:

carlaKC:
@guggero would you mind merging for me? I don't have perms.

carlaKC:
Thanks!

PR #282, opened 2021-08-20
auctioneer: fix stream nil panic

guggero:
We were missing a few selects on the quit channel which could lead to us
making another iteration in the loop while the server stream was already
set to nil. We fix that by making sure we always return one way or
another if the quit channel is closed.
Just to make the panic impossible to happen again, we also return from
the loop early if the server stream is nil, just in case.

### DIFF
--- a/auctioneer/client.go
+++ b/auctioneer/client.go
@@ -966,6 +966,14 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
 		default:
 		}
 
+		// This should never happen as we always close the quit channel
+		// before we set the connection to nil (which _should_ cause us
+		// to return in the first place), but just to be safe and avoid
+		// a panic.
+		if c.serverStream == nil {
+			return
+		}
+
 		// Read next message from server.
 		msg, err := c.serverStream.Recv()
 		log.Tracef("Received msg=%#v, err=%v from server", msg, err)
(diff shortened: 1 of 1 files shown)


auctioneer/client.go
```diff
@@ -966,6 +966,14 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
 		default:
 		}
 
+		// This should never happen as we always close the quit channel
+		// before we set the connection to nil (which _should_ cause us
+		// to return in the first place), but just to be safe and avoid
+		// a panic.
+		if c.serverStream == nil {
```
Roasbeef, line 973: 👍  for the defensive check here

auctioneer/client.go
```diff
@@ -1026,16 +1034,20 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
...
+				case <-c.quit:
+				}
 				return
 			}
 
 			// Inform the subscription about the arrived challenge.
 			select {
 			case acctSub.msgChan <- msg:
 			case <-c.quit:
+				return
```
Roasbeef, line 1050: Ahh nice catch!

auctioneer/client.go
```diff
@@ -1100,13 +1121,15 @@ func (c *Client) readIncomingStream() { // nolint:gocyclo
 			select {
 			case c.FromServerChan <- msg:
 			case <-c.quit:
+				return
```
Roasbeef, line 1124: Hmm, I wonder if there's some sort of static check library that can help us spot these in the future...

Roasbeef (approved):
LGTM 🏎

Issue #263, opened 2021-05-26
sidecar: add RPC for printticket


jamaljsr:
**Is your feature request related to a problem? Please describe.**
In LiT, I'd like to prompt the user for the sidecar ticket, then display some of the decoded data before asking the user to confirm. It would be good for the user to see things like Capacity, PushAmount, and LeaseDurationBlocks in the UI.

**Describe the solution you'd like**
Add an RPC to accept a ticket and responds with the decoded data similar to the CLI command `pool sidecar printticket`:
```
{
	"ID": "7ff67a86665ec22b",
	"Version": 0,
	"State": "offered",
	"Capacity": 1000000,
	"PushAmount": 100000,
	"LeaseDurationBlocks": 2016,
	"OfferSigningPubKey": "024f88664f523458189c224f7f90c7a4ab200f533f8e34e5de430f9a6c6f7f4e0c",
	"RecipientNodePubKey": "",
	"RecipientMultiSigPubKey": "",
	"RecipientMultiSigKeyIndex": 0,
	"OrderNonce": "2f4f0d2bfa7031af48d04cb63cd641777285869cfd7fff1d40d371aeb01d9158"
}
```

**Describe alternatives you've considered**
I considered implementing the decoding logic in TS, but after skimming through the source of [codec.go](https://github.com/lightninglabs/pool/blob/master/sidecar/codec.go#L75) and [tlv.go](https://github.com/lightninglabs/pool/blob/master/sidecar/tlv.go#L111), porting all of this would require a fair amount of effort.

PR #285, opened 2021-09-06
Add DecodeSidecarTicket RPC

guggero:
Fixes https://github.com/lightninglabs/pool/issues/263.

Adds the `DecodeSidecarTicket` RPC.

#### Pull Request Checklist
  used.


rpcserver.go
```diff
@@ -2447,6 +2447,67 @@ func (s *rpcServer) ExpectSidecarChannel(ctx context.Context,
...
+	resp := &poolrpc.DecodedSidecarTicket{
+		Id:                        t.ID[:],
+		Version:                   uint32(t.Version),
+		State:                     t.State.String(),
+		OfferCapacity:             uint64(t.Offer.Capacity),
+		OfferPushAmount:           uint64(t.Offer.PushAmt),
+		OfferLeaseDurationBlocks:  t.Offer.LeaseDurationBlocks,
+		OfferSignPubkey:           serializePubKey(t.Offer.SignPubKey),
+		OfferAuto:                 t.Offer.Auto,
+		ExecutionPendingChannelId: nil,
```
bhandras, line 2478: nit: no need to `nil` `ExecutionPendingChannel`.

rpcserver.go
```diff
@@ -2447,6 +2447,67 @@ func (s *rpcServer) ExpectSidecarChannel(ctx context.Context,
...
+	if t.Order != nil {
+		resp.OrderBidNonce = t.Order.BidNonce[:]
+
+		if t.Order.SigOrderDigest != nil {
+			resp.OrderSignature = t.Order.SigOrderDigest.Serialize()
+		}
+	}
+
+	if t.Execution != nil {
+		resp.ExecutionPendingChannelId = t.Execution.PendingChannelID[:]
```
bhandras, line 2505: Do we want to return raw byte arrays here and for `OrderBidNonce` and `ID`? I see in the next commit we use `printJson` but perhaps we could use a hex encoding for these byte buffers instead?

bhandras (commented):
PR LGTM, just have one question regarding encoding.

rpcserver.go
guggero: Yes, fixed.

rpcserver.go
guggero: We're using `github.com/lightninglabs/protobuf-hex-display` for the `printRespJSON()`, so it will be printed on the command line as hex. I actually prefer having things as raw bytes on the protobuf level. That does make it a bit more difficult to use with REST since it'll be base64 by default. But using hex encoded strings just because of that feels kind of inefficient. Not sure about this tbh. What's your take, @Roasbeef?

bhandras: Great that `protobuf-hex-display` takes care of this. As for REST I'd say base64 is probably more convenient to use than hex encoding.

sidecar/codec_test.go
```diff
@@ -43,10 +43,12 @@ func TestEncodeDecode(t *testing.T) {
...
 				S: new(big.Int).SetInt64(22),
 			},
+			Auto: true,
 		},
 		Recipient: &Recipient{
-			NodePubKey:     testPubKey,
-			MultiSigPubKey: testPubKey2,
+			NodePubKey:       testPubKey,
+			MultiSigPubKey:   testPubKey2,
+			MultiSigKeyIndex: 77,
```
Roasbeef, line 51: 👍

Roasbeef (approved):
LGTM 🏊‍♀️

Issue #191, opened 2020-12-10
accounts: add ability to specify raw sat/byte when creating an account 


Roasbeef:
When making an account on the CLI, we're only able to specify the confirmation target rather than the raw sat/byte. Generally the confirmation target is much less granular than a direct sat/byte due to jumps in the confirmation confidence intervals and user behavior (people like multiples of 6 confirmations as an example).

guggero:
Fixed by #288.

PR #288, opened 2021-09-23
Add sats_per_byte option to newAccounts cmd

positiveblue:
Add ability to specify raw sat/byte when creating an account 

#191


account/manager.go
```diff
@@ -258,7 +256,8 @@ func (m *Manager) QuoteAccount(ctx context.Context, value btcutil.Amount,
 // InitAccount handles a request to create a new account with the provided
 // parameters.
 func (m *Manager) InitAccount(ctx context.Context, value btcutil.Amount,
-	expiry, bestHeight, confTarget uint32) (*Account, error) {
+	feeRate chainfee.SatPerKWeight, expiry, bestHeight uint32,
+) (*Account, error) {
```
guggero, line 260: nit: weird wrapping. Should wrap before `bestHeight` instead.

account/manager.go
```diff
@@ -194,9 +194,7 @@ func (m *Manager) start() error {
 		// TODO(guggero): Refactor this to extract the init/funding
 		// part so we properly abandon the account if it fails before
 		// publishing the TX instead of trying to re-fund on startup.
-		err := m.resumeAccount(
-			ctx, account, true, false, DefaultFundingConfTarget,
-		)
+		err := m.resumeAccount(ctx, account, true, false, 0)
```
guggero, line 197: Should we add a comment here that the fee rate of 0 means "should estimate before publishing"?

rpcserver.go
```diff
@@ -510,17 +510,47 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
 	}
 
-	// Determine the desired transaction fee.
-	confTarget := req.GetConfTarget()
-	if confTarget < 1 {
-		return nil, fmt.Errorf("confirmation target must be " +
-			"greater than 0")
+	var feeRate chainfee.SatPerKWeight
+
+	switch {
```
guggero, line 515: Add error case for when both values are set. And use `>0` in both cases?

rpcserver.go
```diff
@@ -510,17 +510,47 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
+	case req.GetFeeRateSatPerKw() != 0:
+		feeRate = chainfee.SatPerKWeight(req.GetFeeRateSatPerKw())
```
guggero, line 517: nit: with longer `switch` statements we add a newline before each new `case` to give the eye some more orientation where a block ends.

rpcserver.go
```diff
@@ -510,17 +510,47 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
+	case req.GetConfTarget() > 0:
+		// Determine the desired transaction fee.
+		confTarget := req.GetConfTarget()
+
+		resp, err := s.QuoteAccount(context.Background(),
```
guggero, line 522: Since we're already checking that the conf target is larger than 0, we can just call the `s.accountManager.QuoteAccount()` here directly instead of doing a call on the RPC server level.

rpcserver.go
```diff
@@ -510,17 +510,47 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
+
+		log.Infof("Estimated total chain fee of %v for new account with "+
+			"value=%v, conf_target=%v", feeRate, value, confTarget)
+	default:
+		return nil, fmt.Errorf("either sat/vbyte or confirmation target " +
+			"must be specified")
+	}
+
+	if feeRate < chainfee.FeePerKwFloor {
+		return nil, fmt.Errorf("fee rate of %d sat/kw is too low, "+
```
guggero, line 546: I think we can just set the fee rate to the fee flow here, just in case the fee estimation ever returns a value too small.

account/manager.go
guggero: Or perhaps we should extract the call to `m.cfg.Wallet.EstimateFee()` to just before this call so we assume the fee rate is always non-zero in `resumeAccount()`?

account/manager.go
```diff
@@ -480,14 +459,16 @@ func (m *Manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
 		if createTx {
-			// We need a static sat/vByte value for the fee in
-			// SendOutputs, so we use the stored targetConf of the
-			// account to estimate it.
-			feeSatPerKw, err := m.cfg.Wallet.EstimateFee(
-				ctx, int32(fundingConfTarget),
-			)
-			if err != nil {
-				return err
+			if feeRate == 0 {
```
guggero, line 462: See my comment above about removing this `if` by moving the fee estimation outside of this method.

account/manager_test.go
```diff
@@ -531,8 +531,8 @@ func TestResumeAccountAfterRestart(t *testing.T) {
 	// persisted intent to create the account.
 	go func() {
 		_, _ = h.manager.InitAccount(
-			context.Background(), value, expiry, bestHeight,
-			DefaultFundingConfTarget,
+			context.Background(), value, chainfee.FeePerKwFloor, expiry,
```
guggero, line 534: nit: should probably be moved to the previous commit.

cmd/pool/account.go
```diff
@@ -101,6 +101,11 @@ var newAccountCommand = cli.Command{
 				"within",
 			Value: defaultFundingConfTarget,
 		},
+		cli.Uint64Flag{
+			Name: "sat_per_vbyte",
+			Usage: "the fee rate expressed in sat/vbyte that " +
+				"should be used for the renewal transaction",
```
guggero, line 107: nit: s/renewal transaction/account creation transaction/?

cmd/pool/account.go
```diff
@@ -119,10 +124,30 @@ func newAccount(ctx *cli.Context) error {
...
 		AccountValue: amt,
-		Fees: &poolrpc.InitAccountRequest_ConfTarget{
+		Initiator:    defaultInitiator,
+	}
+
+	satPerVByte := ctx.Uint64("sat_per_vbyte")
+	confTarget := ctx.Uint64("conf_target")
+
+	switch {
+	case satPerVByte != 0:
```
guggero, line 134: nit: same here re: `> 0` vs. `!= 0`

cmd/pool/account.go
```diff
@@ -172,30 +199,42 @@ func newAccount(ctx *cli.Context) error {
 	return nil
 }
 
-func printAccountFees(client poolrpc.TraderClient, amt btcutil.Amount,
-	confTarget uint32) error {
+func printAccountFees(
```
guggero, line 202: nit: we don't usually use this more JS type formatting but instead use the full line width for the function header.
Also, after a multi-line function definition, we add a newline to better show where the body begins.

cmd/pool/account.go
```diff
@@ -172,30 +199,42 @@ func newAccount(ctx *cli.Context) error {
...
+	client poolrpc.TraderClient,
+	amt btcutil.Amount,
+	satPerVByte uint64,
+	confTarget uint32,
+) error {
+	if confTarget != 0 {
+		resp, err := client.QuoteAccount(context.Background(),
```
guggero, line 209: nit: arguments should be on new line for multi-line function call.

cmd/pool/account.go
```diff
@@ -172,30 +199,42 @@ func newAccount(ctx *cli.Context) error {
...
+		return nil
+	}
 
 	fmt.Println("-- Account Funding Details --")
 	fmt.Printf("Amount: %v\n", amt)
-	fmt.Printf("Confirmation target: %v blocks\n", confTarget)
-	fmt.Printf("Fee rate (estimated): %.1f sat/vByte\n", satPerVByte)
-	fmt.Printf("Total miner fee (estimated): %v\n",
-		btcutil.Amount(resp.MinerFeeTotal))
+	// TODO(positiveblue): print Total miner fee (estimated) for the
```
guggero, line 235: I don't think we can really say that in the case here. Because above (with the conf target) we ask the wallet for an estimation, so it knows how many inputs it's going to use and whether there is a change output. But here we don't know anything about the actual transaction that's going to be sent out other than the fee rate.

cmd/pool/account.go
```diff
@@ -119,10 +124,30 @@ func newAccount(ctx *cli.Context) error {
...
+		// Enforce a minimum fee rate of 253 sat/kw by rounding up if 1
+		// sat/byte is used.
+		feeRate := chainfee.SatPerKVByte(satPerVByte * 1000).FeePerKWeight()
+		if feeRate < chainfee.FeePerKwFloor {
+			feeRate = chainfee.FeePerKwFloor
+		}
+		req.Fees = &poolrpc.InitAccountRequest_FeeRateSatPerKw{
+			FeeRateSatPerKw: uint64(feeRate),
+		}
+	case confTarget > 0:
```
guggero, line 144: Because of the default value, the `confTarget` here will always be greater than 0 (unless set explicitly to 0 by the user). You need to use `ctx.IsSet("conf_target")` here.

guggero (commented):
Great work on your first Pool PR :100: 

Mostly nits and suggestions and one problem on the command line. Other than that looks pretty great!

account/manager.go
positiveblue: I will leave the `if feeRate == 0 {` as a sanity check and then return an error that "should never happen"

account/manager_test.go
positiveblue: done

rpcserver.go
positiveblue: I added a check to ensure that if the fee estimation is ever too small we set ti to the FeePerKwFloor value

account/manager.go
```diff
@@ -480,26 +468,23 @@ func (m *Manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
-			feeSatPerKw, err := m.cfg.Wallet.EstimateFee(
-				ctx, int32(fundingConfTarget),
-			)
-			if err != nil {
-				return err
+			acctKey := account.TraderKey.PubKey.SerializeCompressed()
+
+			if feeRate == 0 {
+				return fmt.Errorf("unable to create transaction for "+
+					"account with trader key %x, feeRate should be greater "+
```
guggero, line 475: nit: wrap at 80 characters. You might need to update your editor's config to use 8 spaces for a tab character instead of just 4 or 2 to match everyone else's view of what 80 characters width is :)

account/manager.go
```diff
@@ -389,6 +375,8 @@ func (m *Manager) WatchMatchedAccounts(ctx context.Context,
 		// closed because it was used up or pending batch update because
 		// it was recreated. Either way, let's resume it now by creating
 		// the appropriate watchers again.
+		// We set feerate to 0 because we know that we won't need to
+		// create a new transaction for resuming the account
```
guggero, line 379: nit: missing full stop after sentence.

rpcserver.go
guggero: Hmm, now we have the same check twice with different behavior. I think we should either remove this check down here or move the one above to here (meaning: replace the error with just setting it to the fee floor).

cmd/pool/account.go
```diff
@@ -173,29 +207,37 @@ func newAccount(ctx *cli.Context) error {
...
+
+	if confTarget != 0 {
+		resp, err := client.QuoteAccount(
+			context.Background(),
+			&poolrpc.QuoteAccountRequest{
+				AccountValue: uint64(amt),
+				Fees: &poolrpc.QuoteAccountRequest_ConfTarget{
+					ConfTarget: confTarget,
+				},
+			})
```
guggero, line 220: nit: closing parenthesis of multiline argument list should be on new line.

cmd/pool/account.go
```diff
@@ -173,29 +207,37 @@ func newAccount(ctx *cli.Context) error {
 }
 
 func printAccountFees(client poolrpc.TraderClient, amt btcutil.Amount,
-	confTarget uint32) error {
+	satPerVByte uint64, confTarget uint32) error {
+
+	if confTarget != 0 {
```
guggero, line 212: nit: If we used `if confTarget == 0 {` we could have the smaller code block indented and have the larger block below without the additional level of indentation.

guggero (approved):
Nice work, LGTM :tada: 
Just a few minor nits remaining.

account/manager.go
```diff
@@ -188,16 +188,21 @@ func (m *Manager) start() error {
 	if err != nil {
 		return fmt.Errorf("unable to retrieve accounts: %v", err)
 	}
+
+	// We calculate the default fee rate that will be used
+	// for resuming accounts for which we haven't created and broadcast
+	// a transaction for yet
```
bhandras, line 194: nit: not 100% sure but I think there's an extra "for" here.

account/manager.go
```diff
@@ -188,16 +188,21 @@ func (m *Manager) start() error {
...
+	feeRate, err := m.cfg.Wallet.EstimateFee(
+		ctx, int32(DefaultFundingConfTarget),
+	)
+	if err != nil {
+		return fmt.Errorf("unable to estimate default fees for resuming"+
```
bhandras, line 199: nit: add a space at the end after `resuming` as otherwise it'll concatenate with the next line. Also I think the error itself could just be: "unable to estimate default fees". I personally like long errors but overall we tend to not overexplain error cases and try to reduce to the actual cause.

rpcserver.go
```diff
@@ -510,17 +510,53 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 			"must be specified")
 	}
 
-	// Determine the desired transaction fee.
-	confTarget := req.GetConfTarget()
-	if confTarget < 1 {
-		return nil, fmt.Errorf("confirmation target must be " +
-			"greater than 0")
+	if req.GetFeeRateSatPerKw() > 0 && req.GetConfTarget() > 0 {
+		return nil, fmt.Errorf("you must set only one of the sat/vbyte " +
```
bhandras, line 514: nit: in the error message I think you wanted to write sats/kw

rpcserver.go
bhandras: not compelte but here's a generic guide to explain our coding style: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#code-spacing

rpcserver.go
```diff
@@ -510,17 +510,53 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
+		// we set it to a valid minimum
+		if feeRate < chainfee.FeePerKwFloor {
+			feeRate = chainfee.FeePerKwFloor
+		}
+
+		log.Infof("Estimated total chain fee of %v for new account with "+
+			"value=%v, conf_target=%v", feeRate, value, confTarget)
+
+	default:
+		return nil, fmt.Errorf("either sat/vbyte or confirmation target " +
```
bhandras, line 547: sats/kw ?

bhandras: Maybe I'd move the error case from the above `if` statement here too as a case.

bhandras (commented):
Great work! Changes LGMT :hamburger:, added a few small nits + the ones Oli already suggested.

bhandras (approved):
LGTM, great work on this PR! :tada:

rpcserver.go
positiveblue: This one is tricky because I would expect the caller to receive an error if they set a too low fee but in the case of specifying a `conf_target` we ensure that the user does not get that error (ensuring that we set it to,at least the minimum valid value)

account/manager.go
positiveblue: I wonder if we can use a lint step for this, like `lll` or something

rpcserver.go
positiveblue: Let me add another small PR with this fix + adding the same check for `expiryHeight`

cmd/pool/account.go
```diff
@@ -173,27 +207,36 @@ func newAccount(ctx *cli.Context) error {
...
+	)
 	if err != nil {
 		return fmt.Errorf("unable to estimate on-chain fees: "+
 			"%v", err)
 	}
 
 	feeRate := chainfee.SatPerKWeight(resp.MinerFeeRateSatPerKw)
-	satPerVByte := float64(feeRate.FeePerKVByte()) / 1000
-
+	feePerKVByte := float64(feeRate.FeePerKVByte()) / 1000
```
guggero, line 235: nit: we still divide by 1000 so it's `feePerVByte`. Meaning, we can just overwrite the value of the existing variable here.

rpcserver.go
```diff
@@ -510,17 +510,53 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 			"must be specified")
 	}
 
-	// Determine the desired transaction fee.
-	confTarget := req.GetConfTarget()
-	if confTarget < 1 {
-		return nil, fmt.Errorf("confirmation target must be " +
-			"greater than 0")
+	if req.GetFeeRateSatPerKw() > 0 && req.GetConfTarget() > 0 {
```
guggero, line 513: nit: this could just be another `case` in the `switch` statement below.

guggero (approved):
Two last small nits. Feel free to merge any time.

account/manager.go
guggero: The problem is, some lines we can't really break nicely (for example the config with the tagged struct values). So we'd get a lot of false positives. And adding ` // nolint:lll` to each of the long lines just makes them even longer :sweat_smile:

PR #291, opened 2021-10-04
account: handle invalid InitAccount parameters

positiveblue:
Add check for requests with relative AND absolute height provided
instead of using the absolute one by default.

The same check for sats/kw and confirmation target has been moved to the
`switch` statement instead of having it in an extra `if`.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -499,6 +499,11 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 	// Determine the desired expiration value, can be relative or absolute.
 	var expiryHeight uint32
 	switch {
+
+	case req.GetAbsoluteHeight() != 0 && req.GetRelativeHeight() != 0:
+		return nil, fmt.Errorf("you must set only one of the relative " +
+			"and absolute height parameters")
+
 	case req.GetAbsoluteHeight() != 0:
 		expiryHeight = req.GetAbsoluteHeight()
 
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -499,6 +499,9 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 	// Determine the desired expiration value, can be relative or absolute.
 	var expiryHeight uint32
 	switch {
+	case req.GetAbsoluteHeight() != 0 && req.GetRelativeHeight() != 0:
+		return nil, fmt.Errorf("you must set only one of the relative " +
+			"and absolute height parameters")
```
guggero, line 504: nit: add newline before new `case` block here and above.

guggero (approved):
Nice fix, LGTM :tada:

rpcserver.go
```diff
@@ -510,14 +513,12 @@ func (s *rpcServer) InitAccount(ctx context.Context,
...
-		return nil, fmt.Errorf("you must set only one of the sats/kw " +
-			"and confirmation target parameters")
-	}
-
 	var feeRate chainfee.SatPerKWeight
 
 	switch {
+	case req.GetFeeRateSatPerKw() > 0 && req.GetConfTarget() > 0:
+		return nil, fmt.Errorf("you must set only one of the sats/kw " +
+			"and confirmation target parameters")
```
bhandras, line 521: nit: add a new line below

bhandras (approved):
LGTM :+1:

PR #292, opened 2021-10-04
rpcserver: fix feeRate when using conf target

positiveblue:
The `feeRate` variable was shadowing inside the `req.GetConfTarget() > 0`
switch case. The server was returning an error when the `feeRate` was
estimated using the confirmation target.

```
[ERR] RPCS: [/poolrpc.Trader/InitAccount]: fee rate of 0 sat/kw is too low, minimum is 253 sat/kw
```

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -526,7 +526,8 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 		value := btcutil.Amount(req.AccountValue)
 		confTarget := req.GetConfTarget()
 
-		feeRate, _, err := s.accountManager.QuoteAccount(
+		var err error
+		feeRate, _, err = s.accountManager.QuoteAccount(
 			ctx, value, confTarget,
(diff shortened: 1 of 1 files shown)


rpcserver.go
```diff
@@ -526,17 +526,19 @@ func (s *rpcServer) InitAccount(ctx context.Context,
 		value := btcutil.Amount(req.AccountValue)
 		confTarget := req.GetConfTarget()
 
-		feeRate, _, err := s.accountManager.QuoteAccount(
+		feePerKw, _, err := s.accountManager.QuoteAccount(
```
guggero, line 529: I think the shadowing should be fixed to make the code more easy to read.
So just:
```
var err error
feeRate, _, err = s.accountManager.QuoteAccount(
```

guggero (approved):

bhandras (approved):

PR #298, opened 2021-10-20
account: check valid signer when resuming accounts

positiveblue:
If `poold` is using a different `lndclient.SignerClient` than the one
used during an account creation it won't be able to generate valid
signatures. `poold` wasn't checking the signatures before sending them
to the backend and it was sending invalid
`ClientAuctionMessage_Subscribe` messages.

With this change `poold` will detect invalid signatures before sending
them, instead of waiting for the backend to send back an error.


account/manager.go
```diff
@@ -200,6 +201,16 @@ func (m *Manager) start() error {
 	}
 
 	for _, account := range accounts {
+
```
guggero, line 204: nit: unnecessary empty line.

account/manager.go
```diff
@@ -407,6 +417,28 @@ func (m *Manager) maybeBroadcastTx(ctx context.Context, tx *wire.MsgTx,
...
+	secret, err := m.cfg.Signer.DeriveSharedKey(
+		ctx, account.AuctioneerKey, &account.TraderKey.KeyLocator,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to regenerate secret: %v", err)
+	}
+
+	// Here we would detect if the backend LND node (signer) changed.
+	if !bytes.Equal(secret[:], account.Secret[:]) {
+		return fmt.Errorf("unable to verify secret")
```
guggero, line 436: This error message could probably be a bit more descriptive. For example: "couldn't derive account secret; make sure you are using the same lnd node/seed that was used for creating the account".

guggero (approved):
Very nice fix, should keep our logs from getting spammed! LGTM pending two nits :taco:

Issue #295, opened 2021-10-10
Getting an End of File error when registering a sidecar ticket


gkrizek:
_this is from master, commit `9e5e1a8c52afa977e8f6b2e184b7ed0a6739f0a7`_

My pool daemon is frequently getting into a weird state where registering a sidecar channel gives an EOF error. This appears to be related to a sidecar channel being cancelled from the person that created it.

**Expected behavior**

I can continue to claim sidecar tickets even if others are cancelled or problematic.

**Actual behavior**

If the person that created a sidecar ticket cancels the ticket, then my pool installation that I used to redeem the cancelled ticket goes into a states where it can't connect to the auction server for new or existing sidecar tickets.

Logs from a node claiming the ticket:

```
2021-10-09 19:11:52.071 [INF] POOL: Attempting negotiation to receive sidecar ticket: 565dbcd45c81a3e5
2021-10-09 19:11:52.071 [INF] POOL: Creating receiver reply mailbox for ticket=565dbcd45c81a3e5, stream_id=414157fbf23ca1f77b698570bef1b87b5576b9d9a9b1bd9900df57e0249012bfc3e534f8b1c510489bac6ad02489cbd829d9747250de724c8cf490118924ddbf
2021-10-09 19:11:52.165 [INF] POOL: Transmitting registered ticket=565dbcd45c81a3e5 to provider
2021-10-09 19:11:52.165 [INF] POOL: Waiting for ticket (id=565dbcd45c81a3e5) using stream_id=414157fbf23ca1f77b698570bef1b87b5576b9d9a9b1bd9900df57e0249012bfc3e534f8b1c510489bac6ad02489cbd829d9747250de724c8cf490118924ddbf, provider=false
2021-10-09 19:11:52.165 [INF] POOL: Sending ticket(state=registered, id=565dbcd45c81a3e5) to provider stream_id=a2a241366bead79455c94ec58682cba07e58d375515852a6b480bf1bba666db8475712a8b9f07a70d78ff565b86ccb258a2dafd956151eedb7cf723e548d86aa
2021-10-09 19:11:52.334 [INF] POOL: Receive new message for ticket (id=565dbcd45c81a3e5) via stream_id=414157fbf23ca1f77b698570bef1b87b5576b9d9a9b1bd9900df57e0249012bfc3e534f8b1c510489bac6ad02489cbd829d9747250de724c8cf490118924ddbf, provider=false
[...]

Issue #294, opened 2021-10-10
When a sidecar ticket is cancelled, the person claiming never gets updated.


gkrizek:
If you register a sidecar ticket, but the person that made the ticket cancels the order, the recipient's sidecar is still expecting a channel.

**Expected behavior**

When I have a sidecar ticket registered and the creator cancels the order, I expect my pool instance to know this happened and update the ticket status according.

**Actual behavior**

When I have a sidecar tickets registered and the person that created it cancels the order, my `pool sidecar list` command still shows the sidecar ticket as `expecting channel`. I would instead like it to transition to `cancelled`.  I also see this in my receipint pool logs after the order is cancelled:

```
2021-10-09 19:01:21.805 [ERR] RPCS: [/poolrpc.Trader/CancelSidecar]: error canceling order 961526b127b69993ab7a51ae59bad2b0b0c63870a167f08246d75ffeaa401d2d for sidecar ticket cf59074b002a81d8: no order found
```

**To reproduce**

- Create a sidecar ticket with Pool A
- Redeem the sidecar ticket with Pool B
- Cancel the sidecar ticket on Pool A

**System information**

macOS
`poold version 0.5.1-alpha commit=v0.5.1-alpha-24-g9e5e1a8c52afa977e8f6b2e184b7ed0a6739f0a7`

PR #304, opened 2021-10-27
sidecar: inform remote side about cancellation

guggero:
Fixes https://github.com/lightninglabs/pool/issues/295.
Fixes https://github.com/lightninglabs/pool/issues/294.

This PR fixes the issue that when one side of a sidecar channel offering cancels the ticket, the other side isn't informed about it. This leads to follow-up errors on restarts and subsequent ticket submissions.

We solve this by sending a message to the other side if we are the cancelling party.
Because we send the message over the mailbox, we cannot remove it in that case.

### DIFF
--- a/auto_sidecar.go
+++ b/auto_sidecar.go
@@ -553,7 +553,8 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
 		case newTicket := <-packetChan:
 			// The provider has more states it needs to transition
 			// through, so we'll continue until we end up at the
-			// same state (a noop)
+			// same state (a noop).
+		stateUpdateLoop:
 			for {
 				priorState := sidecar.State(atomic.LoadUint32(&a.currentState))
 
@@ -575,9 +576,9 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
 
 				switch {
 				case priorState == newPktState.CurrentState:
-					fallthrough
+					break stateUpdateLoop
 				case newPktState.CurrentState == sidecar.StateExpectingChannel:
-					break
+					break stateUpdateLoop
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


auto_sidecar.go
```diff
@@ -575,9 +576,9 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
 
 				switch {
 				case priorState == newPktState.CurrentState:
-					fallthrough
+					break stateUpdateLoop
```
Roasbeef, line 637: Ahh, solid catch!

auto_sidecar.go
```diff
@@ -579,30 +637,67 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
...
+				err := a.cfg.MailBox.SendSidecarPkt(
+					ctxb, localTicket, false,
+				)
+				if err != nil {
+					log.Errorf("unable to send cancel "+
+						"msg to recipient: %v", err)
+					return
+				}
+
+			// In every other case we can just remote the mailbox:
```
Roasbeef, line 686: remote -> remove

auto_sidecar.go
```diff
@@ -579,30 +637,67 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
...
+				// a background context here.
+				ctxb := context.Background()
+				err := a.cfg.MailBox.SendSidecarPkt(
+					ctxb, localTicket, false,
+				)
+				if err != nil {
+					log.Errorf("unable to send cancel "+
+						"msg to recipient: %v", err)
+					return
+				}
```
Roasbeef, line 684: I think we should fully tear down the goroutine here since at this point all negotiation has ceased. Same for the opposite role as well.

Roasbeef (commented):
Nice solution! Did an initial pass, need to do another to ensure I've wrapped my head around how this handles some of the disconnect/retransmission edge cases.

Roasbeef:
Gonna land this so we can proceed with some of the dependent merges, ideally can address those lingering commits in a follow up change

PR #305, opened 2021-10-27
multi: extend account expiry after using it

positiveblue:
Follows the same format than in `RenewAccount`: one of `absolute_expiry` or `relative_expiry`.

`--expiry_height` or `--expiry_blocks` in `pool account`


cmd/pool/account.go
```diff
@@ -417,6 +418,17 @@ var depositAccountCommand = cli.Command{
 			Usage: "the fee rate expressed in sat/vbyte that " +
 				"should be used for the deposit",
 		},
+		cli.Uint64Flag{
+			Name: accountExpiryAbsolute,
+			Usage: "the new block height at which this account " +
```
guggero, line 423: nit: remove first occurrence of "at"?

guggero: Could also define the whole flag as a variable and then use it twice?

rpcserver.go
```diff
@@ -746,11 +746,22 @@ func (s *rpcServer) DepositAccount(ctx context.Context,
 			"minimum is %d sat/kw", feeRate, chainfee.FeePerKwFloor)
 	}
 
+	bestHeight := atomic.LoadUint32(&s.bestHeight)
+
+	// If provided, determine new expiration value.
+	var expiryHeight uint32
+	switch {
+	case req.GetAbsoluteExpiry() != 0:
```
guggero, line 754: We also need to add the check for both being set, in case someone uses the API directly. Same for the withdrawal.

account/manager_test.go
```diff
@@ -869,7 +869,7 @@ func TestAccountWithdrawal(t *testing.T) {
 	dustOutput := &wire.TxOut{Value: 0, PkScript: p2wsh}
 	_, _, err := h.manager.WithdrawAccount(
 		context.Background(), account.TraderKey.PubKey,
-		[]*wire.TxOut{dustOutput}, feeRate, bestHeight,
+		[]*wire.TxOut{dustOutput}, feeRate, bestHeight, 0,
```
guggero, line 872: Should we also add a test case for a non-zero expiration that checks the account modifications are added correctly?

guggero (commented):
Nice work, was able to test this locally and it all worked great!
Only a few minor comments from my end. The only one blocking IMO is the unit test case.
Other than that LGTM :tada:

account/manager_test.go
positiveblue: totally agree. The tests are currently pretty coupled to mock/hard coded values so let's add it after #313 has been merged and we can manage the expected behavior better.

cmd/pool/account.go
positiveblue: done (it was shared with other cmds too)

cmd/pool/account.go
```diff
@@ -417,6 +418,13 @@ var depositAccountCommand = cli.Command{
 			Usage: "the fee rate expressed in sat/vbyte that " +
 				"should be used for the deposit",
 		},
+		expiryAbsoluteFlag,
+		cli.Uint64Flag{
```
guggero, line 422: nit: extract this flag into a variable as well?

guggero (approved):
Nice work, LGTM :tada:

cmd/pool/account.go
positiveblue: I thought about it but one out of the three usages has a default value set and the other two don't... we can have a separated PR consolidating the flags later.

rpcserver.go
```diff
@@ -746,11 +746,25 @@ func (s *rpcServer) DepositAccount(ctx context.Context,
 			"minimum is %d sat/kw", feeRate, chainfee.FeePerKwFloor)
 	}
 
+	bestHeight := atomic.LoadUint32(&s.bestHeight)
+
+	// If provided, determine new expiration value.
+	var expiryHeight uint32
+	switch {
+	case req.GetAbsoluteExpiry() != 0 && req.GetRelativeExpiry() != 0:
```
bhandras, line 754: nit: is this case even possible? (given it's a `oneof` in the proto).

bhandras (approved):
LGTM :+1:

rpcserver.go
guggero: Ah, you're right, it can't really. Forgot it was a `oneof` when I asked for this case to be added. But doesn't hurt either, so we can just leave it IMO.

PR #307, opened 2021-10-29
CI: switch from Travis to GitHub Actions

guggero:
Removes Travis and adds GitHub Actions as CI instead.

An example run of this code can be found here: https://github.com/guggero/pool/runs/4048500708?check_suite_focus=true

#### Pull Request Checklist
  used.


guggero:
> BTW @guggero shouldn't this branch have some checks already running?

Unfortunately not, they only run once they're in master (security precaution).

PR #311, opened 2021-11-04
pool: conditionally create default macaroon file

guggero:
In some cases we don't want the default macaroon file to be created on
disk, so we allow passing in a boolean that toggles the macaroon
creation.

This is for the LiT stateless integrated mode.

#### Pull Request Checklist
  used.


guggero:
Replaces https://github.com/lightninglabs/pool/pull/272.

bhandras (approved):

positiveblue (approved):

PR #313, opened 2021-11-08
mock: Add mockgen workflow

positiveblue:
- `make mock`: generates all the mocks for testing.
- `mock-check`: generates all the mocks + checks if any *.go file changed.
- `make gen`: generates all the auto generated files (protos + mocks).
- Add `mock-check` to our CI

Create mocks for `sidecar/interfaces.go` and `internal/test/interfaces.go` and use them for rewriting the `TestRegisterSidecar`.


sidecar_acceptor_test.go
```diff
@@ -32,67 +31,46 @@ var (
...
-	acceptor := NewSidecarAcceptor(&SidecarAcceptorConfig{
-		SidecarDB:  nil,
-		AcctDB:     nil,
-		Signer:     mockSigner,
-		Wallet:     mockWallet,
-		NodePubKey: ourNodePubKey,
-		ClientCfg:  auctioneer.Config{},
-	})
+	emptySetter := func(
+		ticket *sidecar.Ticket, sc *test.MockSignerClient,
```
guggero, line 40: I think we don't actually need the `ticket` in the setter at all.

sidecar_acceptor_test.go
```diff
@@ -32,67 +31,46 @@ var (
...
 		ticket: &sidecar.Ticket{
-			ID:    existingTicket.ID,
 			State: sidecar.StateOffered,
-			Offer: sidecar.Offer{
-				SignPubKey:     providerPubKey,
-				SigOfferDigest: testOfferSig,
-			},
 		},
-		expectedErr: "ticket with ID",
+		expectedErr: "error verifying sidecar offer",
```
guggero, line 68: We should actually check that the error contains `offer in ticket is not signed` here.

sidecar_acceptor_test.go
```diff
@@ -102,9 +80,23 @@ func TestRegisterSidecar(t *testing.T) {
 				},
 			},
 		},
-		expectedErr: "signature not valid for public key",
+		expectedErr: "error verifying sidecar offer",
```
guggero, line 83: We should actually check that the error contains `signature not valid for public key` here.

sidecar_acceptor_test.go
```diff
@@ -32,67 +31,46 @@ var (
...
 		expectedErr string
-		check       func(t *testing.T, ticket *sidecar.Ticket)
+		mockSetter  func(
+			ticket *sidecar.Ticket, sc *test.MockSignerClient,
+			wc *test.MockWalletKitClient, store *sidecar.MockStore,
+		)
+		check func(t *testing.T, ticket *sidecar.Ticket)
 	}{{
-		name: "ticket with ID exists",
+		name:        "unable to register sidecar if ticket does not have a valid state",
```
guggero, line 58: A way to remove one level of indentation would be to define these test cases globally and not within the test function.

guggero (commented):
Very cool to see the `gomock` package in action!
I really like fact that we can generate very dynamic mocks on demand. And the way to use them also looks pretty easy to get started with.

This example use case also shows nicely why it's great to work with mocks: one can more easily test all possible branches as you demonstrated by adding two additional test cases.

IMO we should have a discussion about using `gomock` in lnd as well and take this PR as a base for the discussion.

sidecar_acceptor_test.go
positiveblue: Totally right, I put it there at the beginning when I though that the mocks needed take thinks like the ID but they don't :)

sidecar_acceptor_test.go
positiveblue: yes, better be as specific as we can and if we change the message we change the test than to take the "higher level" msg (Y)

sidecar_acceptor_test.go
positiveblue: Done, it is quite hard to keep  < 80 columns in table driven tests, I promise I will do my best :sweat_smile:

yyforyongyu (commented):
Glad to see another mock in action!

Imo mock is a necessary component that's missing from our unit tests. It makes tests much easier and, in particular, since we have many goroutines flying around, I find it very helpful to validate whether a call does happen inside another goroutine.

As mentioned before, `lnd` and `naut` have already started using mocks (thanks to @bhandras!). The challenge I see for `lnd` is that it's missing interfaces, where it makes the mocking impossible. There is an ongoing effort made by @Crypt-iQ where we want to build more interfaces inside [lnd's `peer`](https://github.com/lightningnetwork/lnd/pull/5067). And we def need more to make the mocking happen.

That being said, what's your opinion about the two tools, `go/mock` and `testify/mock`? Curious about your preference here. I chose `testify/mock` because we were already using `testify` inside `lnd` and from [this article](https://blog.codecentric.de/2019/07/gomock-vs-testify/). Tbh I don't have that much experience using mocks inside go so that choice is more or less like, we just need A mock😂

Crypt-iQ:
> As mentioned before, `lnd` and `naut` have already started using mocks (thanks to @bhandras!). The challenge I see for `lnd` is that it's missing interfaces, where it makes the mocking impossible. There is an ongoing effort made by @Crypt-iQ where we want to build more interfaces inside [lnd's `peer`](https://github.com/lightningnetwork/lnd/pull/5067). And we def need more to make the mocking happen.

When we get native fuzzing in golang, mocks will be very helpful. Since I imagine the golang fuzzer is coverage-driven, it is good to separate components and make them "dumb" to not confuse the coverage signals (e.g. so they don't spawn goroutines, have nondeterministic behavior, etc). Also if you want to run a lot of fuzzers, you don't want something like an `invoiceregistry` or `htlcswitch` in lnd on each fuzzing instance if it's just testing round-trip parsing of a particular component, so mocking helps there too.

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,199 @@ var (
...
+		},
+	},
+	expectedErr: "signature not valid for public key",
+	mockSetter: func(
+		sc *test.MockSignerClient, wc *test.MockWalletKitClient,
+		store *sidecar.MockStore,
+	) {
+		sc.EXPECT().
+			VerifyMessage(
+				gomock.Any(),
```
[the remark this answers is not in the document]
positiveblue: That's why I added the ticket to the mockSetter @guggero

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,215 @@ var (
...
-func newTestDB(t *testing.T) (*clientdb.DB, func()) {
-	tempDir, err := ioutil.TempDir("", "client-db")
-	require.NoError(t, err)
-
-	db, err := clientdb.New(tempDir, clientdb.DBFilename)
-	if err != nil {
-		require.NoError(t, os.RemoveAll(tempDir))
-		t.Fatalf("unable to create new db: %v", err)
-	}
+var registerSidecarEmptySetter = func(
```
guggero, line 34: double nit: could just be an empty func, no need to declare it as a `var`. Also, the first argument should be on the same line if it fits, otherwise this looks like a function _call_ instead of a _definition_ at first glance.

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,215 @@ var (
...
+	ticket *sidecar.Ticket, sc *test.MockSignerClient,
+	wc *test.MockWalletKitClient, store *sidecar.MockStore,
+) {
+}
 
-	return db, func() {
-		require.NoError(t, db.Close())
-		require.NoError(t, os.RemoveAll(tempDir))
-	}
+var registerSidecarEmptyCheck = func(t *testing.T, ticket *sidecar.Ticket) {
```
guggero, line 40: nit: same here, can just be a func.

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,215 @@ var (
...
+	offerDigest, _ := ticket.OfferDigest()
+	sigOfferDigest := ticket.Offer.SigOfferDigest.Serialize()
 
-	existingTicket, err := sidecar.NewTicket(
-		sidecar.VersionDefault, 1_000_000, 200_000, 2016,
-		providerPubKey, false,
+	return offerDigest[:], sigOfferDigest, offerPubKeyRaw
+}
+
+var registerSidecarTestCases = []struct {
```
guggero, line 54: :+1:

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,215 @@ var (
...
-			Offer: sidecar.Offer{
-				SignPubKey: providerPubKey,
-				SigOfferDigest: &btcec.Signature{
-					R: new(big.Int).SetInt64(33),
-					S: new(big.Int).SetInt64(33),
-				},
-			},
+	},
+	expectedErr: "signature not valid for public key",
+	mockSetter: func(
```
guggero, line 91: nit: same here re function call vs. definition formatting.

sidecar_acceptor_test.go
```diff
@@ -32,136 +31,215 @@ var (
...
+		ticket *sidecar.Ticket, sc *test.MockSignerClient,
+		wc *test.MockWalletKitClient, store *sidecar.MockStore,
+	) {
+		p1, p2, p3 := getVerifyParameters(ticket)
+
+		sc.EXPECT().
+			VerifyMessage(
+				gomock.Any(),
+				p1,
```
guggero, line 159: Nice!

guggero (approved):
LGTM, just a few nits :tada: 

Sorry if I'm being extra nit-picky, but if this is going to be our template for mock-usage, things should look extra tidy because they are going to be copy/pasted a lot :sweat_smile:

yyforyongyu (approved):
Go mock go!

guggero:
Just waiting for @Roasbeef's opinion on using generated mocks, otherwise this is ready to be merged.

PR #316, opened 2021-11-21
refactor `account/watcher` and add mocks/tests

positiveblue:
- Make `account/manager` take an interface for the watcher instead of a explicit type.
- Split the watcher in two components:
    - Controller: initialize/delete logic + dispatch handlers
    - Watcher: watcher state + handle logic
- Add tests for controller + watcher


account/watcher/watcher.go
```diff
@@ -0,0 +1,165 @@
+package watcher
+
+import (
+	"sync"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/chainntnfs"
+)
+
+type WatcherInterface interface { // nolint:golint
```
guggero: Same here re naming: Just call this `Watcher` and the implementation `watcher`?
Or, together with the other suggestion of extracting the `Handle*` methods into their own interface, this would only handle account expirations. So it could also be called `ExpiryWatcher`?

guggero: Also: missing godoc and move to `interfaces.go` directly?

account/watcher/controller_test.go
```diff
@@ -0,0 +1,363 @@
...
+		blockChan := make(chan int32)
+		errChan := make(chan error)
+		mockChainNotifier.EXPECT().
+			RegisterBlockEpochNtfn(gomock.Any()).
+			Return(blockChan, errChan, nil)
+
+	},
+	expectedError: "",
+},
+	{
```
guggero, line 45: nit: put opening bracket on previous line to remove extra indentation.

account/watcher/controller_test.go
```diff
@@ -0,0 +1,363 @@
...
+)
+
+var (
+	errCtrlExpected = errors.New("random error")
+)
+
+var controllerLifeCycleTestCases = []struct {
+	name       string
+	mockSetter func(
+		mockChainNotifier *test.MockChainNotifierClient,
```
guggero, line 27: nit: function definition can fit on one line to increase readability. Same below in the implementations.

guggero (commented):
Very nice to see the new mocks in action. The refactor looks pretty good to me, just a few suggestions for the naming and how to simplify it even more.

account/watcher/controller.go
```diff
@@ -67,8 +66,8 @@ type Watcher struct {
 }
 
 // New instantiates a new chain watcher backed by the given config.
-func New(cfg *Config) *Watcher {
-	return &Watcher{
+func NewController(cfg *Config) *controller { // nolint:golint
```
guggero: Ah, I guess that's one downside of having a non-exported implementation. But I think that's okay. Or we could break the "return structs" philosophy here and return the interface instead of silencing the linter. Either way is fine with me, your call.

account/watcher/interfaces.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// Controller is the interface used by other components to communicate with the
+// watcher.
+
```
guggero: nit: extra newline.

guggero: Ah, it's removed in a later commit. Can move it here.

account/manager.go
```diff
@@ -153,13 +153,17 @@ func NewManager(cfg *ManagerConfig) *Manager {
 		quit: make(chan struct{}),
 	}
 
-	m.watcherCtrl = watcher.NewController(&watcher.Config{
-		ChainNotifier:       cfg.ChainNotifier,
+	eventHandler := watcher.NewEventHandler(&watcher.Handlers{
```
guggero, line 156: I think we should just export those `handle*` methods on the manager and just pass `m` here instead of adding this wrapper struct.

account/watcher/watcher.go
```diff
@@ -35,6 +35,9 @@ func NewExpiryWatcher(handlers EventHandler) *expiryWatcher { // nolint:golint
 
 // NewBlock updates the current bestHeight.
 func (w *expiryWatcher) NewBlock(bestHeight uint32) {
+	w.expirationsMtx.Lock()
+	defer w.expirationsMtx.Unlock()
```
guggero, line 39: Nice find!

guggero (changes requested):
Looking good! Way better with the `EventHandler` interface IMO. See my comment about getting rid of the `Handlers` struct as well.
After that we're good to go :tada:

positiveblue:
@guggero I don't know why I cannot reply under every comment so I will reply to all of them here:

> one downside of having a non-exported implementation
yes, I would avoid both: return a non exported struct type and return an interface. However, given that this is only used in `account/manager` having it non exported is not a big problem.

> export those handle* methods on the manager and just pass m here instead of adding this wrapper struct.
(Y)

account/watcher/controller.go
```diff
@@ -67,8 +66,8 @@ type Watcher struct {
...
+	return &controller{
```
guggero, line 48: nit: Godoc comment doesn't match.

account/manager.go
```diff
@@ -919,7 +918,7 @@ func (m *Manager) handleAccountConf(traderKey *btcec.PublicKey,
 // only track the spend of the latest batch, after it confirmed. So the account
 // output in the spend transaction should always match our database state if
 // it was a cooperative spend.
-func (m *Manager) handleAccountSpend(traderKey *btcec.PublicKey,
+func (m *Manager) HandleAccountSpend(traderKey *btcec.PublicKey,
```
guggero, line 918: nit: Godoc comment doesn't match. Same below.

account/watcher/controller.go
```diff
@@ -26,33 +26,23 @@ type expiryReq struct {
 
 // Config contains all of the Watcher's dependencies in order to carry out its
 // duties.
-type Config struct {
+type CtrlConfig struct {
```
guggero, line 18: nit: Godoc comment doesn't match, same for `NewController`.

guggero (approved):
Great initiative on this :100: Some final nits, otherwise LGTM :tada:

account/watcher/watcher.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// NewBlock updates the current bestHeight.
+func (w *expiryWatcher) NewBlock(bestHeight uint32) {
+	w.bestHeight = bestHeight
+}
+
+// OverdueExpirations handles the expirations for the given block.
+func (w *expiryWatcher) OverdueExpirations(blockHeight uint32) {
```
bhandras, line 45: I think a better design could be to have a priority queue of accounts where the priority is defined by the block height. This would allow us to expire everything before and at the specified block height.

account/watcher/watcher.go
```diff
@@ -0,0 +1,103 @@
...
+		// expiration set, then the request was
+		// canceled and there's nothing for us to do.
+		// Similarly, if the request was updated to
+		// track a new height, then we can skip it.
+		curExpiry, ok := w.expirations[accountKey]
+		if !ok || blockHeight != curExpiry {
+			continue
+		}
+
+		err := w.handlers.HandleAccountExpiry(
```
bhandras, line 63: Do we need to lock for `HandleAccountExpiry`? An alternative could be to collect expiring accounts in the lock then expire them outside the lock.

account/watcher/watcher.go
```diff
@@ -0,0 +1,103 @@
...
+	return &expiryWatcher{
+		handlers:             handlers,
+		expirations:          make(map[[33]byte]uint32),
+		expirationsPerHeight: make(map[uint32][]*btcec.PublicKey),
+	}
+}
+
+// NewBlock updates the current bestHeight.
+func (w *expiryWatcher) NewBlock(bestHeight uint32) {
+	w.bestHeight = bestHeight
```
bhandras, line 41: Since expiry happens when the height is changed, it'd be possible to simplify the interface a bit and do the height change and the expiry handling atomically under the same lock. By only expiring the current block if two blocks come right after each other in succession we may miss an expiration (almost no chance in practice but still).

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			RegisterBlockEpochNtfn(gomock.Any()).
+			Return(blockChan, errChan, nil)
+
+	},
+	expectedError: "",
+}, {
+	name: "unable to start watcher because of " +
+		"RegisterBlockEpochNtfn register error",
+	mockSetter: func(mockChainNotifier *test.MockChainNotifierClient) {
+
```
bhandras, line 44: nit: extra nl

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			watcherController.Stop()
+
+			select {
+			case <-watcherController.quit:
+				return
+			case <-time.After(2 * time.Second):
+				t.Error("watcher controller not closed on time")
+			}
+
```
bhandras, line 93: nit: extra nl

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			)
+	},
+	expectedError: errCtrlExpected.Error(),
+}}
+
+func TestWatcherControllerLifeCycle(t *testing.T) {
+	for _, tc := range controllerLifeCycleTestCases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
```
bhandras, line 62: Could run in parallel?

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+var controllerWatchAccountTestCases = []struct {
+	name        string
+	expectedErr string
+}{{
+	name: "Watch account happy path",
+	// TODO (positiveblue): add tests for `cancel` logic
+}}
+
+func TestWatcherControllerWatchAccount(t *testing.T) {
+
```
bhandras, line 169: nit: extra nl

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			confChan := make(chan *chainntnfs.TxConfirmation)
+			errChan := make(chan error)
+
+			doneChan := make(chan error)
```
bhandras, line 197: nit: could just `chan struct{}`

account/watcher/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+				})
+
+			cfg := &CtrlConfig{
+				ChainNotifier: chainNotifier,
+				Handlers:      handlers,
+			}
+
+			watcherController := NewController(cfg)
+
+			_ = watcherController.Start()
```
bhandras, line 324: I think it's worth checking the errors from these functions?

account/watcher/watcher_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+			watcher := NewExpiryWatcher(handlers)
+
+			if len(tc.initialExpirations) > 0 {
+				watcher.expirations = tc.initialExpirations
+			}
+			watcher.bestHeight = tc.bestHeight
+
+			for trader, height := range tc.expirations {
+
```
bhandras, line 235: nit: extra nl

account/watcher/watcher_test.go
```diff
@@ -0,0 +1,255 @@
...
+func TestAddAccountExpiration(t *testing.T) {
+	for _, tc := range addAccountExpirationTestCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			handlers := NewMockEventHandler(mockCtrl)
+
```
bhandras, line 226: nit: extra nl

bhandras (commented):
Nice refactors, and tests! My only major comment is about the expiry watcher, other than that looks good!

account/watcher/watcher.go
bhandras: Or alternatively make newblock/expiry atomic.

positiveblue:
Feedback updated :+1: 

For some of the question: 

> better design could be to have a priority queue of accounts where the priority is defined by the block height

Totally agree. I think it is something we can add later though. Do you know any good implementation that allows update the priority of the elements in it?

> By only expiring the current block if two blocks come right after each other in succession we may miss an expiration (almost no chance in practice but still).

Yes I thought about this but it cannot happen because the controller was calling the two funcs (new block + overdue expirations) before being able to receive another block (iterate the select). I made them atomic, better put it together so we do not get confused and use the API wrong

account/watcher/controller.go
```diff
@@ -47,9 +47,8 @@ type Config struct {
 	HandleAccountExpiry func(*btcec.PublicKey, uint32) error
 }
 
-// Watcher is responsible for the on-chain interaction of an account, whether
-// that is confirmation or spend.
-type Watcher struct {
+// controller implements the Controller interface
+type controller struct {
```
carlaKC, line 28: Add compile time assertion that we satisfy the controller interface?
`var _ Controller = (*controller)(nil)`

account/watcher/controller.go
carlaKC: nit: needs update

account/watcher/interfaces.go
```diff
@@ -44,3 +45,31 @@ type Controller interface {
...
+	// its user sub key (i.e., trader key).
+	HandleAccountSpend(*btcec.PublicKey, *chainntnfs.SpendDetail) error
+
+	// HandleAccountExpiry the operations that should be perform for an
+	// account once it's expired. The account is identified by its user sub
+	// key (i.e., trader key).
+	HandleAccountExpiry(*btcec.PublicKey, uint32) error
+}
+
+// ExpiryWatcher is the interface for the component in charge of the accounts' expiration.
```
carlaKC, line 68: nit: wrap at 80? 
These look a bit over

account/watcher/controller_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+	}
+}
+
+var controllerWatchAccountSpendTestCases = []struct {
+	name        string
+	expectedErr string
+}{{
+	name: "Watch account spend happy path",
+	// TODO (positiveblue): add tests for `cancel` logic
```
carlaKC, line 261: happening in this PR or a follow up?

carlaKC (commented):
Just did a high level pass since there are already 2 reviewers. Great work on interface refactors + mock :)

bhandras (approved):
LGTM :100:

account/watcher/controller.go
```diff
@@ -161,34 +140,7 @@ func (c *controller) expiryHandler(blockChan chan int32, errChan chan error) {
...
-				if err != nil {
-					log.Errorf("Unable to handle "+
-						"expiration of account %x: %v",
-						traderKey.SerializeCompressed(),
-						err)
-				}
-			}
-
-			delete(expirationsPerHeight, bestHeight)
+			c.watcher.NewBlock(uint32(newBlock))
```
bhandras, line 129: One thing to keep in mind is that if `HandleAccountExpiry` blocks for some time, then the `blockChan` might get clogged.

PR #319, opened 2021-11-29
Fix pool CLI short name conflict

### LINKED ISSUE #318: Pool cli shortname collision for "renew" and "recover" account actions
**Expected behavior**

Running `./pool a r` calls account recovery, as `./pool a -h` shows `r` as a viable shortname for recovery

**Actual behavior**
Running `./pool a r` calls account renewal.

it appears this is due to a collision in shortname here:
https://github.com/lightninglabs/pool/blob/acdc235d50bee417e8637872650672a909860cc7/cmd/pool/account.go#L718-L721
https://github.com/lightninglabs/pool/blob/acdc235d50bee417e8637872650672a909860cc7/cmd/pool/account.go#L297-L300

```
$ ./pool a r -h | head -n2
NAME:
   pool accounts renew - renew the expiration of an account
```

**To reproduce**
`./pool a r`

**System information**
N/A

guggero:
Fixes #318.

#### Pull Request Checklist
  used.

### DIFF
--- a/cmd/pool/account.go
+++ b/cmd/pool/account.go
@@ -716,8 +716,7 @@ func bumpAccountFee(ctx *cli.Context) error {
 }
 
 var recoverAccountsCommand = cli.Command{
-	Name:      "recover",
-	ShortName: "r",
+	Name: "recover",
 	Usage: "recover accounts after data loss with the help of the " +
 		"auctioneer",
 	Description: `
--- a/cmd/pool/auction.go
+++ b/cmd/pool/auction.go
@@ -73,7 +73,7 @@ type Snapshot struct {
 	MatchedMarkets         Markets `json:"matched_markets"`
 }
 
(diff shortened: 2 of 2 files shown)


bhandras (approved):
LGTM 🎉

positiveblue (approved):
LGTM, however, given that the cmds are `recover` and `renew` I wonder if it is better to remove both shortcuts so there is no possible confusion.

guggero:
> LGTM, however, given that the cmds are `recover` and `renew` I wonder if it is better to remove both shortcuts so there is no possible confusion.

Good point! I decided to remove the shortcut for `recover` since that has no arguments and can screw up your state if you run it on a non-empty DB (which we should probably fix as well). So it shouldn't have a shortcut in the first place.
The `renew` command has required flags so the user accidentally typing `pool accounts r` shouldn't result in any action directly.

PR #323, opened 2021-12-10
Enable batch participants to sponsor the batch by renewing their account

positiveblue:
Bring back the `sponsor a batch` feature. 

Accounts participating in a batch will be able to sponsor a batch by renewing their account. That applies to all the accounts, included the ones with the `autorenew` batch version, **as long as the batch confirmation does not trigger an expiry extension**.

Added some tests to the `RenewAccount` RPC endpoint. I created a couple of interfaces (Account and Order Managers) to decouple a bit the logic and use mocks for the tests. Same applies to the account rpc marshaler.


rpcserver.go
```diff
@@ -43,6 +43,99 @@ const (
 	getInfoTimeout = 5 * time.Second
 )
 
+// marshalerConfig contains all of the marshaler's dependencies in order to
+// carry out its duties.
+type marshalerConfig struct {
+	// GetOrders returns all orders that are currently known to the store.
+	GetOrders func() ([]order.Order, error)
+	// Terms returns the current dynamic auctioneer terms like max account
```
guggero, line 51: nit: add empty line between elements.

rpcserver.go
```diff
@@ -43,6 +43,99 @@ const (
 	getInfoTimeout = 5 * time.Second
 )
 
+// marshalerConfig contains all of the marshaler's dependencies in order to
+// carry out its duties.
+type marshalerConfig struct {
```
guggero, line 48: nit: move to own file (e.g. `marshaler.go`)?

interfaces.go
```diff
@@ -0,0 +1,16 @@
+package pool
```
guggero, line 1: nit: typo in commit message.

rpcserver.go
```diff
@@ -878,6 +878,21 @@ func (s *rpcServer) WithdrawAccount(ctx context.Context,
 	}, nil
 }
 
+// hasPendingExtension returns if the accountKey is waiting for a batch to be
+// confirmed to extend the expiry height.
+func (s *rpcServer) hasPendingExtension(accountKey *btcec.PublicKey) bool {
+	if s.orderManager.HasPendingBatch() {
```
guggero, line 884: Unfortunately this won't really work. The `orderManager.HasPendingBatch()` only returns true while a current batch is **ongoing** and hasn't been fully signed yet. As soon as the final batch TX is published this will be empty again.

I'm trying to remember, what was the original reason we didn't allow an account extension to be issued by the user if the account was still in the state `StatePendingBatch`? Because it could potentially throw the account out of sync with the server if another update TX was added?
Maybe I gave you some incorrect info there...

guggero (commented):
Nice update to the mocks and unit tests! Great to already see them in action for every code addition :tada: 

I think we need to discuss the batch sponsoring again, it's possible I misunderstood the problem when we added the original restriction. Unfortunately we don't currently have an easy way to find out if the account was extended in the last batch, at least not on the client side.

rpcserver.go
positiveblue: :disappointed: thanks for catching this one

guggero:
!lightninglabs-deploy mute 2022-Feb-01

PR #326, opened 2021-12-14
Add rpcserver test for `RenewAccount` endpoint

positiveblue:
Code extracted from #323 (removing the sponsoring part)


rpcserver_test.go
```diff
@@ -0,0 +1,177 @@
...
+	return req
+}
+
+var renewAccountTestCases = []struct {
+	name          string
+	getReq        func() *poolrpc.RenewAccountRequest
+	checkResponse func(*poolrpc.RenewAccountResponse) error
+	mockSetter    func(*poolrpc.RenewAccountRequest,
+		*account.MockManager, *MockMarshaler,
+	)
```
guggero, line 57: nit: closing parenthesis of method definition should be on same line as last argument (to better distinguish from method call). Same in the actual implementation below.

guggero (approved):
Nice, LGTM :tada:

order/interfaces.go
```diff
@@ -23,6 +23,13 @@ import (
 // Nonce is a 32 byte pseudo randomly generated unique order ID.
 type Nonce [32]byte
 
+// The size of a SHA256 checksum in bytes.
+//
+// Note: this matches the sha256.Size definition. However, mockgen
+// complains about not being able to find the sha256 package. When
+// that bug is fixed, we can change back to [sha256.Size]byte.
+const size = 32
```
bhandras, line 31: nit: `checksumSize`/`hashSize`?

account/manager.go
```diff
@@ -123,7 +123,7 @@ type ManagerConfig struct {
 }
 
 // Manager is responsible for the management of accounts on-chain.
-type Manager struct {
+type manager struct {
```
bhandras, line 126: It's helpful to add a compile time enforcer for the interface:
```
var _ Manager = (*manager)(nil)
```

rpcserver_test.go
```diff
@@ -0,0 +1,177 @@
...
+			srv.bestHeight = 100
+
+			ctx, cancel := context.WithTimeout(
+				context.Background(), ctxTimeout,
+			)
+			defer cancel()
+
+			resp, err := srv.RenewAccount(ctx, req)
+			if tc.expectedError != "" {
+				assert.EqualError(t, err, tc.expectedError)
```
bhandras, line 168: nit: `require` and the expectation comes first iirc

bhandras (approved):
LGTM :hamburger:

rpcserver_test.go
positiveblue: It looks like for `EqualError` this is the right order? 

https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1367

PR #329, opened 2021-12-16
order: set `ExtendAccount` as default batch version

positiveblue:


order/batch.go
```diff
@@ -53,7 +53,7 @@ const (
 
 	// CurrentBatchVersion points to the version used by the client for
 	// verifying a batch.
-	CurrentBatchVersion = DefaultBatchVersion
+	CurrentBatchVersion = ExtendAccountBatchVersion
```
Roasbeef, line 56: Maybe group this in another `const` block so it doesn't look like we're declaring more `BatchVersion`s?

Roasbeef (approved):
LGTM 🦚

PR #330, opened 2021-12-17
multi: remove static default channel types instead select them based on lnd's verison

Roasbeef:
In this PR, we remove the defaults and the CLI and RPC level. We replace them with a new set of functional options that'll let us decide dynamically which features to use based on the lnd version. This lets us support older versions, while being able to activated newer features as we detect them based on what lnd offers. 

#### Pull Request Checklist
- [x] LndServices minimum version has been updated if new lnd apis/fields are
  used.

### DIFF
--- a/order/rpc_parse.go
+++ b/order/rpc_parse.go
@@ -19,10 +19,41 @@ import (
 	"github.com/lightningnetwork/lnd/tor"
 )
 
+// OrderParseOption defines a set of functional param options that can be used
+// to modify our we parse orders based on some optional directives.
+type OrderParseOption func(*parseOptions)
+
+// parseOptions houses the set of functional options used to parse RPC orders.
+type parseOptions struct {
+	chanTypeSelector func() ChannelType
+}
+
+// ChanTypeSelector defines a function capable of selecting a channel type
+// based on the version of an lnd node.
+type ChanTypeSelector func() ChannelType
+
+// WithDefaultChannelType allows a caller to select a default channel type
+// based on the version of the lnd node attempting to create the order.
+func WithDefaultChannelType(selector ChanTypeSelector) OrderParseOption {
+	return func(o *parseOptions) {
+		o.chanTypeSelector = selector
+	}
+}
+
+// defaultParseOptions returns the set of default parse options.
+func defaultParseOptions() *parseOptions {
+	return &parseOptions{}
+}
+
 // ParseRPCOrder parses the incoming raw RPC order into the go native data
 // types used in the order struct.
 func ParseRPCOrder(version, leaseDuration uint32,
-	details *poolrpc.Order) (*Kit, error) {
+	details *poolrpc.Order, parseOpts ...OrderParseOption) (*Kit, error) {
+
+	opts := defaultParseOptions()
+	for _, newOpt := range parseOpts {
+		newOpt(opts)
+	}
 
 	var nonce Nonce
 	copy(nonce[:], details.OrderNonce)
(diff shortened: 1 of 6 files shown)


rpcserver.go
```diff
@@ -1185,18 +1186,66 @@ func prepareAndSubmitOrder(ctx context.Context, o order.Order,
 	)
 }
 
+// lndVersionIsCompatible makes sure the detected lnd version is compatible
+// with our current version requirements. True is returned if the version is
+// compatible, and false otherwise.
+func lndVersionIsCompatible(actual *verrpc.Version,
```
positiveblue, line 1192: I have some dobuts here:

- Do we need to check something else beyond the lnd version (`featurebits`?)

- Do we need to check the `BuildTags`?

- May be worth moving this function to `"github.com/lightningnetwork/lnd/lnrpc/verrpc"` itself so we do not have to copy paste it in multiple projects.

- Here we assume that all versions will be backwards compatible, we need to keep this in mind in the future then.

order/rpc_parse.go
```diff
@@ -19,10 +19,41 @@ import (
...
+}
+
 // ParseRPCOrder parses the incoming raw RPC order into the go native data
 // types used in the order struct.
 func ParseRPCOrder(version, leaseDuration uint32,
-	details *poolrpc.Order) (*Kit, error) {
+	details *poolrpc.Order, parseOpts ...OrderParseOption) (*Kit, error) {
+
+	opts := defaultParseOptions()
+	for _, newOpt := range parseOpts {
```
positiveblue, line 54: nice

positiveblue (approved):
LGTM but I have some doubts that may be blocking

rpcserver.go
```diff
@@ -1185,18 +1186,66 @@ func prepareAndSubmitOrder(ctx context.Context, o order.Order,
...
+	expected *verrpc.Version) bool {
+
+	// We need to check the versions parts sequentially as they are
+	// hierarchical.
+	if actual.AppMajor != expected.AppMajor {
```
positiveblue, line 1197: BTW linter is complaining about not using the pattern

```
if actual.AppMajor != expected.AppMajor {
    return actual.AppMajor > expected.AppMajor
}
```

server.go
```diff
@@ -128,6 +142,14 @@ func (s *Server) Start() error {
 		return nil
 	}
 
+	// Now that we have lnd, lets extract the current version so we can use
+	// this to gate features that we'll use based on the functionality
+	// available.
+	s.lndVersion, err = s.lndServices.Versioner.GetVersion(context.Background())
```
guggero, line 148: We already have this in `lndServices.Version`, no need to call it again.

rpcserver.go
guggero: It's possible the user has lnd `v0.14.x` but explicitly disables the script enforcement functionality. But then they would just reject any matched orders in the channel acceptor.

We don't need to check build tags though. They are checked on startup already. So we could leave them empty in the `scriptEnforceVersion` variable.

I created a PR for exporting `lndclient.AssertVersionCompatible`, you just need to bump the `lndclient` version to `v0.14.0-7` to get it.

rpcserver.go
```diff
@@ -1185,18 +1186,66 @@ func prepareAndSubmitOrder(ctx context.Context, o order.Order,
...
 
+	// We'll use this channel type selector to pick a channel type based on
+	// the current connected lnd version. This lets us graceful update to
+	// new features as they're available, while still supporting older
+	// versions of lnd.
+	chanTypeSelector := order.WithDefaultChannelType(func() order.ChannelType {
+		// If they didn't specify a value, then we'll select one based
+		// on the version of lnd we detect.
+		if lndVersionIsCompatible(s.server.lndVersion, scriptEnforceVersion) {
+			return order.ChannelTypeScriptEnforced
```
guggero, line 1236: Nice use of functional options :100:

guggero (changes requested):
Can be simplified by re-using code from `lndclient`. Other than that LGTM :tada:

rpcserver.go
Roasbeef: Thanks for that PR @guggero! Using that function directly now instead of copying things over.

rpcserver.go
Roasbeef: Should be gone now that the newly exported `lndclient` method is used instead.

server.go
Roasbeef: Nice, I must've missed that when initially looking over the `godoc`.

guggero (approved):
LGTM :tada:

positiveblue (approved):

guggero:
The linter complains about the naming, will fix in separate PR.

PR #331, opened 2021-12-21
Conditional batch version for itests

guggero:
Makes the batch version configurable to enable integration testing of different client batch versions in the same batch.

#### Pull Request Checklist
  used.

### DIFF
--- a/config.go
+++ b/config.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcutil"
+	"github.com/lightninglabs/pool/order"
 	"github.com/lightningnetwork/lnd/cert"
 	"github.com/lightningnetwork/lnd/lncfg"
 	"github.com/lightningnetwork/lnd/lnrpc"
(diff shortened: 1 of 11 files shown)


order/batch_verifier.go
```diff
@@ -75,13 +75,12 @@ type batchVerifier struct {
...
-	// than we use for creating the batch. Otherwise we bail out of the batch.
-	// This should already be handled when the client connects/authenticates.
-	// But doesn't hurt to check again.
+	// First, make sure the server used the same batch version than we use
+	// for creating the batch. Otherwise, we bail out of the batch.
+	// This should already be handled when the client connects/
+	// authenticates. But doesn't hurt to check again.
 	if batch.Version != v.version {
-		return ErrVersionMismatch
+		return NewErrVersionMismatch(v.version)
```
positiveblue, line 83: maybe it's worth to add both versions to the log message. Something like

```
client version (%d) mismatches server version (%d)
```

positiveblue (approved):
LGTM, this will make testing/debugging easier.

order/batch_verifier.go
guggero: Good idea, done.

config.go
```diff
@@ -160,6 +161,15 @@ type Config struct {
...
 	// dialing the auctioneer server.
 	AuctioneerDialOpts []grpc.DialOption
+
+	// DebugConfig is a set of debug options used for development and
+	// testing only.
+	DebugConfig *DebugConfig `group:"debug" namespace:"debug" hidden:"true"`
+}
+
+// DebugConfig is a set of debug options used for development and testing only.
+type DebugConfig struct {
```
bhandras, line 171: Very nice!

bhandras (approved):
LGTM :1st_place_medal:

PR #335, opened 2022-01-10
Client side account recovery

positiveblue:
Enable full client-side recovery of account state.

Using the same LND node as a backend, the process looks like following:
- The user provides a target (number of accounts to be recovered)
- The pool client is able to derive a number of potential accounts using the auctioneer public key (hardcoded) and the derived secret (the user is using the same LND node as before)
- We then need to find the initial state for the accounts that were created. For that, we explore the account space (expiry height and batchKey) and look for a valid account script **in each of the LND node transactions**.
- Once all the initial states have been found, we need to get the accounts to their latest state. For that, we need to find account modifications (updates) that trigger an on-chain tx. Again, this process involves searching for the right `expiry` height.

After the process is complete, the user is able to close the left expired accounts without any server assistance.


account/recovery.go
```diff
@@ -17,6 +20,69 @@ var (
...
+func getAuctioneerData(network string) (*auctioneerData, error) {
+	var auctioneerKey string
+	var batchKey string
+	var fstBlock uint32
+
+	switch network {
+	case "mainnet":
+		auctioneerKey = "034d1fd68b5cd92f694766aa8b3f90131613d1b731" +
+			"caf03be5d2d34c333163c606"
+		batchKey = "02824d0cbac65e01712124c50ff2cc74ce22851d7b444c1" +
```
guggero, line 50: The batch key is hard coded in the server and is always this value, in all environments.

account/recovery.go
```diff
@@ -17,6 +20,69 @@ var (
...
+
+// getAuctioneerData returns the auctioner data for a given environment.
+func getAuctioneerData(network string) (*auctioneerData, error) {
+	var auctioneerKey string
+	var batchKey string
+	var fstBlock uint32
+
+	switch network {
+	case "mainnet":
+		auctioneerKey = "034d1fd68b5cd92f694766aa8b3f90131613d1b731" +
```
guggero, line 48: Where did you get that value from? I think it's actually `028e87bdd134238f8347f845d9ecc827b843d0d1e27cdcb46da704d916613f4fce` (also see https://github.com/guggero/chantools/blob/master/cmd/chantools/closepoolaccount.go#L37 which was successfully used to recover an account).

account/recovery.go
```diff
@@ -17,6 +20,69 @@ var (
...
+			"caf03be5d2d34c333163c606"
+		batchKey = "02824d0cbac65e01712124c50ff2cc74ce22851d7b444c1" +
+			"bf2ae66afefb8eaf27f"
+		fstBlock = 648168
+	case "testnet":
+		// TODO: add default values for testnet.
```
guggero, line 54: Testnet values are:
auctioneer key: `025dea8f5c67fb3bdfffb3123d2b7045dc0a3c75e822fabb39eb357480e64c4a8a`
first block: 1834898 (txid b9bb97e655317e07b9015cd0b7fdd05e33f15864a0292bf2001cdcc17968cc78)

account/recovery.go
```diff
@@ -188,17 +192,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+		}
+		possibleAccounts[acc] = struct{}{}
+	}
+
+	return possibleAccounts, nil
+}
+
+// findAccounts tries to find on-chain footprint for the creation of a set
+// of possible pool accounts.
+func findAccounts(cfg RecoveryConfig,
```
guggero, line 236: I think we should add some test cases for this.

account/recovery.go
```diff
@@ -17,6 +20,69 @@ var (
...
+	switch network {
+	case "mainnet":
+		auctioneerKey = "034d1fd68b5cd92f694766aa8b3f90131613d1b731" +
+			"caf03be5d2d34c333163c606"
+		batchKey = "02824d0cbac65e01712124c50ff2cc74ce22851d7b444c1" +
+			"bf2ae66afefb8eaf27f"
+		fstBlock = 648168
+	case "testnet":
+		// TODO: add default values for testnet.
+	case "regtest":
```
guggero, line 55: I wonder if instead we should make it possible to overwrite these values through the CLI/gRPC instead? So there would be a field called `auctioneer_key` and `first_block` in the `RecoverAccounts` RPC that could be left empty to use default values but set for regtest.

guggero (commented):
Nice work on this! Exciting to see the last milestone for leaving the `alpha` stage being so close to completion :tada: 

Main comments are about unit test cases and ease of use in regtest.

poolrpc/trader.proto
```diff
@@ -870,6 +870,15 @@ enum MatchRejectReason {
 }
 
 message RecoverAccountsRequest {
+    /*
+    Fetch the latest account states from the Lightning Labs server.
+    */
+    bool server_assisted = 1;
```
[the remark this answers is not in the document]
positiveblue: Changed to `full_client`

account/recovery.go
positiveblue: updated, I thought I got it from your chantools but maybe I updated it for testing and forgot about it.

account/recovery.go
positiveblue: I added a couple of tests for this, I will add another one for the "findAccountUpdate"

rpcserver.go
```diff
@@ -1091,8 +1091,10 @@ func (s *rpcServer) RecoverAccounts(ctx context.Context,
 	var recoveredAccounts []*account.Account
 	var err error
 
+	target := req.AccountTarget
+
 	if !req.FullClient {
-		recoveredAccounts, err = s.serverAssistedRecovery(ctx)
+		recoveredAccounts, err = s.serverAssistedRecovery(ctx, target)
```
guggero, line 1166: According to the commit message this should now fall back to `DefaultAccountKeyWindow`. Or only if `target == 0`?

rpcserver.go
```diff
@@ -1074,6 +1074,75 @@ func (s *rpcServer) serverAssistedRecovery(ctx context.Context, target uint32) (
...
+	req *poolrpc.RecoverAccountsRequest) ([]*account.Account, error) {
+
+	txs, err := s.lndServices.Client.ListTransactions(ctx, 0, -1)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get txs from lnd: %v", err)
+	}
+
+	chainInfo, err := s.lndServices.Client.GetInfo(ctx)
+	if err != nil {
+		return nil, fmt.Errorf(
```
guggero, line 1088: nit: `Errorf` should be written in compact formatting.

account/recovery.go
```diff
@@ -17,6 +17,24 @@ var (
...
 )
 
+// GetAuctioneerData returns the auctioner data for a given environment.
+func GetAuctioneerData(network string) (string, uint32) {
+	var auctioneerKey string
+	var fstBlock uint32
+
+	switch network {
+	case "mainnet":
+		auctioneerKey = "028e87bdd134238f8347f845d9ecc827b843d0d1e2" +
```
guggero, line 32: Define these as nice constants at the top of the file? Together with the batch key in the next commit.

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
 	return accounts, nil
 }
 
+// recreatePossibleAccounts returns a set of potentially valid accounts
+// by generating a set of recovery keys.
+func recreatePossibleAccounts(ctx context.Context, cfg RecoveryConfig) (
+	map[*Account]struct{}, error) {
```
guggero, line 134: I don't think we should use pointers as map keys. Why not just use a slice here?

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+
+	possibleAccounts := make(map[*Account]struct{}, cfg.AccountTarget)
+
+	// Prepare the keys we are going to try. Possibly not all of them will
+	// be used.
+	KeyDescriptors, err := GenerateRecoveryKeys(
+		ctx, cfg.AccountTarget, cfg.Wallet,
+	)
+	if err != nil {
+		return possibleAccounts, fmt.Errorf(
```
guggero, line 144: nit: `Errorf()` and `Debugf()` calls should use the compact formatting to better distinguish against normal function/method calls.

account/recovery.go
guggero: Or use the account index as key if we need to be able to delete things in the map later.

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+	target := cfg.AccountTarget
+	accounts := make([]*Account, 0, target)
+
+	batchKey := cfg.InitialBatchKey
+	// DecrementKey so the first for iteration starts with the
+	// first batchKey value.
+	batchKey = poolscript.DecrementKey(batchKey)
+
+searchLoop:
+	for batchCounter := 0; batchCounter < 5000; batchCounter++ {
```
guggero, line 190: Should extract this as a constant and perhaps mention that currently there are about 1100 batches (Jan 2022) on mainnet. So we'll remember what this number relates to and when we need to update it.

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+							"height: %v\n"+
+							"batchKey: %v\n"+
+							"traderKey: %v\n",
+						h,
+						batchKey,
+						hex.EncodeToString(
+							key.SerializeCompressed(),
+						),
+					)
+					goto nextAccount
```
guggero, line 215: Can just be a `continue` then we don't need a second label.

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+				}
+
+				if tx, idx := appearsInTxs(
+					acc, script, cfg.Transactions,
+				); tx != nil {
```
guggero, line 220: Continue on `tx == nil` to avoid indentation?

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
+}
+
+// appearsInTxs tries to locate a script in any of the provided transactions.
+func appearsInTxs(acc *Account, script []byte, txs []lndclient.Transaction) (
+	*lndclient.Transaction, uint32) {
+
+	for _, tx := range txs {
+		idx, ok := poolscript.LocateOutputScript(tx.Tx, script)
+		if ok {
+			tx := tx // make scopelint happy
```
guggero, line 260: convention is to move this directly after the `range` line.

account/recovery.go
```diff
@@ -123,17 +128,254 @@ func recoverInitalState(ctx context.Context, cfg RecoveryConfig) (
...
-	// TODO (positiveblue): update account states
+nextAccount:
+	for _, acc := range accounts {
+		log.Debugf(
+			"Updating state for account with trader key %v",
+			hex.EncodeToString(
+				acc.TraderKey.PubKey.SerializeCompressed(),
+			),
+		)
+		for _, tx := range cfg.Transactions {
```
guggero, line 287: This will work as long as `lnd` wasn't also recovered from seed. If the node was recovered from the seed then the initial account funding transaction will be found on chain because it was funded with lnd wallet coins. But all subsequent account updates won't be recognized since lnd doesn't know the possible scripts.
That's why we'd need to follow the spends on chain. But I'm okay with us defining that as an edge case for now.

cmd/pool/account.go
```diff
@@ -68,6 +68,9 @@ const (
 	defaultExpiryRelative = 30 * 144
 
 	defaultFundingConfTarget = 6
+
+	// Account target for the account recovery process.
+	defaultAccountTarget = 10
```
guggero, line 73: nit: godoc comment format.

cmd/pool/account.go
```diff
@@ -782,6 +785,29 @@ var recoverAccountsCommand = cli.Command{
 	All open or pending orders of any recovered account will be canceled on
 	the auctioneer's side and won't be restored in the trader's database.
 	`,
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name: "full_client",
+			Usage: "fetch the latest account state from the " +
+				"server. If false, the full process will" +
+				"run locally, but it could take hours.",
```
guggero, line 793: Should add the same comment to the gRPC field comment for API users.

cmd/pool/account.go
```diff
@@ -782,6 +785,29 @@ var recoverAccountsCommand = cli.Command{
...
+		cli.Uint64Flag{
+			Name: "account_target",
+			Usage: "number of accounts that we are looking to " +
+				"recover.",
+			Value: defaultAccountTarget,
+		},
+		cli.StringFlag{
+			Name:  "auctioneer_key",
+			Usage: "Auctioneer's public key.",
+			Value: "test",
```
guggero, line 804: Test value still here.

account/recovery_test.go
```diff
@@ -4,9 +4,251 @@ import (
...
+	"github.com/lightninglabs/pool/poolscript"
+	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+// getInitialBatchKey returns the first batchkey value.
+func getInitialBatchKey() *btcec.PublicKey {
+	batchKey, _ := DecodeAndParseKey(
+		"02824d0cbac65e01712124c50ff2cc74ce22851d7b444c1bf2ae66afefb" +
```
guggero, line 21: Use the constants I mentioned in an earlier comment.

account/recovery_test.go
```diff
@@ -4,9 +4,251 @@ import (
...
+				acc,
+				tx,
+			)
+			if tc.expectedError != "" {
+				assert.EqualError(t, err, tc.expectedError)
+				return
+			}
+			require.NoError(t, err)
+
+			if tc.expectedExpiry != res.Expiry {
```
guggero, line 238: nit: `require.Equal()`

guggero (commented):
Getting pretty close! Mostly down to style and UX comments now.

Will run a set of manual tests in the next round.

account/recovery_test.go
```diff
@@ -4,9 +4,251 @@ import (
...
+								},
+							},
+						},
+					},
+				)
+			}
+
+			accounts := findAccounts(tc.config, possibleAccounts)
+
+			if len(accounts) != tc.expectedAccounts {
```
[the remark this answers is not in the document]
positiveblue: I did not know this one!

poolrpc/trader.proto
```diff
@@ -870,6 +870,29 @@ enum MatchRejectReason {
...
+    bool full_client = 1;
+
+    /*
+    Number of accounts that we are trying to recover. Used during the
+    full_client recovery process.
+    */
+    uint32 account_target = 2;
+
+    /*
+    Auctioneer's public key. Used during the full_client recovery process.
```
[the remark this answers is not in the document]
positiveblue: I will leave it without the comment for the height hint because it could be used in the future to narrow the searches if you know approx when you created your account.

rpcserver.go
positiveblue: Comment & code updated.

guggero:
I tested this on regtest today.
I created two accounts, the second one was involved in a batch, then I did a deposit and a withdrawal, both times extending the expiry by 1440 blocks.

The recovery is still running after almost 9 hours, and it looks like it's working on the last transaction:

```
2022-01-21 14:44:19.681 [DBG] ACCT: found accout with trader key 0308d7767416cbc8f0d8b287fbec4f4f7ab20795241fd609fc12ce30554a37d2ed
2022-01-21 14:46:26.426 [DBG] ACCT: found accout with trader key 020d51947a486b49137b839f8abcdab92d4c6cb958f4a5f55fa8366c9c5ceb9042
2022-01-21 14:50:39.885 [DBG] ACCT: found accout with trader key 020d51947a486b49137b839f8abcdab92d4c6cb958f4a5f55fa8366c9c5ceb9042
2022-01-21 14:52:46.219 [DBG] ACCT: found accout with trader key 020d51947a486b49137b839f8abcdab92d4c6cb958f4a5f55fa8366c9c5ceb9042
```

Any idea what could be the problem? Doesn't look like this is going to finish.

account/recovery.go
```diff
@@ -220,16 +379,121 @@ func recoverInitialState(ctx context.Context, cfg RecoveryConfig) ([]*Account,
...
+		}
+
+		for _, tx := range block.Transactions {
+			for idx, acc := range accounts {
+				// It's safe to ignore accounts that are in a
+				// terminal state.
+				if acc.State == StateClosed {
+					continue
+				}
+				found := poolscript.IncludesPreviousOutPoint(
```
guggero, line 402: Very nice, this now enables users to recover even if they needed to restore `lnd` from seed. And checking for the outpoint here should speed things up as much as possible.

account/recovery.go
```diff
@@ -220,16 +379,121 @@ func recoverInitialState(ctx context.Context, cfg RecoveryConfig) ([]*Account,
...
+
+				if !found {
+					continue
+				}
+
+				newAcc, err := findAccountUpdate(
+					cfg, acc, blockHeight, tx,
+				)
+				if err != nil {
+					log.Debugf("unable to find account "+
```
guggero, line 414: I think if we can't find an account update, we need to consider the account to be closed. Perhaps we should update this log to info and say something like `Account was spent in tx %x but not re-created. Assuming account was fully spent or closed.` and then set the state to closed (also see comment below about the dust value).

account/recovery.go
```diff
@@ -220,16 +379,121 @@ func recoverInitialState(ctx context.Context, cfg RecoveryConfig) ([]*Account,
...
+	tx *wire.MsgTx) (*Account, error) {
+
+	newAcc := acc.Copy()
+	newAcc.BatchKey = poolscript.IncrementKey(newAcc.BatchKey)
+
+	idx, ok := matchScript(newAcc, newAcc.Expiry, tx)
+	if ok {
+		newAcc.Value = btcutil.Amount(tx.TxOut[idx].Value)
+		// TODO (positiveblue): change 0 for dust limit.
+		if newAcc.Value == 0 {
```
guggero, line 437: We never re-create an account below the dust value. So if an account is fully spent (e.g. its balance falls below the dust threshold), its account isn't re-created. In that case we won't be able to find an output for it on chain. See comment above about just logging and assuming it to be closed.

account/recovery.go
```diff
@@ -220,16 +379,121 @@ func recoverInitialState(ctx context.Context, cfg RecoveryConfig) ([]*Account,
...
+
+		return newAcc, nil
+	}
+
+	return nil, fmt.Errorf("account update not found")
+}
+
+// matchScript creates a new Account script and tries to locate the output in a
+// transaction.
+func matchScript(acc *Account, expiry uint32, tx *wire.MsgTx) (uint32, bool) {
```
guggero, line 480: nit: since this is now only used by `findAccountUpdate`, we could use the helper here too to speed things up a bit.

guggero (commented):
Very nice work on this :tada: 
We're very very close now, only the TODOs about the account being closed need to be addressed, the rest looks good.
I was able to recover and close my regtest account with this :100:

guggero (approved):
tACK, great work on this :100: I'm very happy about how this turned out!

Pushed a small commit to fix a log message (because `chainhash.Hash` has a `String()` method, using `%x` causes double hex encoding).

PR #336, opened 2022-01-13
multi: use lndclient MacaroonService

ellemouton:
Since the code for creating and using a macaroon service is the same for
multiple projects (pool, loop, litd etc), the code has been unified in
lndclient. So this commit removes the macaroon service code and instead
uses the lndclient code.

Depends on https://github.com/lightninglabs/lndclient/pull/86

### DIFF
--- a/server.go
+++ b/server.go
@@ -340,8 +340,7 @@ func (s *Server) Start() error {
 // StartAsSubserver is an alternative start method where the RPC server does not
 // create its own gRPC server but registers on an existing one.
 func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
-	lndGrpc *lndclient.GrpcLndServices,
-	createDefaultMacaroonFile bool) error {
+	lndGrpc *lndclient.GrpcLndServices, withMacaroonService bool) error {
 
 	if atomic.AddInt32(&s.started, 1) != 1 {
 		return fmt.Errorf("trader can only be started once")
@@ -366,7 +365,7 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
 		}
 	}()
 
-	if createDefaultMacaroonFile {
+	if withMacaroonService {
 		// Create and start the macaroon service and let it create its default
 		// macaroon in case it doesn't exist yet.
 		var err error
--- a/server.go
+++ b/server.go
@@ -25,7 +25,6 @@ import (
 	"github.com/lightninglabs/pool/order"
 	"github.com/lightninglabs/pool/poolrpc"
 	"github.com/lightninglabs/pool/terms"
-	"github.com/lightningnetwork/lnd/kvdb"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/verrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -355,15 +367,33 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
 		}
 	}()
 
-	// Start the macaroon service and let it create its default macaroon in
-	// case it doesn't exist yet.
-	if err := s.startMacaroonService(createDefaultMacaroonFile); err != nil {
+	// Create and start the macaroon service and let it create its default
+	// macaroon in case it doesn't exist yet.
+	var err error
+	if s.macaroonService, err = lndclient.NewMacaroonService(
```
guggero, line 373: nit: that `if` is easily overlooked IMO. Separate into assignment and then condition?

server.go
```diff
@@ -161,12 +159,26 @@ func (s *Server) Start() error {
...
+		&lndclient.MacaroonServiceConfig{
+			DBPath:           s.cfg.BaseDir,
+			DBTimeout:        clientdb.DefaultPoolDBTimeout,
+			MacaroonLocation: poolMacaroonLocation,
+			MacaroonPath:     s.cfg.MacaroonPath,
+			Checkers: []macaroons.Checker{
+				macaroons.IPLockChecker,
+			},
+			RequiredPerms: RequiredPermissions,
+			DBPassword:    macDbDefaultPw,
```
guggero, line 174: See my comment in `lndclient` about migrating to a non-empty password. We might need to supply the lnd client, ephemeral key (should probably export that in `lndclient` for re-use instead of only having it in LiT)

guggero (commented):
Yayy, deleted code :tada: 

LGTM apart from the database password migration issue.

server.go
```diff
@@ -161,12 +161,35 @@ func (s *Server) Start() error {
...
+			LndClient:     &s.lndServices.LndServices,
+			EphemeralKey:  lndclient.MacaroonPubKey,
+			KeyLocator: &keychain.KeyLocator{
+				Family: 21,
+				Index:  0,
+			},
```
ellemouton, line 182: Not sure what to use here? can we just create any constant? does it necessarily need to differ from what Lit uses?

server.go
```diff
@@ -355,12 +378,38 @@ func (s *Server) StartAsSubserver(lndClient lnrpc.LightningClient,
 		}
 	}()
 
-	// Start the macaroon service and let it create its default macaroon in
-	// case it doesn't exist yet.
-	if err := s.startMacaroonService(createDefaultMacaroonFile); err != nil {
-		return err
+	if createDefaultMacaroonFile {
```
ellemouton, line 381: macaroon db not used if in subserver + stateless mode

server.go
guggero: I'd say let's also create a key locator variable in `lndclient` that can be used in all projects. I think it makes sense to use the same shared key password for all daemons, hence the same key locator everywhere.

server.go
guggero: Maybe it would be better to rename this to `withOwnMacaroonValidation` or `withMacaroonService` since this variable does a bit more now?

guggero (approved):
Very nice, nits only!

positiveblue (approved):
Thanks for this @ellemouton! :tada:

PR #337, opened 2022-01-14
rpcserver: fix nil version in subserver mode

guggero:
A new variable for the connected lnd node's version was added but was
not given a value when run in subserver mode within LiT.
This commit uses the lnd service's cached version directly and removes
the uninitialized variable.


Roasbeef (approved):
LGTM ⛳️

PR #340, opened 2022-02-14
sidecar: Init mailbox after server restarts

positiveblue:
Whenever we are are unable to receive a sidecar packet in an auto
sidecar negotiator try to reinitialize its mailbox. This will allow us to
continue reading/writing in the mailbox in cases where the error was
caused by the server going down (and already came back).

### DIFF
--- a/auto_sidecar.go
+++ b/auto_sidecar.go
@@ -314,6 +314,23 @@ func (a *SidecarNegotiator) autoSidecarReceiver(ctx context.Context,
 
 				select {
 				case <-retryTimer.backOff(backoffLabel):
+					// It is possible that we were not able to receive the
+					// packet because the server went down. In that case,
+					// we will try to init the sidecar mailbox again.
+					// There is no need to log/check this error. There are
+					// three possibilities:
+					//
+					// 1) If the error was not related with the sever being
+					// down, we will get an `AlreadyExists` error.
+					//
+					// 2) If the server is back, we will be able to reconnect
+					// successfully and receive the sidecar pkt in the next
+					// iteration.
+					//
+					// 3) If the server is down, we won't be able to reconnect.
+					_ = a.cfg.MailBox.InitSidecarMailbox(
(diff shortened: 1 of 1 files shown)


auto_sidecar.go
```diff
@@ -314,6 +314,23 @@ func (a *SidecarNegotiator) autoSidecarReceiver(ctx context.Context,
 
 				select {
 				case <-retryTimer.backOff(backoffLabel):
+					// It is possible that we were not able to receive the
```
guggero, line 317: nit: line length.

auto_sidecar.go
```diff
@@ -314,6 +314,23 @@ func (a *SidecarNegotiator) autoSidecarReceiver(ctx context.Context,
...
+					// packet because the server went down. In that case,
+					// we will try to init the sidecar mailbox again.
+					// There is no need to log/check this error. There are
+					// three possibilties:
```
guggero, line 321: nit: typo in possibilties (same below).

auto_sidecar.go
```diff
@@ -588,6 +605,21 @@ func (a *SidecarNegotiator) autoSidecarProvider(ctx context.Context,
...
+					//
+					// 1) If the error was not related with the sever being
+					// down, we will get an `AlreadyExists` error.
+					//
+					// 2) If the server is back, we will be able to reconnect
+					// successfully and receive the sidecar pkt in the next
+					// iteration.
+					//
+					// 3) If the server is down, we won't be able to reconnect.
+					_ = a.cfg.MailBox.InitAcctMailbox(streamID, acct.TraderKey)
```
guggero, line 622: :+1: 
I think this is a good fix.

guggero (approved):
Fix looks good to me. We could even try to init the mailbox every time we try to read from it. But then that would perhaps cause too many "already exists" errors to be logged server side. So this is a nice middle ground.

PR #341, opened 2022-02-18
Fix sidecar message handlers

positiveblue:
The first commit cleans up the shims whenever we reject a batch.  
The second commit populates the `OrderMatchReject` msg properly.

### DIFF
--- a/sidecar_acceptor.go
+++ b/sidecar_acceptor.go
@@ -525,41 +525,23 @@ func (a *SidecarAcceptor) handleServerMessage(
 	switch msg := serverMsg.Msg.(type) {
 
 	case *auctioneerrpc.ServerAuctionMessage_Prepare:
-		batchID := msg.Prepare.BatchId
 
 		sdcrLog.Tracef("Received prepare msg from server, "+
-			"batch_id=%x: %v", batchID, spew.Sdump(msg))
+			"batch_id=%x: %v", msg.Prepare.BatchId, spew.Sdump(msg))
 
-		newBatch, err := a.matchPrepare(a.pendingBatch, msg.Prepare)
-		if err != nil {
-			sdcrLog.Errorf("Error handling prepare message: %v",
-				err)
-			return a.sendRejectBatch(batchID, err)
+		if err := a.matchPrepare(msg.Prepare); err != nil {
+			sdcrLog.Errorf("unable to handle prepare message: %v", err)
+			return a.sendRejectBatch(msg.Prepare.BatchId, nil, err)
 		}
-
-		// We know we're involved in a batch, so let's store it for the
-		// next step.
-		a.pendingBatch = newBatch
 
 	case *auctioneerrpc.ServerAuctionMessage_Sign:
-		batchID := msg.Sign.BatchId
 
 		sdcrLog.Tracef("Received sign msg from server, batch_id=%x: %v",
-			batchID, spew.Sdump(msg))
+			msg.Sign.BatchId, spew.Sdump(msg))
 
-		// Assert we're in the correct state to receive a sign message.
-		if a.pendingBatch == nil ||
-			!bytes.Equal(batchID, a.pendingBatch.ID[:]) {
-
-			err := fmt.Errorf("error processing batch sign "+
-				"message, unknown batch with ID %x", batchID)
-			sdcrLog.Errorf("Error handling sign message: %v", err)
-			return a.sendRejectBatch(batchID, err)
-		}
-
-		if err := a.matchSign(a.pendingBatch); err != nil {
-			sdcrLog.Errorf("Error handling sign message: %v", err)
-			return a.sendRejectBatch(batchID, err)
+		if err := a.matchSign(msg.Sign); err != nil {
+			sdcrLog.Errorf("unable to handle sign message: %v", err)
+			return a.sendRejectBatch(a.pendingBatch.ID[:], a.pendingBatch, err)
 		}
 
 	case *auctioneerrpc.ServerAuctionMessage_Finalize:
@@ -568,13 +550,8 @@ func (a *SidecarAcceptor) handleServerMessage(
 		sdcrLog.Tracef("Received finalize msg from server, "+
 			"batch_id=%x: %v", batchID, spew.Sdump(msg))
 
-		// All we need to do now is some cleanup. Even if the cleanup
-		// fails, we want to clear the pending batch as we won't receive
-		// any more messages for it.
-		batch := a.pendingBatch
-		a.pendingBatch = nil
-
-		a.matchFinalize(batch)
+		// This operation cannot fail.
+		a.matchFinalize()
 
 	default:
 		sdcrLog.Debugf("Received msg %v from auctioneer on sidecar "+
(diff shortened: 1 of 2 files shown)


sidecar_acceptor.go
```diff
@@ -617,8 +595,7 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
 	// that's being used to pay for the sidecar channel.
 	err = a.cfg.FundingManager.PrepChannelFunding(batch, a.getSidecarAsOrder)
 	if err != nil {
-		return nil, fmt.Errorf("error preparing channel funding: %v",
-			err)
+		return fmt.Errorf("error preparing channel funding: %v", err)
```
guggero, line 598: We need to wrap this error, otherwise the `errors.As(failure, &partialReject):` below won't trigger.

sidecar_acceptor.go
```diff
@@ -633,20 +610,48 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
...
+		return fmt.Errorf("pending batchID was: %x got: %x",
+			a.pendingBatch.ID[:], msg.BatchId)
+	}
+
+	batch := a.pendingBatch
+	batchID := a.pendingBatch.ID[:]
+
 	channelInfos, err := a.cfg.FundingManager.SidecarBatchChannelSetup(
-		a.pendingBatch, a.pendingOpenChanClient, a.getSidecarAsOrder,
+		batch, a.pendingOpenChanClient, a.getSidecarAsOrder,
```
guggero, line 654: Same here re wrapping the error.

sidecar_acceptor.go
```diff
@@ -763,14 +774,42 @@ func (a *SidecarAcceptor) getSidecarAsOrder(o order.Nonce) (order.Order, error)
...
+		versionMismatch *order.ErrVersionMismatch
+	)
+	switch {
+	case errors.As(failure, &versionMismatch):
+		msg.Reject.ReasonCode = auctioneerrpc.OrderMatchReject_BATCH_VERSION_MISMATCH
+
+	case errors.Is(failure, order.ErrMismatchErr):
+		msg.Reject.ReasonCode = auctioneerrpc.OrderMatchReject_SERVER_MISBEHAVIOR
+
+	case errors.As(failure, &partialReject):
```
guggero, line 808: Aaah, nice. Now I see why we were matched over and over again...

sidecar_acceptor.go
```diff
@@ -777,6 +777,15 @@ func (a *SidecarAcceptor) getSidecarAsOrder(o order.Nonce) (order.Order, error)
 func (a *SidecarAcceptor) sendRejectBatch(batchID []byte, batch *order.Batch,
 	failure error) error {
 
+	if batch != nil {
+		// As we're rejecting this batch, we'll now cancel all funding shims
+		// that we may have registered since we may be matched with a distinct
+		// set of channels if this batch is repeated.
+		if err := a.removeShims(batch); err != nil {
```
guggero, line 784: :+1:

guggero (commented):
Great fixes and nice find! A couple of comments about error wrapping. But I'm pretty confident you found the underlying cause for the problem Voltage described.

I think we should add a server side integration test for this specific case, pulling in the version of this PR to verify.

sidecar_acceptor.go
```diff
@@ -525,41 +525,23 @@ func (a *SidecarAcceptor) handleServerMessage(
 	switch msg := serverMsg.Msg.(type) {
 
 	case *auctioneerrpc.ServerAuctionMessage_Prepare:
-		batchID := msg.Prepare.BatchId
```
Roasbeef, line 528: Typo in commit message: rejectig -> rejecting

sidecar_acceptor.go
```diff
@@ -525,41 +525,23 @@ func (a *SidecarAcceptor) handleServerMessage(
...
-				err)
-			return a.sendRejectBatch(batchID, err)
+		if err := a.matchPrepare(msg.Prepare); err != nil {
+			sdcrLog.Errorf("unable to handle prepare message: %v", err)
+			return a.sendRejectBatch(msg.Prepare.BatchId, nil, err)
 		}
 
-		// We know we're involved in a batch, so let's store it for the
-		// next step.
-		a.pendingBatch = newBatch
```
Roasbeef, line 542: Why do we no longer store the batch here? We'll need to remove it if the execution fails for w/e reason.

sidecar_acceptor.go
```diff
@@ -589,25 +566,25 @@ func (a *SidecarAcceptor) handleServerMessage(
...
 		batch.ID[:], len(batch.MatchedOrders))
 
 	// If there is still a pending batch around from a previous iteration,
 	// we need to clean up the pending channels first.
-	if pendingBatch != nil {
-		if err := a.removeShims(pendingBatch); err != nil {
-			return nil, fmt.Errorf("unable to cleanup previous "+
-				"batch: %v", err)
+	if a.pendingBatch != nil {
+		if err := a.removeShims(a.pendingBatch); err != nil {
```
Roasbeef, line 584: So this is part of the core fix here? That we wipe the batch after we remove the shim?

sidecar_acceptor.go
```diff
@@ -633,23 +609,51 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
...
 	})
 	if err != nil {
-		return nil, fmt.Errorf("error sending accept msg: %v", err)
+		return fmt.Errorf("error sending accept msg: %v", err)
 	}
 
-	return batch, nil
+	// We know we're involved in a batch, so let's store it for the
+	// next step.
+	a.pendingBatch = batch
```
Roasbeef, line 617: Ok I see we only store it now once all the other operations succeed...

sidecar_acceptor.go
```diff
@@ -633,23 +609,51 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
...
 // a bid order) the tasks are simplified compared to normal bid order execution.
 //
 // NOTE: The lock must be held when calling this method.
-func (a *SidecarAcceptor) matchSign(batch *order.Batch) error {
+func (a *SidecarAcceptor) matchSign(
+	msg *auctioneerrpc.OrderMatchSignBegin) error {
+
+	// Assert we're in the correct state to receive a sign message.
+	if !a.isPending(msg.BatchId) {
+		return fmt.Errorf("pending batchID was: %x got: %x",
```
Roasbeef, line 645: Seems this would indicate a logic error elsewhere if we see it frequently in practice, or even a bug in the server's execution logic.

sidecar_acceptor.go
Roasbeef: Ah yeah makes sense, otherwise we'll only cancel the shims the _next_ time a batch is proposed.

sidecar_acceptor.go
positiveblue: Yes, we do, but all the logic is inside the `a.matchPrepare`, `a.matchSign` and `a.matchFinalize`

sidecar_acceptor.go
positiveblue: Yes, the fix was merged in another PR (already in master). Usually we want to remove the shims of the previous batch, but it was possible to end up in a state were there was not previous batch stored and some shims were created

sidecar_acceptor.go
positiveblue: Agreed, it should not be the normal flow. I have never seen it in our logs, but the code was already there and I think it does not hurt us to keep it. (I hope we end up in this state though*)

kornpow:
Running this branch on my machine.
Registered a sidecar ticket, and I got the connection error that I had seemed to get stuck on before for my first batch, but then when the second batch rolled around I got matched to somebody else and it worked!

I am hopeful that this fixes it!

Issue #327, opened 2021-12-14
sidecar channels: don't try to clean up previous batch state if in provider mode


guggero:
If a node is providing sidecar channels, it won't be involved in the actual channel opening process.
So in case of a batch rollback (new `PrepareMsg` without a finalize of the previous batch), there is nothing to clean up.
In fact, the cleanup attempt creates log entries with errors:

```
2021-12-14 18:15:28.432 [INF] RPCS: Received PrepareMsg for batch=020xxxx, num_orders=1
2021-12-14 18:15:28.439 [WRN] FNDG: Unable to unregister funding shim (pendingChanID=2dxxxx) for order=8ff7xxx
2021-12-14 18:15:28.444 [DBG] FNDG: Cleaning up incomplete/replaced pending channel in lnd was unsuccessful for order=8ff7xxxx (channel_point=d3c5faxxxx:0), assuming timeout when funding: rpc error: code = Unknown desc = unable to find closed channel summary
```

PR #342, opened 2022-02-24
sidecar: clean old shims for orders in current batch

positiveblue:
Quick (dirty) fix to ensure that we do not end up in an endless loop of batch rejections. One side effect is that we pollute the logs a bit more (issue #327).

#341 will fix this properly, by ensuring that shim creation after matching is atomic. Meanwhile this unblocks some of our traders.

### DIFF
--- a/sidecar_acceptor.go
+++ b/sidecar_acceptor.go
@@ -601,6 +601,13 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
 	sdcrLog.Infof("Received PrepareMsg for batch=%x, num_orders=%v",
 		batch.ID[:], len(batch.MatchedOrders))
 
+	// Ensure that we do not have any registered shims for the orders
+	// in this batch. This is not supposed to happen but we have a bug.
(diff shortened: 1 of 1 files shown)


sidecar_acceptor.go
```diff
@@ -601,6 +601,13 @@ func (a *SidecarAcceptor) matchPrepare(pendingBatch *order.Batch,
 	sdcrLog.Infof("Received PrepareMsg for batch=%x, num_orders=%v",
 		batch.ID[:], len(batch.MatchedOrders))
 
+	// Ensure that we do not have any registered shims for the orders
+	// in this batch. This is not supposed to happen but we have a bug.
+	if err = a.removeShims(batch); err != nil {
```
guggero, line 606: Perhaps link to the PR where this will be fixed for good?
Also add a comment that `removeShims` will only return an error for serious problems (for example not being able to read orders from disk) and not if there is no pending shim (only a warning will be logged but no error returned).

guggero (approved):
LGTM, nice quick fix!

PR #344, opened 2022-03-02
cmd: improve InitAccount cli message

positiveblue:
Our message was really confusing. If `sat_per_vbyte` was provided, Init account was printing a `conf_target` that was not used at all.


Roasbeef (approved):
LGTM 🦗

Thx for the quick fix here!

PR #347, opened 2022-03-15
rpcserver: report lease expiries correctly

positiveblue:
Thaw height is now absoulte (used to be relative). We could check the
channel type (for type SCRIPT_ENFORCED_LEASE is always absolute) but
we can simply follow the ["500,000" value rule](https://github.com/lightningnetwork/lnd/blob/262591c3331c2170697fe89b8c49ab857db0adf7/channeldb/channel.go#L726).


guggero (approved):
LGTM, nice fix :tada:

PR #350, opened 2022-03-28
Filter order matches by node id

positiveblue:
This PR includes the client changes for including node id constrains to pool orders. The users will be able to add a list of allowed/not allowed node ids for their order.

Thinks to take into account:
- Every order has its own constrains. I may not want to match with someone in order X but not filter him out in order Y.
- The filters are included during the order creation. This is important for sidecars because the provider may need to get the information from the recipient before creating the sidecar.


poolrpc/trader.proto
```diff
@@ -620,6 +620,12 @@ message Order {
 
     // The channel type to use for the resulting matched channels.
     OrderChannelType channel_type = 13;
+
+    // List of nodes that will be allowed to match with our order.
+    repeated bytes allowed_node_ids = 14;
+
+    // List of nodes that won't be allowed to match with our order.
```
guggero, line 627: nit: add a comment to both that the lists are mutually exclusive.

cmd/pool/order.go
```diff
@@ -116,6 +116,18 @@ var sharedFlags = []cli.Flag{
 			channelTypePeerDependent,
 			channelTypeScriptEnforced),
 	},
+	cli.StringSliceFlag{
+		Name: "allowed_node_ids",
```
guggero, line 120: I assume this is specified multiple times to add multiple IDs? Perhaps the flag name should be singular then and the description should say "can be specified multiple times".

cmd/pool/order.go
```diff
@@ -281,6 +319,26 @@ func parseAccountKey(ctx *cli.Context, args cli.Args) ([]byte, error) {
 	return hex.DecodeString(acctKeyStr)
 }
 
+// parseNodePubKeySlice parses the list of node ids in the paramater matching
+// the given `key`.
+//
+// NOTE: the parameter must contain a string slice. The strings are decoded
+// but not parsed as a btcec.PublicKey.
```
guggero, line 326: nit: "are hex decoded but not parsed..."

order/rpc_parse.go
```diff
@@ -115,6 +115,22 @@ func ParseRPCOrder(version, leaseDuration uint32,
 			details.ChannelType)
 	}
 
+	if len(details.AllowedNodeIds) > 0 &&
+		len(details.NotAllowedNodeIds) > 0 {
+
+		return nil, errors.New("my story")
```
guggero, line 121: :joy:

order/rpc_parse.go
```diff
@@ -115,6 +115,22 @@ func ParseRPCOrder(version, leaseDuration uint32,
...
+	}
+
+	kit.AllowedNodeIDs = make([][33]byte, len(details.AllowedNodeIds))
+	for idx, nodeID := range details.AllowedNodeIds {
```
guggero, line 125: We should parse the public keys here to make sure no invalid data is given to the RPC call.

rpcserver.go
```diff
@@ -1245,6 +1245,16 @@ func (s *rpcServer) validateOrder(order order.Order, acct *account.Account,
 			leaseDuration, auctionTerms.LeaseDurationBuckets)
 	}
 
+	// If the order does not specify any AllowedNodeIDs/NotAllowedNodeIDs
+	// it means that it can match with any other order. However, both
+	// fields cannot be set at the same time.
+	if len(order.Details().AllowedNodeIDs) > 0 &&
+		len(order.Details().NotAllowedNodeIDs) > 0 {
+
+		return fmt.Errorf("invalid matching filtering: allowed and " +
```
guggero, line 1254: nit: "allowed and not allowed node ids set at the same time"

clientdb/order.go
```diff
@@ -689,6 +698,29 @@ func deserializeOrderTlvData(r io.Reader, o order.Order) error {
 		o.Details().ChannelType = order.ChannelType(channelType)
 	}
 
+	if allowedNodeIDsBytes, ok := parsedTypes[allowedNodeIDsType]; ok {
```
guggero, line 701: This isn't really correct. You only get the `allowedNodeIDsBytes != nil` when the TLV library considers this to be an "unknown record type". So instead you should add a primitive record to the top that just expects a byte slice. Then here you can just check `t, ok := parsedTypes[allowedNodeIDSType]; ok && t == nil` and expect that byte slice to be filled with the data you want.

clientdb/order.go
```diff
@@ -689,6 +698,29 @@ func deserializeOrderTlvData(r io.Reader, o order.Order) error {
...
+		numberOfItems := len(allowedNodeIDsBytes) / 33
+		o.Details().AllowedNodeIDs = make([][33]byte, numberOfItems)
+		if err := binary.Read(
```
guggero, line 704: This falls back to some reflection magic. Since we know exactly that we're serializing and de-serializing a slice of arrays, should we add two helper functions that then do the explicit 33 byte calculation? Then we can just use `copy()` and don't need `binary.Read()` or a buffer for the reverse.

clientdb/order_test.go
```diff
@@ -43,6 +43,12 @@ func TestSubmitOrder(t *testing.T) {
 	}
 	o.Details().MinUnitsMatch = 10
 	o.Details().ChannelType = order.ChannelTypeScriptEnforced
+
+	// It is not possible for a oder to have AllowedNodeIDs and
+	// NotAllowedNodeIDs at the same time but we want to test
+	// serialization/deserialization here.
+	o.Details().AllowedNodeIDs = [][33]byte{{1, 2, 3}}
```
guggero, line 50: we should add at least two IDs to make sure multiple entries can be stored and retrieved correctly.

auctioneerrpc/auctioneer.proto
```diff
@@ -886,6 +886,12 @@ message ServerOrder {
     expressed in satoshis per 1000 weight units (sat/kW).
     */
     uint64 max_batch_fee_rate_sat_per_kw = 13;
+
+    // List of nodes that will be allowed to match with our order.
+    repeated bytes allowed_node_ids = 14;
```
guggero, line 891: Same here re adding a comment that the fields are mutually exclusive.

order/manager.go
```diff
@@ -343,6 +343,26 @@ func (m *manager) OrderMatchValidate(batch *Batch, bestHeight uint32) error {
...
 
+	// Verify that the id of the matched node was not filtered out.
+	for nonce, matches := range batch.MatchedOrders {
+		o, err := m.cfg.Store.GetOrder(nonce)
+		if err != nil {
+			return fmt.Errorf("error validating matched orders: %w",
+				err)
+		}
+
+		for _, match := range matches {
```
guggero, line 354: Nice validation!

guggero (commented):
Very nice and clean PR!
Only a few minor comments about the TLV encoding/decoding, everything else looks pretty good :tada:

clientdb/order.go
positiveblue: Do you mean flattering the `[][33]byte` to a `[]byte` and then use the default tlv types serialize/deserialize? then add a function to transform from `[]byte` to `[][33]byte`?

guggero: Yes, exactly. Feels more robust and performant than relying on some reflection code of `binary.Read()`.

cmd/pool/order.go
```diff
@@ -116,6 +116,20 @@ var sharedFlags = []cli.Flag{
 			channelTypePeerDependent,
 			channelTypeScriptEnforced),
 	},
+	cli.StringSliceFlag{
+		Name: "allowed_node_id",
+		Usage: "the list of nodes this order is able to match with; " +
+			"if empty, the order will be able to match with any " +
+			"node unless not_allowed_node_id is set. Cann be " +
```
guggero, line 123: nit: typo in "Cann"

cmd/pool/order.go
```diff
@@ -116,6 +116,20 @@ var sharedFlags = []cli.Flag{
...
+			"specified multiple times",
+	},
+	cli.StringSliceFlag{
+		Name: "not_allowed_node_id",
+		Usage: "the list of nodes this order is not able to match " +
```
guggero, line 128: nit: s/not able to match with/not allowed to match with? "not able" implies a problem, but this is just a user preference.

guggero, line 128: nit: s/not able to match with/not allowed to match with? "not able" implies a problem, but this is just a user preference.

order/interfaces.go
```diff
@@ -340,6 +340,14 @@ type Kit struct {
 	// ChannelType denotes the channel type that must be used for the
 	// resulting matched channels.
 	ChannelType ChannelType
+
+	// AllowedNodeIDs is the list of node ids this order is allowed to
+	// macth with.
```
guggero, line 345: nit: typo in "match", here and comment below.

guggero, line 345: nit: type in "match".

order/interfaces.go
```diff
@@ -340,6 +340,14 @@ type Kit struct {
...
+	AllowedNodeIDs [][33]byte
+
+	// NotAllowedNodeIDs is the list of node ids this order is not allowed
+	// to macth with.
```
guggero, line 349: nit: same typo here.

clientdb/order.go
```diff
@@ -728,10 +752,59 @@ func serializeOrderTlvData(w io.Writer, o order.Order) error {
...
 	if err != nil {
 		return err
 	}
 
 	return tlvStream.Encode(w)
 }
+
+// FlatterPubKeySlice returns a flatten array of bytes from the
+// given array of public keys.
+func FlatterPubKeySlice(pubKeys [][33]byte) []byte {
```
guggero, line 783: nit: should be called `FlattenPubKeySlice`?

guggero, line 783: nit: should be called `FlattenPubKeySlice`?

clientdb/order.go
```diff
@@ -728,10 +752,59 @@ func serializeOrderTlvData(w io.Writer, o order.Order) error {
...
+	flatten := make([]byte, len(pubKeys)*33)
+
+	for i, key := range pubKeys {
+		for j, v := range key {
```
guggero, line 787: nit: could also use `copy()` here to avoid a nested loop.

guggero, line 787: nit: could also use `copy()` here to avoid a nested loop.

clientdb/order.go
```diff
@@ -728,10 +752,59 @@ func serializeOrderTlvData(w io.Writer, o order.Order) error {
...
+// AssemblePubKeySlice is the inverse of FlatterPubKeySlice. Given a slice of
+// bytes representing a flattened slice of pubkeys returns the original slice
+// of pubkeys.
+func AssemblePubKeySlice(flatten []byte) [][33]byte {
+	numOfPubKeys := len(flatten) / 33
+	pubKeys := make([][33]byte, 0, numOfPubKeys)
+
+	for i := 0; i < numOfPubKeys; i++ {
+		var pubKey [33]byte
+		for j := 0; j < 33; j++ {
```
guggero, line 804: Same here, can just use copy with the correct slice indices.

guggero, line 804: Same here, can just use copy with the correct slice indices.

order/manager.go
```diff
@@ -522,3 +542,34 @@ func parseNodeUris(uris []string) ([]net.Addr, error) {
...
+	if len(notAllowed) > 0 {
+		for _, notAllowedNodeID := range notAllowed {
+			if nodeID == notAllowedNodeID {
+				return false
+			}
+		}
+		return true
+	}
+
+	// If we get here is because both lists were empty
```
guggero, line 573: nit: missing full stop.

guggero, line 573: nit: missing full stop.

guggero (approved):
Very nice, only nits remaining. Going to do a full test run while reviewing the subasta PR.

clientdb/order.go
```diff
@@ -728,10 +752,55 @@ func serializeOrderTlvData(w io.Writer, o order.Order) error {
...
+	}
+
+	return flatten
+}
+
+// AssemblePubKeySlice is the inverse of FlattenPubKeySlice. Given a slice of
+// bytes representing a flattened slice of pubkeys returns the original slice
+// of pubkeys.
+func AssemblePubKeySlice(flatten []byte) [][33]byte {
+	numOfPubKeys := len(flatten) / 33
```
bhandras, line 797: nit: we could add a check make sure that `len(flatten) % 33 == 0`

auctioneer/client.go
```diff
@@ -431,6 +431,20 @@ func (c *Client) SubmitOrder(ctx context.Context, o order.Order,
 		MaxBatchFeeRateSatPerKw: uint64(o.Details().MaxBatchFeeRate),
 	}
 
+	details.AllowedNodeIds = make(
+		[][]byte, len(o.Details().AllowedNodeIDs),
+	)
+	for idx, nodeID := range o.Details().AllowedNodeIDs {
+		details.AllowedNodeIds[idx] = nodeID[:]
```
bhandras, line 438: Not sure if this is safe? Since `nodeID` is restricted to the scope of the for loop.

bhandras (approved):
Great PR @positiveblue just a few nits, otherwise LGTM :tada:

auctioneer/client.go
positiveblue: Good intuition, In Go, the runtime is the one that decides what goes in the stack/heap (the specification does not define it). So, even for pointers, the obj survives the scope.

cmd/pool/order.go
```diff
@@ -281,6 +321,26 @@ func parseAccountKey(ctx *cli.Context, args cli.Args) ([]byte, error) {
...
+//
+// NOTE: the parameter must contain a string slice. The strings are hex decoded
+// but not parsed as a btcec.PublicKey.
+func parseNodePubKeySlice(ctx *cli.Context, key string) ([][]byte, error) {
+	hexNodeIDs := ctx.StringSlice(key)
+	nodeIDs := make([][]byte, 0, len(hexNodeIDs))
+	for _, hexNodeID := range hexNodeIDs {
+		nodeID, err := hex.DecodeString(hexNodeID)
+		if err != nil {
+			return nil, fmt.Errorf("invalid node ID %v: %v", nodeID,
```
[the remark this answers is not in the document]
positiveblue: Much better :+1:

cmd/pool/order.go
```diff
@@ -116,6 +116,20 @@ var sharedFlags = []cli.Flag{
...
 	},
+	cli.StringSliceFlag{
+		Name: "allowed_node_id",
+		Usage: "the list of nodes this order is allowed to match " +
+			"with; if empty, the order will be able to match " +
+			"with any node unless not_allowed_node_id is set. " +
+			"Can be specified multiple times",
+	},
+	cli.StringSliceFlag{
+		Name: "not_allowed_node_id",
```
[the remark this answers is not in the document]
positiveblue: Not sure about this one. I could change it only in the CLI, but it could leak from some `rpc_server` errors and potentially confuse the user.  For example, parsing the order could return `invalid not_allowed_node_id`. 

Is it because for native speakers `prohibited_node_id` sounds better than `not_allowed_node_id`? If so I can change it everywhere, just not sure if it is just a matter of taste.

clientdb/order.go
positiveblue: +1 to this sanity check

auctioneer/client.go
bhandras: I actually ran a little experiment, because I wasn't quite sure myself. PTAL: https://go.dev/play/p/geeXf0C5E6m

positiveblue: ohhh,  you are right. The var will out live the loop but all the elements in the slice of `[]byte` will be equal to the last one!
:pray: Thank you for catching this @bhandras

bhandras: It's tricky and I think the compiler/linter should be more vocal about this as it's easy to miss.

cmd/pool/order.go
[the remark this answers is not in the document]
bhandras: Yeah, I'm not sure either given I'm not native but just thought it's simpler as "prohibited" is just one word. Regardless I'm fine with the current choice too.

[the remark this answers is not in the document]
guggero: Prohibited sounds good to me. Or just disallowed. "not allowed" does sound a bit bumpy.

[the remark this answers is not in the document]
guggero: But this probably also requires quite a big refactor/renaming so we can also just leave it the way it is.

guggero (commented):
Code looks good to me now!

Though I think we should wait with merging this until we have the server version deployed! Otherwise users will still get matched according to the old rules and be unsure why. Perhaps we should even go ahead and add a new order version for this?

auctioneer/client.go
guggero: Oh wow, good catch! I wouldn't have caught that one :open_mouth:

cmd/pool/order.go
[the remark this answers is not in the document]
positiveblue: Did you already have the tab for the review open? It looks like it was already fixed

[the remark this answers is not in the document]
guggero: Oh yeah. Old comment, sorry. Please ignore :sweat_smile:

Issue #349, opened 2022-03-27
Ask orders don't get fulfilled


rkfg:
I'm not sure what I'm doing wrong but I haven't got a single ask order fulfillment despite having them waiting for days (one bid order was fulfilled though and I got a channel opened to me). I'm using Lightning Terminal and since order creation is currently completely broken I used CLI to create these two ask orders.
![2022-03-27_15-47-24](https://user-images.githubusercontent.com/184066/160282136-642ac9a7-5a1e-40ca-baf5-4c174a83a353.png)
There are plenty transactions with much higher BPS than what I offer, they're recent (14-15 hours ago) and my offer is a week old already. But they never get picked by the auctioneer. The order dump (one of them, the other is similar):
```json
                {
                        "details": {
                                "trader_key": "-----------",
                                "rate_fixed": 1488,
                                "amt": "1600000",
                                "max_batch_fee_rate_sat_per_kw": "25000",
                                "order_nonce": "-----------",
                                "state": "ORDER_SUBMITTED",
                                "units": 16,
                                "units_unfulfilled": 16,
                                "reserved_value_sat": "1858032",
                                "creation_timestamp_ns": "1647772381608079026",
                                "events": [
                                ],
                                "min_units_match": 1,
                                "channel_type": "ORDER_CHANNEL_TYPE_UNKNOWN"
                        },
                        "lease_duration_blocks": 2016,
                        "version": 5
                }
```
I created it using the following command (my key is censored here): `./pool --rpcserver=localhost:8443 --tlscertpath=~/.lit/tls.cert o s ask 1600000 <removed> --interest_rate_percent 0.3 --lease_duration_blocks=2016 --min_chan_amt 100000`

I read the docs but either I miss something here or there's a bug. Any hints?

guggero:
Is your node a TIER_1 node? Is your node Tor only? Do you see connection attempts in your `lnd` log that fail? What does your Pool log look like?

rkfg:
My node is T1, it uses Tor but also has a WAN IP. There are indeed failed connections like `unable to read message` for various peer, what exactly should I look for and from what address? I use Lightning Terminal with Pool integrated, there's nothing interesting in its log. It's full of stuff like this:
```
2022-03-26 20:35:47.732 [INF] RPCS: Received new block notification: height=729136
2022-03-26 20:35:48.304 [INF] LITD: Handling gRPC web request: /poolrpc.Trader/ListOrders
[...]

PR #354, opened 2022-04-11
rpcserver: roll back #330

guggero:
To fix an accidental market segregation for lnd 0.14.x users that didn't
set an explicit channel type in their bid orders, we roll back the
changes from #330.
We can re-enable this default value selection once a large number of ask
orders support the new channel type.

Fixes https://github.com/lightninglabs/pool/issues/349.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1303,7 +1303,13 @@ func (s *rpcServer) SubmitOrder(ctx context.Context,
 			s.server.lndServices.Version, scriptEnforceVersion,
 		)
 		if verErr == nil {
-			return order.ChannelTypeScriptEnforced
+			// TODO(guggero): Use order.ChannelTypeScriptEnforced
+			// here as soon as a large percentage of ask orders
+			// support script enforced channel types to not cause a
+			// market segregation for lnd 0.14.x or later users
+			// without them being aware of why their bids aren't
+			// getting matched.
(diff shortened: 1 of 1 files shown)


positiveblue (approved):

Roasbeef (approved):
LGTM ⛩

Roasbeef:
What if we also made this default commitment type be `ChannelTypeScriptEnforced` if neither side explicitly specified a type? This way we don't fragment the order book (this rollback), but can optimistically promote the latest channel type (script enforce in this case) from the PoV of the maker (the one opening the channel in the first place).

guggero:
> What if we also made this default commitment type be `ChannelTypeScriptEnforced` if neither side explicitly specified a type? This way we don't fragment the order book (this rollback), but can optimistically promote the latest channel type (script enforce in this case) from the PoV of the maker (the one opening the channel in the first place).

I think the problem with this is that we don't know if both nodes actually _support_ the script enforced channel type on the `lnd` level. That's why we added this version based bump in the first place, to signal readiness. Or did I misunderstand what you meant?

Roasbeef:
> I think the problem with this is that we don't know if both nodes actually support the script enforced channel type on the lnd level.

Assuming we can use the order versioning sent as a proxy, then I think that should work?

Issue #357, opened 2022-04-13
Unable to connect to the auctioneer (`no order found`)


positiveblue:
**Describe the issue**
Some users are unable to trade because the auctioneer ends the connection with the client. The client has two accounts that get disconnected after starting `poold` because of the server getting an unrecoverable error (`[/poolrpc.ChannelAuctioneer/OrderState]: rpc error: code = Unknown desc = no order found`)

**What behavior did you expect?**

The user is able to start the client and participate in trades for the orders that the server stored properly. Even if we consider this a critical error (it is a state that should never happen) `poold` should not stop without given the user a chance to fix the problem (cancel the order? force sync orders with the server?)

**What was the actual behavior (provide screenshots if the issue is UI related)?**
The user cannot restart the deamon. It exits during the [startup process](https://github.com/lightninglabs/pool/blob/master/server.go#L773)

relevant logs:

```
2022-04-12 23:10:48.972 [INF] RPCS: Trader server is now active
2022-04-12 23:10:49.043 [INF] POOL: Starting gRPC listener
2022-04-12 23:10:49.044 [INF] POOL: Starting REST proxy listener
2022-04-12 23:10:49.068 [INF] POOL: RPC server listening on [::]:8443
2022-04-12 23:10:49.068 [INF] POOL: REST proxy listening on 127.0.0.1:828
...
2022-04-12 23:10:50.519 [INF] AUCT: Shutting down auctioneer client
2022-04-12 23:10:50.519 [DBG] AUCT: Closing server stream
2022-04-12 23:10:50.519 [TRC] AUCT: Received msg=(*auctioneerrpc.ServerAuctionMessage)(nil), err=rpc error: code = Canceled desc = context canceled from server
2022-04-12 23:10:50.519 [INF] RPCS: Trader server stopping
2022-04-12 23:10:50.519 [INF] AUCT: Shutting down auctioneer client
...
2022-04-12 23:28:39.013 [DBG] LNDC: Closing lnd connection
2022-04-12 23:28:39.013 [DBG] LNDC: Wait for client to finish
2022-04-12 23:28:39.013 [DBG] LNDC: Wait for chain notifier to finish
2022-04-12 23:28:39.014 [DBG] LNDC: Wait for invoices to finish
2022-04-12 23:28:39.014 [DBG] LNDC: Wait for router to finish
2022-04-12 23:28:39.014 [DBG] LNDC: Lnd services finished
[...]

PR #358, opened 2022-04-13
cmd/debug: add DeleteOrder command

positiveblue:
Enable deleting orders in the local db using the debug cli. This can be used to clean the local db in case it gets out of sync with the server.

**DANGER**: if the user deletes an order from the local db and that order exists in the server, the user will get disconnected when that order matches (and won't be able to trade). The only solution then is to nuke the local db, restore the account and recreate the orders.

unblocks #357

### DIFF
--- a/order/mock_test.go
+++ b/order/mock_test.go
@@ -83,8 +83,8 @@ func (s *mockStore) GetOrders() ([]Order, error) {
 	return orders, nil
 }
 
-// DelOrder removes the order with the given nonce from the local store.
-func (s *mockStore) DelOrder(nonce Nonce) error {
+// DeleteOrder removes the order with the given nonce from the local store.
+func (s *mockStore) DeleteOrder(nonce Nonce) error {
 	delete(s.orders, nonce)
 	return nil
 }
--- a/server.go
+++ b/server.go
@@ -770,7 +770,8 @@ func (s *Server) syncLocalOrderState() error {
 			context.Background(), orderNonce,
 		)
 		if err != nil {
-			return fmt.Errorf("unable to fetch order state: %v", err)
+			return fmt.Errorf("unable to fetch order(%v): %v",
+				orderNonce, err)
 		}
 		remoteOrderState, err := rpcOrderStateToDBState(
 			orderStateResp.State,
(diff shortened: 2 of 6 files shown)


order/mock_test.go
```diff
@@ -83,8 +83,8 @@ func (s *mockStore) GetOrders() ([]Order, error) {
 	return orders, nil
 }
 
-// DelOrder removes the order with the given nonce from the local store.
-func (s *mockStore) DelOrder(nonce Nonce) error {
+// DeleteOrder removes the order with the given nonce from the local store.
```
positiveblue, line 86: `DelOrder` wasn't used anywhere

server.go
```diff
@@ -770,7 +770,8 @@ func (s *Server) syncLocalOrderState() error {
 			context.Background(), orderNonce,
 		)
 		if err != nil {
-			return fmt.Errorf("unable to fetch order state: %v", err)
+			return fmt.Errorf("unable to fetch order(%v): %v",
```
positiveblue, line 773: This is needed so the user knows which ones are the offender orders

guggero (approved):
Code looks good to me. Though I wonder if updating the state to `StateCanceled` wouldn't also solve the issue for the user and fail a bit more gracefully if the order was ever matched?

positiveblue:
I think it would not. The problem this is is trying to solve is when the user has some orders in the local db that the server does not know about so those ones need to be deleted (the order sync fails). Updating the status to Cancel would make the client fail during the start up process too (unable to sync that order anyway).

This is just an extra `debug` subcommand that we can use. It would be nice to have a reconciliation process during the start ups:
- mark up local invoices that do not exist in the server side as missing 
- fetch orders that exist in the server but not in the local db
- overwrite local status with the one from the server for existing invoices (what we do now)

Roasbeef:
>  It would be nice to have a reconciliation process during the start ups:

Agreed, we do something similar today when it comes to syncing up batches, but don't do this for orders. The great part about this design is that the server can't _spoof_ orders since they're all signed by the trader.

Roasbeef (approved):
LGTM 🛤

Should also get some basic itest coverage on the server side as well.

PR #359, opened 2022-04-13
docker: bump go version + set arch platform

positiveblue:
Some users were not able to build [multi arch images with docker](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/). It looks like we need to propagate the `${BUILDPLATFORM}` to all the docker stages.

I also had to bump the go version to make this work `docker buildx build --platform linux/arm64/v8 -f Dockerfile .`, not sure if something changed in the last three versions of go for arm :eyes:


guggero (approved):
tACK, LGTM. Thanks for the fix!

PR #360, opened 2022-04-19
order: Mark missing orders as failed in the client db

positiveblue:
The client and server db can get out of sync if some orders are not recorded properly in the server. This becomes problematic because the deamon won't start until all the bad invoices have been cleaned out. Right now the clients need to fix these edge cases manually.

The idea is to simply mark those invoices as failed and let the client continue with the start up.

### DIFF
--- a/server.go
+++ b/server.go
@@ -754,6 +754,16 @@ func (s *Server) syncLocalOrderState() error {
 		noncesToUpdate []order.Nonce
 		ordersToUpdate [][]order.Modifier
 	)
+
+	addUpdate := func(nonce order.Nonce, state order.State) {
+		noncesToUpdate = append(noncesToUpdate, nonce)
+		ordersToUpdate = append(
+			ordersToUpdate,
+			[]order.Modifier{
+				order.StateModifier(state),
+			},
+		)
+	}
 	for _, dbOrder := range dbOrders {
 		orderNonce := dbOrder.Nonce()
 		localOrderState := dbOrder.Details().State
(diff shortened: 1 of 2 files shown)


clientdb/order.go
```diff
@@ -65,6 +66,13 @@ var (
 	orderTlvKey = []byte("order-tlv")
 )
 
+// IsNotFoundErr tries to match the given error with ErrNoOrder.
+func IsOrderNotFoundErr(err error) bool {
```
positiveblue, line 70: It looks like we are not using grpc error codes. I am not sure if there is any other way to check for what type of error we are receiving from the server that string matching here...

guggero: Yeah, we currently don't use any custom codes on the server side. String matching is okay here. Though I'm not sure if this function better fits into the `auctioneer` package?

server.go
```diff
@@ -770,9 +780,22 @@ func (s *Server) syncLocalOrderState() error {
...
 		if err != nil {
-			return fmt.Errorf("unable to fetch order(%v): %v",
-				orderNonce, err)
+			switch {
+			// If the server does not know about this order mark it
+			// as failed and allow the client start up with the
+			// rest of valid orders.
+			//
+			// NOTE: This should never happen.
+			case clientdb.IsOrderNotFoundErr(err):
```
guggero, line 789: Yeah, reading this I first thought this meant "the local DB didn't find the order". So moving the function to the auctioneer package should make that more clear.

server.go
```diff
@@ -770,9 +780,22 @@ func (s *Server) syncLocalOrderState() error {
...
 		)
 		if err != nil {
-			return fmt.Errorf("unable to fetch order(%v): %v",
-				orderNonce, err)
+			switch {
+			// If the server does not know about this order mark it
+			// as failed and allow the client start up with the
+			// rest of valid orders.
+			//
+			// NOTE: This should never happen.
```
guggero, line 788: nit: "This should never happen under normal conditions. This was added to handle a specific failure that's being investigated"?

guggero (approved):
Nice fix, LGTM :tada:

bhandras (approved):
LGTM, sorry for the late review, was buried under my other notifications :+1:

PR #363, opened 2022-05-05
poolscript: make sure negation works with btcec v2

guggero:
Something I noticed when taking a closer look at how the btcec V2 library works.
Not sure if this is really necessary or not, any thoughts, @Roasbeef?
The unit tests all passed fine without this, but maybe this could be an edge case?

### DIFF
--- a/poolscript/script.go
+++ b/poolscript/script.go
@@ -304,9 +304,13 @@ func DecrementKey(pubKey *btcec.PublicKey) *btcec.PublicKey {
 	// Multiply G by 1 to get G.
 	secp.ScalarBaseMultNonConst(new(secp.ModNScalar).SetInt(1), &g)
 
-	// Get -G by negating the Y axis.
+	// Get -G by negating the Y axis. We normalize first, so we can negate
+	// with the magnitude of 1 and then again to make sure everything is
+	// normalized again after the negation.
+	g.ToAffine()
 	g.Y.Normalize()
 	g.Y.Negate(1)
+	g.Y.Normalize()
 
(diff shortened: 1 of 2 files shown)


poolscript/script.go
```diff
@@ -304,9 +304,13 @@ func DecrementKey(pubKey *btcec.PublicKey) *btcec.PublicKey {
 	// Multiply G by 1 to get G.
 	secp.ScalarBaseMultNonConst(new(secp.ModNScalar).SetInt(1), &g)
 
-	// Get -G by negating the Y axis.
+	// Get -G by negating the Y axis. We normalize first, so we can negate
```
Roasbeef, line 307: Are we doing this anywhere else across our packages? Agree that it's safer to first map to affine coordinate, and one _always_ needs to normalize after modifying any of the field elements. 

One way we can try to trigger a failure here would be using a quickcheck test which may eventually hit an edge case that causes an issue to be triggered here.

guggero: > Are we doing this anywhere else across our packages?

Not that I'm aware of. But I need to check the `lnd` codebase that we always to the normalize after modifying field elements.

I added some quickcheck tests and also increased the number of iterations. Both works fine with and without this diff. So I guess it can't hurt to add it anyway?

Roasbeef: Yeah agreed, in the future we should spin this out into methods in the `btcec` package to make things generally easy to use.

Roasbeef (approved):
LGTM 🐇

PR #368, opened 2022-06-02
mod: bump auctioneerrpc and aperture versions

guggero:
We need to bump our auctioneerrpc version so it's correctly used when
Pool is pulled in as a library in LiTd. To make things easier, we also
bump aperture to the latest version.

#### Pull Request Checklist
  used.


guggero:
cc @dstadulis (part of LiT update I mentioned).

positiveblue (approved):

PR #369, opened 2022-06-08
Add forgotten server account state

guggero:
We add a state that existed on the server side for a while but we never exposed on the RPC level.

We'll push a new tag `auctioneerrpc/v1.0.7` after this PR is merged.

#### Pull Request Checklist
  used.


positiveblue (approved):
LGTM

I wonder if there is a tool for checking this things (that we have the relative field in a proto for a iota in Go) :eyes:

PR #370, opened 2022-06-09
Switch account signing over to PSBT, allow deposits/account creation from P2TR inputs

guggero:
Depends on https://github.com/lightninglabs/lndclient/pull/105.

This PR moves over the signing operations from `SignOutputRaw` and `ComputeInputScript` to `SignPsbt` and `FinalizePsbt` in order to prepare for Taproot/MuSig2 compatible trader accounts.

With this change we also allow using P2TR wallet inputs for creating new accounts or for depositing into existing accounts.

#### Pull Request Checklist
  used.


account/manager.go
```diff
@@ -1677,11 +1666,21 @@ func (m *manager) signSpendTx(ctx context.Context, account *Account,
 
 	ourSig := pIn.PartialSigs[0].Signature
 
-	// We temporarily set the final witness to the partial sig to allow
-	// extraction of the final TX.
-	pIn.FinalScriptWitness, err = serializeWitness([][]byte{ourSig})
+	// We temporarily set the final witness to the partial sig to allow the
+	// extraction of the final TX. Unless we're using the expiry path in
+	// which case we _can_ create the full and final witness.
```
positiveblue, line 1671: :+1:

positiveblue (approved):

account/manager.go
```diff
@@ -1786,7 +1837,7 @@ coinSelection:
 	for {
 		utxos, err := m.cfg.Wallet.ListUnspent(ctx, 1, math.MaxInt32)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
```
Roasbeef, line 1839: Can't we just cut this out all together? Since we can reply in the wallet to just do `FundPsbt` and handle all the UTXO selection aspects?

account/manager.go
```diff
@@ -1839,20 +1890,70 @@ coinSelection:
...
+
+	// Add all newly added inputs that represent the deposit amount (plus
+	// change).
+	for _, coin := range inputs {
+		packet.UnsignedTx.TxIn = append(
+			packet.UnsignedTx.TxIn, &wire.TxIn{
+				PreviousOutPoint: coin.OutPoint,
+			},
+		)
+		packet.Inputs = append(packet.Inputs, psbt.PInput{
```
Roasbeef, line 1921: Same here re just letting `walletrpc` handle the PSBT specific semantics?

account/manager.go
```diff
@@ -1931,35 +2031,11 @@ func locateAccountInput(tx *wire.MsgTx, account *Account) (int, error) {
...
-}
-
-// signAccountInput signs the account input in the spending transaction of an
+// decorateAccountInput signs the account input in the spending transaction of an
 // account. If the account is being spent with cooperation of the auctioneer,
 // their signature will be required as well.
-func (m *manager) signAccountInput(ctx context.Context, tx *wire.MsgTx,
-	account *Account, idx int, sigHashType txscript.SigHashType) ([]byte,
-	[]byte, error) {
+func (m *manager) decorateAccountInput(account *Account, packet *psbt.Packet,
```
Roasbeef, line 2036: `godoc` can be updated a bit: this doesn't aisng anylonger, but adds the pool specific PSBT information so it can sign properly.

account/manager.go
```diff
@@ -2084,3 +2163,24 @@ func makeTxnLabel(labelPrefix, contextLabel string) string {
 
 	return label
 }
+
+// serializeWitness turns a wire witness into its serialized form.
+func serializeWitness(witness wire.TxWitness) ([]byte, error) {
+	var buf bytes.Buffer
+	if err := psbt.WriteTxWitness(&buf, witness); err != nil {
```
Roasbeef, line 2169: Ah cool I was actually looking for something like this in a related project (along w/ reading an outpoint serialized in bitcoin wire format), nice to know it was in the `psbt` package.

account/coin_selection.go
```diff
@@ -68,6 +68,10 @@ func coinSelection(coins []chanfunding.Coin, amt btcutil.Amount,
 				weightEstimator.AddP2WKHInput()
 			case txscript.IsPayToScriptHash(utxo.PkScript):
 				weightEstimator.AddNestedP2WKHInput()
+			case txscript.IsPayToTaproot(utxo.PkScript):
```
Roasbeef, line 71: Similar Q here re just letting the PSBT funder to all the coin selection.

account/manager.go
```diff
@@ -1422,12 +1422,14 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
...
 
-		spendPkg, err = m.spendAccountExpiry(
-			ctx, account, packet, bestHeight,
+		spendPkg, err = m.signSpendTx(
+			ctx, account, packet, bestHeight, witnessType,
 		)
 
 	case multiSigWitness:
-		spendPkg, err = m.signSpendTx(ctx, account, packet, 0)
+		spendPkg, err = m.signSpendTx(
```
Roasbeef, line 1430: I think this can be further simplified to just calling `signSpendTx` while passing the `witnessType` in directly.

Roasbeef: Also here is there any reason why we wouldn't just always set the height here?

account/manager.go
```diff
@@ -1630,38 +1652,69 @@ func (m *manager) createSpendTx(ctx context.Context, account *Account,
...
-	// Our account input signature isn't always all that's required to spend
-	// it, so we'll take care of forming a proper signature later.
-	witnessScript, ourSig, err := m.signAccountInput(
-		ctx, tx, account, accountInputIdx, sigHashType,
-	)
+	ourSig := pIn.PartialSigs[0].Signature
+
+	// We temporarily set the final witness to the partial sig to allow
+	// extraction of the final TX.
+	pIn.FinalScriptWitness, err = serializeWitness([][]byte{ourSig})
```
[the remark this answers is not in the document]
guggero: Yeah, the signing step only adds a raw signature to the PSBT but we need to create the full witness stack, which might also require us to call the auctioneer for its sig. So here we just add something to the final script witness so we can convert the PSBT into a normal TX (which does the work for us in terms of assembling the signatures for the other inputs in case of a deposit).

account/manager.go
```diff
@@ -1630,38 +1652,69 @@ func (m *manager) createSpendTx(ctx context.Context, account *Account,
...
+	// normal wallet and can be signed for without additional PSBT metadata
+	// fields.
+	var signedTx *wire.MsgTx
+	if signedPacket.IsComplete() {
+		err = psbt.MaybeFinalizeAll(signedPacket)
+		if err != nil {
+			return nil, fmt.Errorf("error finalizing PSBT: %v", err)
+		}
+
+		signedTx, err = psbt.Extract(signedPacket)
```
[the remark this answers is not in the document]
guggero: Yes, the TX will be invalid from the point of view of the chain but not the PSBT library. We just want to turn the PSBT into a raw TX, which we can only do if we set something as the sig. We'll fix this in a later step (and in a later commit we also simplify this part a bit).

account/manager.go
guggero: No, I don't think there's a reason. We should always be able to set the lock time to the current height. Fixed.

guggero:
Addressed all comments but something doesn't seem to be working just yet. Will debug further tomorrow.

account/coin_selection.go
```diff
@@ -1,102 +0,0 @@
-package account
```
Roasbeef, line 1: 😈

account/manager.go
```diff
@@ -1754,105 +1818,112 @@ func valueAfterAccountUpdate(account *Account, outputs []*wire.TxOut,
...
-	*wire.TxOut, error) {
+func (m *manager) inputsForDeposit(ctx context.Context, account *Account,
+	newAccountOutput *wire.TxOut, depositAmount btcutil.Amount,
+	witnessType witnessType, feeRate chainfee.SatPerKWeight) (*psbt.Packet,
+	func(), error) {
+
+	// Unfortunately the FundPsbt call doesn't allow us to specify _any_
+	// inputs, otherwise it won't perform coin selection at all. So what we
+	// do instead is to fund our account output just for the funding amount
+	// plus whatever we need to pay for the additional input (which we know
```
Roasbeef, line 1826: Nice work around! We'll probably eventually patch this behavior, but good to be abble to get by w/ less as well.

account/manager.go
```diff
@@ -1754,105 +1818,112 @@ func valueAfterAccountUpdate(account *Account, outputs []*wire.TxOut,
...
+	if err != nil {
+		return nil, nil, err
+	}
+	acctInputWeight := int64(acctInputEstimator.Weight())
+	acctInputFee := feeRate.FeeForWeight(acctInputWeight)
 
-	// We'll start by obtaining our global lock ID.
-	lockID, err := m.cfg.Store.LockID()
+	outputToFund := &wire.TxOut{
+		Value:    int64(depositAmount + acctInputFee),
```
Roasbeef, line 1838: This _should_ be ok, but part of me wonders if there's some weird edge case here w/ rounding and amounts. Worst case, I guess we get rejected due to having insufficient fees.

Roasbeef (commented):
Love all the code deletion in this latest diff! Happy we were able to work around the lack of functionality on the lnd PSBT side. I think we'll need to do something similar on a new project we're working on as well

account/manager.go
guggero: Yeah, we _should_ always over estimate by a little bit anyway since most of the time the signatures don't end up being the full 73 bytes. But I went ahead and added a min relay fee check anyway.

guggero:
Okay, after spending way too much time finding out that my problem was actually https://github.com/lightningnetwork/lnd/issues/6626 I created the fix in https://github.com/lightningnetwork/lnd/pull/6687.
Now we have the problem that we can't accept deposits from NP2WKH inputs if you don't have `lnd v0.15.1-beta` or later.
I added a workaround that just errors out in that case (which is important to do _before_ we contact the auctioneer, otherwise we'll end up with an account diff that never confirms).

I hope that workaround is okay!

dstadulis:
Done, and ready for review from @positiveblue  @Roasbeef  https://github.com/lightninglabs/pool/pull/370 https://github.com/lightninglabs/subasta/pull/410

Roasbeef:
> I added a workaround that just errors out in that case (which is important to do before we contact the auctioneer, otherwise we'll end up with an account diff that never confirms).

Ahh, makes sense. I think this is OK, as it lets us have the code be uniform vs special casing the `np2wkh` case. TBH, I don't think it's used all that often these days, since even `bech32m` usage is on the rise.

Roasbeef (approved):
LGTM 🎊

positiveblue (approved):
:rocket:

account/manager.go
```diff
@@ -2080,3 +2226,24 @@ func makeTxnLabel(labelPrefix, contextLabel string) string {
...
+func serializeWitness(witness wire.TxWitness) ([]byte, error) {
+	var buf bytes.Buffer
+	if err := psbt.WriteTxWitness(&buf, witness); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+// sigHashForScript returns the sighash flag for the given UTXO's pkScript.
```
positiveblue, line 2240: May be worth [adding this to btcd](https://github.com/btcsuite/btcd/pull/1868) for default script types

PR #375, opened 2022-06-16
Upgrade trader accounts to Taproot/MuSig2

guggero:
Depends on https://github.com/lightninglabs/pool/pull/370.

#### Pull Request Checklist
  used.


account/manager.go
```diff
@@ -1679,42 +1839,122 @@ func (m *manager) signSpendTx(ctx context.Context, account *Account,
...
+	sessionInfo, err := poolscript.TaprootMuSig2SigningSession(
+		ctx, account.Expiry, account.TraderKey.PubKey, account.BatchKey,
+		account.Secret, account.AuctioneerKey, m.cfg.Signer,
+		&account.TraderKey.KeyLocator,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Don't leave incomplete sessions in memory.
```
positiveblue, line 1871: :+1:

account/manager.go
```diff
@@ -1679,42 +1839,122 @@ func (m *manager) signSpendTx(ctx context.Context, account *Account,
...
+				sessionInfo.SessionID[:], err)
 		}
-	} else {
-		// We should be able to extract the final TX now, even if the
-		// witness isn't yet fully correct just yet.
-		_, signedTx, err = m.cfg.Wallet.FinalizePsbt(
-			ctx, signedPacket, "",
+	}
+	defer func() {
+		// Can't defer cleanupSession() directly since we might change
```
positiveblue, line 1880: nit: maybe we can bind a variable and do something like

```
var succeeded = false

defer func() {
    if succeeded {
        // nothing to do.
        return
    }
    // clean up session...
}()
```

server.go
```diff
@@ -832,3 +830,33 @@ func (s *Server) syncLocalOrderState() error {
...
+	}
+
+	// We can only auto-upgrade accounts to Taproot if the underlying node
+	// has the required capabilities.
+	if isTaprootCompatible {
+		return order.UpgradeAccountTaprootBatchVersion
+	}
+
+	// Fall back to the previous default version.
+	return order.LatestBatchVersion
```
positiveblue, line 861: q: can this logic bring us problems in the near future? If we add a new batch version, ex: private channels. We can have a new pool version with private channels but not taproot (using an old lnd version). But we can also have a new pool version with private channels AND taproot :eyes: 

Applies to ` SupportsAccountTaprootUpgrade()` too.

rpcserver.go
```diff
@@ -2269,8 +2269,15 @@ func (s *rpcServer) prepareLeasesResponse(ctx context.Context,
 
 			// Estimate the chain fees paid for the number of
 			// channels created in this batch and tally them.
+			//
+			// TODO(guggero): This is just an approximation! We
```
positiveblue, line 2296: :+1:

positiveblue (commented):
First pass done, this looks really good @guggero :clap:

account/interfaces.go
```diff
@@ -586,20 +629,23 @@ type Manager interface {
 	// to lnd may be added to the deposit transaction.
 	DepositAccount(ctx context.Context, traderKey *btcec.PublicKey,
 		depositAmount btcutil.Amount, feeRate chainfee.SatPerKWeight,
-		bestHeight, expiryHeight uint32) (*Account, *wire.MsgTx, error)
+		bestHeight, expiryHeight uint32, newVersion Version) (*Account,
```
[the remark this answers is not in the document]
guggero: I used the same strategy as we did for the account expiry extension. I think it's nice to be able to extend and/or upgrade your account with any of the possible modifications. So IMO we should give the user the option to upgrade outside of batches.

account/manager.go
guggero: Ah yes, much cleaner that way.

guggero: Simplified the whole MuSig2 signing quite a bit so we can re-use the same code on both client and server side, which made this easier as well.

guggero:
I addressed all comments and got almost all itests working on the server side.
This is now officially not WIP anymore :tada:

JESSENIA1983 (approved):

clientdb/account.go
```diff
@@ -10,6 +10,14 @@ import (
 	"go.etcd.io/bbolt"
 )
 
+const (
+	// accountStateVersionedMask is a bit mask for detecting from the state
+	// of an account whether that account has a version field encoded with
+	// it or not. We use the first bit of the uint8 state field because we
+	// are unlikely to ever have more than 127 different states.
+	accountStateVersionedMask account.State = 0b1000_0000
```
positiveblue, line 18: I fail to see why we don't use TLV fiels in a new bucket for this and future updates

positiveblue (approved):
LGTM :taco: 

There is a small comment about how we are storing the new field but nothing blocking.

I would add the version field in multiple rpc calls, like `ListAccounts`. Also it could be useful to add it to the terminal confirmation message before creating an account. However, I can add those two things later :+1:

order/batch.go
```diff
@@ -51,8 +51,27 @@ const (
 	// ExtendAccountBatchVersion is the first version where accounts expiry
 	// are extended after participating in a batch.
 	ExtendAccountBatchVersion BatchVersion = 1
+
+	// UpgradeAccountTaprootBatchVersion is the batch version where accounts
+	// are automatically upgraded to Taproot accounts. We leave a gap up to
```
Roasbeef, line 56: 👍

order/batch_storer.go
```diff
@@ -110,6 +110,17 @@ func (s *batchStorer) StorePendingBatch(batch *Batch) error {
 				)
 			}
 
+			// The account version can be upgraded during the batch
+			// if the client supports it.
+			if batch.Version.SupportsAccountTaprootUpgrade() &&
```
Roasbeef, line 115: Ideally we'd add a test/assertion for this upgrade path, though I think this is already covered w/ the backend itests?

order/batch.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
-	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr/musig2"
```
Roasbeef, line 11: In commit message: oder -> order

Roasbeef (changes requested):
Amazing PR!! 

Love how we're able to handle all the upgrade logic behind the scenes, hidden away from the user. This change will also make it even _cheaper_ (on chain) to buy/sell leases as well, since the account witness is much smaller. 

Completed an initial pass, with the most major comments being some wanted fuzzing/testing of the new witness parsing logic as well as all the new scripts added for taproot.

clientdb/account.go
```diff
@@ -156,9 +181,15 @@ func readAccount(sourceBucket *bbolt.Bucket,
 }
 
 func serializeAccount(w *bytes.Buffer, a *account.Account) error {
+	rawState := a.State
+	accountIsVersioned := a.Version > account.VersionInitialNoVersion
+	if accountIsVersioned {
+		rawState = setVersionBit(rawState)
```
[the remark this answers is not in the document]
guggero: No, the version wasn't stored before. So if we read an existing account, we don't know whether to read an additional byte for the version or not. And because the accounts are embedded within the same data blob for batch snapshots, we also can't use the EOF trick. We could use a separate bucket with additional fields. But that just makes it way more complicated as we'd need to treat the normal accounts and those embedded within the batch snapshots differently.
But I am going to add TLV for all new fields, so adding new ones will become easier.

clientdb/account.go
```diff
@@ -21,6 +29,23 @@ var (
...
+
+// setVersionBit sets the version bit in the given account state.
+func setVersionBit(state account.State) account.State {
+	return state | accountStateVersionedMask
+}
+
+// clearVersionBit clears the version bit in the given account state.
+func clearVersionBit(state account.State) account.State {
+	// The &^ operator means AND NOT, also known as the Bitclear operator.
+	return state &^ accountStateVersionedMask
```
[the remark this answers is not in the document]
guggero: Yes, see explanation above.

poolscript/log.go
```diff
@@ -0,0 +1,25 @@
+package poolscript
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+const Subsystem = "SCRP"
```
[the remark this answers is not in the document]
guggero: I did, but in a later commit. Makes sense to move it here though, so I did that.

account/manager.go
```diff
@@ -1252,6 +1265,12 @@ func (m *manager) RenewAccount(ctx context.Context,
 		return nil, nil, err
 	}
 
+	// Can't downgrade an account.
+	if newVersion < account.Version {
```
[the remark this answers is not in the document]
guggero: `validateAccountParams` is only called for new accounts, where there is only one version. And the logic for withdraw, deposit and renew are all slightly different, so I'd rather keep this more explicit.

clientdb/account.go
```diff
@@ -175,19 +206,36 @@ func serializeAccount(w *bytes.Buffer, a *account.Account) error {
 		}
 	}
 
+	// The version flag encoded within the state will inform the
+	// de-serialize method that it should read another field. Therefore, we
+	// can safely write it here.
+	if accountIsVersioned {
+		if err := WriteElement(w, a.Version); err != nil {
```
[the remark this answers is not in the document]
guggero: Good idea. Added the new field as a TLV stream.

guggero:
I addressed all comments except for the additional unit tests. I agree they should be added, but I'm going to work on them next week.

dstadulis:
might need to change unit test frame work

poolscript/script.go
```diff
@@ -277,6 +437,30 @@ func IsExpirySpend(witness wire.TxWitness) bool {
 	return true
 }
 
+// IsTaprootExpirySpend determines whether the provided witness corresponds to
+// the expiration script path of a Taproot enabled (version 1) account.
+func IsTaprootExpirySpend(witness wire.TxWitness) bool {
```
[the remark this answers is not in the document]
guggero: I added a quick test and a fuzz test for all the witness detection functions. I ran the fuzz test for a couple of million executions, looks good so far.

poolscript/script.go
```diff
@@ -277,6 +437,30 @@ func IsExpirySpend(witness wire.TxWitness) bool {
...
+	if len(witness[2]) < 33 {
+		return false
+	}
+
+	// The control block must start with the leaf version, which will always
+	// be 0xc0 in our case.
+	if witness[2][0] != byte(txscript.BaseLeafVersion) {
+		return false
+	}
+	return true
```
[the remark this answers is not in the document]
guggero: We don't know the full script here as we don't know the key or expiry. But we can make more assumptions about the length and some of the opcodes being used.

order/batch_storer.go
guggero: Yes, the various upgrade paths are pretty well covered in integration tests. And also in the account manager test.

poolscript/script.go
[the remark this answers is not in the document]
guggero: I removed the fuzz test again, because that would mean we'd need to update to Go 1.18 in CI which requires a few other changes because of the linter. So I'll add that in a separate PR.

guggero:
I added a set of unit tests for the new spend paths. And I also ran the fuzzer locally for almost 2 hours and didn't find anything so far.

poolscript/script_test.go
```diff
@@ -89,3 +107,337 @@ func TestIncrementBatchKey(t *testing.T) {
...
+			auctioneer, _ := btcec.PrivKeyFromBytes(pkBytes)
+
+			values[1] = reflect.ValueOf(trader)
+			values[0] = reflect.ValueOf(auctioneer)
+		},
+	}
+	require.NoError(t, quick.Check(scenario, quickCfg))
+}
+
+// TestTaprootSpend tests that the taproot key and script spends can be executed
```
Roasbeef, line 270: 👍

Roasbeef (approved):
LGTM 🐞

dstadulis:
Ready for deployment modulo when the subasta PR is merged.  Do not deploy until then

dstadulis:
⚠️ PR will need to go back to sprint as changes have been made to MuSig2 spec regarding x-only keys

Deliverable outcome:
- [ ] Modify the implemented MuSig2 spec to incoporate recent changes which will promote ISP

guggero:
> warning PR will need to go back to sprint as changes have been made to MuSig2 spec regarding x-only keys
> 
> Deliverable outcome:
> 
>     * [ ]  Modify the implemented MuSig2 spec to incoporate recent changes which will promote ISP

Are we sure we want to wait for that? Updating to the latest spec means:
 - Implement new API in `btcd`, merge.
 - Implement changes to MuSig2 API in `lnd`, merge, release
 - Use new API in Pool.

I think it would make more sense to instead add a version parameter to the MuSig2 API when we implement this, so we can upgrade later.

Roasbeef:
> PR will need to go back to sprint as changes have been made to MuSig2 spec regarding x-only keys

No, it can proceed as defined. As mentioned, those changes themselves are still in a draft state, with no other implementations currently tracking the change. Correctness is still retained _without_ the changes. It would matter more in a p2p setting as you need cross implementation compatibility.

Roasbeef:
@guggero agreed, iiuc we already have the versioning in the proper places as well. An account version is the most apt, since the serialization changes will change the actual aggregate key, so making the jump a future version would be an actual account version bump.

PR #377, opened 2022-07-08
feature: add support for unannounced channels

positiveblue:
Askers are able to decide (and price properly) in which market they want to provide liquidity:
- No preference (default)
- Only announced channels
- Only unannounced channels

Bidders can define what kind of channels they are interested in: announced or unannounced (exclusive or)


auctioneerrpc/auctioneer.proto
```diff
@@ -965,6 +965,12 @@ message ServerBid {
     correspond to the recipient node's details.
     */
     bool is_sidecar_channel = 7;
+
+    /*
+    Flag used to decided if the channels resulting from this order's matches
+    should be announced to the network or not.
+    */
+    bool private_channel = 8;
```
Roasbeef, line 973: Perhaps we should rename this to "unadvertised"?

Roasbeef: Or even move it ot the `ServerOrder` field instead. 

I think one other question here is how we want to implement matching:
  * Should we let only the bid specify if the channel is private or not? 
  * Or should we also give the ask side the freedom to express if they want to accept an unadvertised channel or not?

clientdb/order.go
```diff
@@ -34,6 +34,10 @@ const (
 	// notAllowedNodeIDsType is the tlv type we use to store the list of
 	// node ids the order is not allowed to match with.
 	notAllowedNodeIDsType tlv.Type = 5
+
+	// privateChannelType is the tlv type we use to store the flag for
+	// private channel order constrains.
+	privateChannelType tlv.Type = 6
```
Roasbeef, line 40: I think one other option here is to: expose this as either a channel type, or channel type modifier. In the former case, it'd be an entirely new time. In the latter case, it would be part of another (?) bitvector that specified modifications on the base channel type.

cmd/pool/order.go
```diff
@@ -100,6 +100,11 @@ var baseBidFlags = []cli.Flag{
 			"from our account into the channel; can be " +
 			"used to create up to 50/50 balanced channels",
 	},
+	cli.BoolFlag{
+		Name: "private_channel",
```
Roasbeef, line 104: Carry over comment here: re the name and if asks should be able to specify if they want to accept such a channel or not.

order/batch.go
```diff
@@ -68,6 +72,12 @@ func (bv BatchVersion) SupportsAccountExtension() bool {
 	return bv >= ExtendAccountBatchVersion
 }
 
+// SupportsPrivateChannels is a helper function to easily check if a version
+// supports or not account extension after participating in a batch.
+func (bv BatchVersion) SupportsPrivateChannels() bool {
+	return bv >= PrivateChannelsBatchVersion
```
Roasbeef, line 78: Interesting, I thought we'd instead expose this at the client/server handshake level.

Roasbeef: Given that this information doesn't need to be sent/modified aside the batch prepare/accept message since it's already contained in the orders sent.

funding/manager.go
```diff
@@ -705,6 +705,7 @@ func (m *Manager) BatchChannelSetup(
 					matchedOrderBid.SelfChanBalance,
 				),
 				CommitmentType: commitmentType,
+				Private:        matchedOrderBid.PrivateChannel,
```
Roasbeef, line 708: I think even if we don't spill things over into the ask side of things, we also want to add a channel acceptor check here to ensure the channel is properly formed. This becomes more important when someone wants to buy a: unadvertised zero conf channel.

Roasbeef (commented):
Nice work! 

Once we slot in the zero conf channels, the combo of: zero conf, this, and sidecar channels will allow from pretty much the best on boarding experience imo.

auctioneerrpc/auctioneer.proto
dstadulis: >should we also give the ask side the freedom to express if they want to accept an unadvertised channel 
Should this read: "accept **an order for** an unadvertized..."

Seems the domains for decisions here are:
{bid,ask} {exclusively unannounced channel, ambivalent}

cmd/pool/order.go
```diff
@@ -100,6 +100,11 @@ var baseBidFlags = []cli.Flag{
...
+		Usage: "signal if the new channel should be announced to " +
+			"the network or not.",
```
guggero, line 106: micro-nit: it looks like we don't use punctuation for command line flag descriptions.

order/batch.go
```diff
@@ -51,6 +51,10 @@ const (
 	// ExtendAccountBatchVersion is the first version where accounts expiry
 	// are extended after participating in a batch.
 	ExtendAccountBatchVersion BatchVersion = 1
+
+	// ExtendAccountBatchVersion is the first version where accounts expiry
+	// are extended after participating in a batch.
+	PrivateChannelsBatchVersion BatchVersion = 2
```
guggero, line 57: We should also bump the latest batch version.

order/batch.go
guggero: Let me re-cap the different versions here (we should probably spin this out into a doc or something):

## order version

Is used to signal to the auctioneer what features/fields on the order it should expect. This is mostly used to distinguish between a default value of 0 on an order field between "the client didn't set a value" and "the client didn't know the field, so the default value was used".
So strictly speaking we _should_ add a new order version for the new `private` flag, even though the default value of `false` will be ignored no matter if the client is old or didn't set the flag.

## batch version

This purely signals what runtime features are available to the connected trader. This guards things like "I understand the `NewExpiry` field on a batch, so my account can be auto extended" or "my lnd is >= 0.15 so I can auto-upgrade my account to Taproot in a batch", or, in this case, "I understand the private channel flag and will use a channel acceptor to verify the incoming channel".

So strictly speaking this batch version bump should only be added in conjunction with the channel acceptor being added to check the incoming channel.

## how to implement ask side preference?

Now the big question here is: How do we implement the preference of the maker whether they want to support unannounced channels (no_preference=I don't care, exclusively_announced=only want to provide public channels, exclusively_unannounced=only want to provide unannounced channels).

My suggestion would be to add a new enum on the ask order side that has these three values (with the default value of 0 being no_preference) in this PR, that would be covered by the same order version as the unannounced flag. That way all existing ask orders would become no_preference as soon as the maker updates their software (new batch version that signals that they now understand the unannounced flag and actually set it when opening the channel). And all new ask orders can express explicit preference.

cc @dstadulis, @Roasbeef

guggero (commented):
Did a first pass. Main thing missing IMO is the channel acceptor that makes sure a new channel is actually opened with the correct flag.

clientdb/order.go
```diff
@@ -669,11 +677,13 @@ func DeserializeOrder(nonce order.Nonce, r io.Reader) (
...
-		sidecarTicket     []byte
-		channelType       uint8
-		allowedNodeIDs    []byte
-		notAllowedNodeIDs []byte
+		selfChanBalance            uint64
+		sidecarTicket              []byte
+		channelType                uint8
+		allowedNodeIDs             []byte
+		notAllowedNodeIDs          []byte
+		bidUnannoucedChannels      uint8
```
positiveblue, line 685: we cannot store `bool`

funding/manager.go
positiveblue: Channel acceptor added

auctioneerrpc/auctioneer.proto
positiveblue: discussed offline with @dstadulis, new field for asks added.

cmd/pool/order.go
positiveblue: Added, see comment above :+1:

positiveblue:
Ready for a second pass @Roasbeef @guggero  Main changes:

- Asker has a new field to specify which market s/he is interested selling the liquidity in
- New channel acceptor to ensure that the new channel matches the announcement constraints.
- Bump latest BatchVersion to signal that the client supports unannounced channels
-  `MatchAnnouncementConstraints` only takes into account if the orders are compatible or not, the server is responsible of filtering clients who do not support unannounced channels  yet.

Roasbeef:
> My suggestion would be to add a new enum on the ask order side that has these three values (with the default value of 0 being no_preference) in this PR, that would be covered by the same order version as the unannounced flag. That way all existing ask orders would become no_preference as soon as the maker updates their software (new batch version that signals that they now understand the unannounced flag and actually set it when opening the channel).

This approach SGTM!

auctioneerrpc/auctioneer.proto
```diff
@@ -920,6 +920,13 @@ enum NodeTier {
     TIER_1 = 2;
 }
 
+// Channel announcement constraints for matched channels.
+enum ChannelAnnouncementConstraints {
+    ANNOUNCEMENT_NO_PREFERENCE = 0;
```
Roasbeef, line 925: +1 for this default approach

clientdb/order.go
Roasbeef: Yeah makes sense, oh all those wasted bits...

channel_acceptor.go
```diff
@@ -174,5 +174,27 @@ func (s *ChannelAcceptor) acceptChannel(_ context.Context,
 		}, nil
 	}
 
+	fundingFlags := lnwire.FundingFlag(req.ChannelFlags)
+	isPrivateChan := fundingFlags&lnwire.FFAnnounceChannel == 0
+
+	// Check that the new channel is announced/unannounced as expected.
+	if isPrivateChan != expectedChanBid.UnannouncedChannels {
```
Roasbeef, line 181: 👍

Roasbeef (approved):
The latest diff is looking really good, nice work! 

I think we just need to cap this off with proper itests on the server side.

clientdb/order.go
```diff
@@ -719,6 +743,12 @@ func deserializeOrderTlvData(r io.Reader, o order.Order) error {
 				return err
 			}
 		}
+
+		t, ok := parsedTypes[bidUnannoucedChannelsType]
+		if ok && t == nil {
```
guggero, line 748: This check only makes sure we were able to parse the value correctly. But we still need to add `&& bidUnannoucedChannels == 1`.

clientdb/order.go
```diff
@@ -747,10 +777,19 @@ func deserializeOrderTlvData(r io.Reader, o order.Order) error {
...
+		tlvRecords                 []tlv.Record
+		askAnnouncementConstraints uint8
+		bidUnannoucedChannels      bool
+	)
 
 	switch castOrder := o.(type) {
 	case *order.Ask:
+		noPreference := order.AnnouncementNoPreference
+		if castOrder.AnnouncementConstraints != noPreference {
+			askAnnouncementConstraints =
```
guggero, line 790: nit: format as:
```
	askAnnouncementConstraints = uint8(
		castOrder.AnnouncementConstraints,
	)
```?

File: clientdb/order.go [hunk 13]
```diff
@@ -747,10 +777,19 @@ func deserializeOrderTlvData(r io.Reader, o order.Order) error {
...
+	var (
+		tlvRecords                 []tlv.Record
+		askAnnouncementConstraints uint8
+		bidUnannoucedChannels      bool
+	)
 
 	switch castOrder := o.(type) {
 	case *order.Ask:
+		noPreference := order.AnnouncementNoPreference
+		if castOrder.AnnouncementConstraints != noPreference {
```
guggero, line 789: Do we really need this check when we have `if askAnnouncementConstraints != 0` further down?

order/interfaces.go
```diff
@@ -256,6 +274,26 @@ var (
 	ZeroNonce Nonce
 )
 
+// MatchAnnouncementConstraints returns true when the asker announcement
+// constraints match the bidder announcement preferences.
+func MatchAnnouncementConstraints(asker ChannelAnnouncementConstraints,
```
guggero, line 279: Nice idea to put it here, makes a lot of sense.
This could even be a method on the `ChannelAnnouncementConstraints` type if we want to be fancy.

order/interfaces.go
```diff
@@ -592,6 +634,10 @@ type Bid struct {
 	// happens out of band). This will only be used if the order version is
 	// VersionSidecarChannel or greater.
 	SidecarTicket *sidecar.Ticket
+
+	// UnannouncedChannels signals if the resulting channel needs to be
+	// announced or not.
+	UnannouncedChannels bool
```
guggero, line 640: nit: shouldn't this be singular? Though we can be matched multiple times for the same bid so it could be plural. But then the comment should batch the name.

order/batch.go
```diff
@@ -51,6 +51,10 @@ const (
 	// ExtendAccountBatchVersion is the first version where accounts expiry
 	// are extended after participating in a batch.
 	ExtendAccountBatchVersion BatchVersion = 1
+
+	// UnannouncedChannelsBatchVersion is the first version where orders
+	// can set the flags to only match with announced/unannonced channels.
```
guggero, line 56: Should we add a comment here that the batch version is used to signal readiness for setting the `private` flag when opening the channel on the asker side as well as being able to check the `private` flag in the channel acceptor on the bidder side? Otherwise we might ask ourselves "why wasn't this an order version" later down the line?

guggero (commented):
Very clean and straightforward PR, great work :tada: 

Almost ready to merge, found just one small bug.

auctioneerrpc/auctioneer.proto
```diff
@@ -965,6 +972,11 @@ message ServerBid {
     correspond to the recipient node's details.
     */
     bool is_sidecar_channel = 7;
+
+    /*
+    Signals if this bid is interested in announced or unannounced channels.
+    */
+    bool unannounced_channel = 8;
```
guggero, line 979: nit: unify singular vs. plural here with the CLI flags?

clientdb/order.go
positiveblue: technically the value should be there ONLY IF it was true `bidUnannoucedChannels == 1` I added a comment but I guess that I can make it even clear checking for it anyway. :+1:

clientdb/order.go
positiveblue: TLVs types need to be added in increasing order. Is there any "more clear way" to do it?

clientdb/order.go
guggero: Ah, I see, the whole record wouldn't be there so the `ok` wouldn't be true. Fair enough, but even clearer now I think.

clientdb/order.go
guggero: I mean we could just set `askAnnouncementConstraints = uint8(castOrder.AnnouncementConstraints))` here without the `if castOrder.AnnouncementConstraints != noPreference {` as we check for `if askAnnouncementConstraints != 0 {` again further down.

guggero (approved):
Very nice, LGTM :tada: 

I think this now needs to be stacked on top of #375 because the server side PR for it was just merged. Otherwise you won't be able to pull in this version without getting errors.

guggero (approved):

positiveblue:
closed in favor of #385

PR #379, opened 2022-07-12
GitHub: add automatic Docker build

guggero:
@Roasbeef could you please add the `DOCKER_USERNAME` and `DOCKER_API_KEY` environment variables to this repo?


Roasbeef:
^ the values should now be set 👍

.github/workflows/main.yml
```diff
@@ -21,7 +21,7 @@ env:
   # If you change this value, please change it in the following files as well:
   # /.travis.yml
   # /Dockerfile
-  GO_VERSION: 1.17.2
+  GO_VERSION: 1.17.12
```
bhandras, line 24: not: why not 1.18?

bhandras (approved):

.github/workflows/main.yml
guggero: There are some unresolved problems with the linter and 1.18. We need to solve them in lnd first and then do the same in all other projects.

positiveblue (approved):

PR #380, opened 2022-07-15
funding: don't abort if cleanup fails

guggero:
If we can't clean up pending channels from a previous batch attempt,
that shouldn't block startup. So instead of shutting down, we just log
the warning and continue.

Not sure how the DB got into that state, but this error is what's preventing LiT from starting in this case: https://github.com/lightninglabs/lightning-terminal/issues/372

#### Pull Request Checklist
  used.

### DIFF
--- a/funding/batch.go
+++ b/funding/batch.go
@@ -2,10 +2,10 @@ package funding
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/btcsuite/btcd/wire"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/pool/order"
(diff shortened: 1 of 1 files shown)


levmi:
Thanks for this change. I'm hoping it helps with the repeated issues we've had with litd startup in Umbrel recently.

funding/batch.go
```diff
@@ -2,10 +2,10 @@ package funding
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/btcsuite/btcd/wire"
+	"github.com/davecgh/go-spew/spew"
```
dstadulis, line 8: >a deep pretty printer for Go data structures to aid in debugging

Roasbeef (approved):
LGTM 🧜‍♀️

PR #384, opened 2022-08-03
tools: update Golang to 1.18, add fuzz tests

guggero:
Wanted to add fuzz test coverage in another PR and noticed we'd need to bump the tool chain (and with it the linter) to Go 1.18 first.


sidecar_acceptor_test.go
```diff
@@ -227,7 +213,10 @@ func TestRegisterSidecar(t *testing.T) {
 			}
 
 			require.NoError(t, err)
-			tc.check(t, ticket)
+
+			if tc.check != nil {
```
positiveblue, line 217: :+1:

sidecar_acceptor_test.go
```diff
@@ -202,7 +186,9 @@ func TestRegisterSidecar(t *testing.T) {
 			wallet := test.NewMockWalletKitClient(mockCtrl)
 			store := sidecar.NewMockStore(mockCtrl)
 
-			tc.mockSetter(tc.ticket, signer, wallet, store)
+			if tc.mockSetter != nil {
```
positiveblue, line 189: :+1:

poolscript/script_test.go
```diff
@@ -89,3 +89,25 @@ func TestIncrementBatchKey(t *testing.T) {
...
+		if num > 1 {
+			witness[1] = b
+		}
+		if num > 2 {
+			witness[2] = c
+		}
+		if num > 3 {
+			witness[3] = d
+		}
+		_ = IsMultiSigSpend(witness)
```
positiveblue, line 110: q: is this only looking for a panic?

.golangci.yml
```diff
@@ -9,23 +9,105 @@ linters-settings:
...
+        json: snake
+  whitespace:
+    multi-func: true
+    multi-if: true
+  gosec:
+    excludes:
+      - G402 # Look for bad TLS connection settings.
+      - G306 # Poor file permissions used when writing to a new file.
+  staticcheck:
+    go: "1.17"
```
positiveblue, line 24: Q: does static check not support go "1.18" yet? Would this fail if we add any generic function because the tool will fail to parse it?

positiveblue (approved):
LGTM

we just need to run `go get github.com/ory/go-acc` inside the tools folder so it gets added to the module + fix the mock gen. 
I was able to run `make gen` without any issues locally in the branch so I am not sure if there is a cache problem here.

poolscript/script_test.go
guggero: Yes. Given it's random input, the return value (true/false) doesn't tell us anything. The goal of fuzzing is to prevent panics in general.

.golangci.yml
```diff
@@ -9,23 +9,105 @@ linters-settings:
...
+    - nestif
+    - wsl
+    - exhaustive
+    - forcetypeassert
+    - nilerr
+    - nilnil
+    - stylecheck
+    - thelper
+    - nosprintfhostport
+    - errchkjson
```
sputn1ck, line 102: Didn't check with other linters, but `errchkjson` and `nosprintfhostport` could be removed and the linter still runs through just well. At some point we might need to go through all the linters and either code-adjust or change the comment above.

sputn1ck (approved):
LGTM!

Just noticed that we skip more linters here than in loop (namely `forbidigo` and `promlinter`).

However at some point we should just review which linters we want to enforce and codeadjust for between our projects.

.golangci.yml
guggero: Yes, I agree, we should probably take the time to set a fixed list of linters we want to have active and then synchronize all the linter setting files across all our projects.

guggero:
I have no idea why the mock generation fails on GitHub. I cannot reproduce it locally, neither by running the command directly nor with the dockerized GitHub actions runner `act`... @positiveblue or @sputn1ck can you reproduce the error `Error: 2022/08/04 12:44:43 Loading input failed: sidecar/interfaces.go:300:9: failed parsing arguments: sidecar/interfaces.go:300:39: unknown package "btcec"`?

sputn1ck:
> I have no idea why the mock generation fails on GitHub. I cannot reproduce it locally, neither by running the command directly nor with the dockerized GitHub actions runner `act`... @positiveblue or @sputn1ck can you reproduce the error `Error: 2022/08/04 12:44:43 Loading input failed: sidecar/interfaces.go:300:9: failed parsing arguments: sidecar/interfaces.go:300:39: unknown package "btcec"`?

Can't reproduce either

positiveblue:
@guggero i added a commit on top of yours fixing the mock gen step. We were sharing the go cache with the host and that was causing some permission problems.

guggero:
@positiveblue thanks for figuring it out :tada:

PR #385, opened 2022-08-11
feature: add support for unannounced and zero conf channels 

positiveblue:
For unannounced channels:

[Old discussion bout unannounced channels](https://github.com/lightninglabs/pool/pull/377)

Askers are able to decide (and price properly) in which market they want to provide liquidity:

    No preference (default)
    Only announced channels
    Only unannounced channels

Bidders can define what kind of channels they are interested in: announced or unannounced (exclusive or)

For *zero conf*:

Askers are able to decide (and price properly) in which market they want to provide liquidity:
- No preference (default)
- Only confirmed channels
- Only zero conf channels

Bidders can define what kind of channels they are interested in: confirmed/zero conf (exclusive or)

In order to match zero conf orders the client needs to be running in a batch version that support this feature *at runtime* (during the match). 

The channel acceptor works a bit different than the one for unannounced channels. More in [LND zero conf channels docs](https://github.com/lightningnetwork/lnd/blob/6c915484ba056870f9ed8b57f043d51f26137507/docs/zero_conf_channels.md#initiator-requirements)
 

#### Pull Request Checklist
  used.


order/interfaces.go
```diff
@@ -264,6 +264,22 @@ const (
 	OnlyUnannounced ChannelAnnouncementConstraints = 2
 )
 
+// ChannelAnnouncementConstraints is a numerical type used to denote if the
+// channels created from a match can be announced or not.
+type ChannelConfirmationConstraints uint8
+
+const (
+	ConfirmationNoPreference ChannelConfirmationConstraints = 0
```
Roasbeef, line 272: Style nit: missing godoc comment. Few areas below as well

order/interfaces.go
```diff
@@ -293,6 +309,26 @@ func MatchAnnouncementConstraints(asker ChannelAnnouncementConstraints,
 	}
 }
 
+// MatchAnnouncementConstraints returns true when the asker announcement
```
Roasbeef, line 312: Should be `ZeroConfConstraints`, also still mentions announcement.

cmd/pool/order.go
```diff
@@ -100,13 +100,21 @@ var baseBidFlags = []cli.Flag{
 			"from our account into the channel; can be " +
 			"used to create up to 50/50 balanced channels",
 	},
-	cli.Uint64Flag{
+	cli.BoolFlag{
```
Roasbeef, line 103: In commit message: confrimation -> confirmation

channel_acceptor.go
```diff
@@ -196,5 +196,14 @@ func (s *ChannelAcceptor) acceptChannel(_ context.Context,
 		}, nil
 	}
 
-	return &lndclient.AcceptorResponse{Accept: true}, nil
+	if expectedChanBid.ZeroConfChannel {
```
Roasbeef, line 199: This should also look at the channel type bit as well (the stricter version of the protocol).

positiveblue: This one is a bit tricky because I think we do not have it in the same way we had FF for channel announcement. However, LND will fail to open the channel IF we tell him "accept this channel, it's supposed to be a zeroconf one" and later on the opener does not say anything about it being a zero conf one. That's achieved by setting `ZeroConf: true` here and lndcilent sets it in the  `lnrpc.ChannelAcceptResponse`

guggero (commented):
Code looks pretty good, though I was unable to test with current `lnd` master (`0.15.0-beta commit=v0.15.0-beta-226-gd0d4782c3`).

channel_acceptor.go
Roasbeef: I'm a bit confused. What I mean here is that we want to verify the remote party actually intends to open a zero conf channel as specified in the bid order. Otherwise it seems possbile that we'll permit that any trader can open a zero conf channel here, and we'll accept it, no matter what the bid says. 

The acceptor request has a new field that can be checked: https://github.com/lightningnetwork/lnd/blob/d0d4782c3efce192b6b85160b1bd78ca7b83d528/lnrpc/lightning.proto#L930-L932.

auctioneer/client.go
```diff
@@ -464,14 +464,18 @@ func (c *Client) SubmitOrder(ctx context.Context, o order.Order,
 	// Split into server message which is type specific.
 	switch castOrder := o.(type) {
 	case *order.Ask:
-		constraints := auctioneerrpc.ChannelAnnouncementConstraints(
+		announcement := auctioneerrpc.ChannelAnnouncementConstraints(
 			castOrder.AnnouncementConstraints,
 		)
+		confirmations := auctioneerrpc.ChannelConfirmationConstraints(
```
guggero, line 470: I think we should add an extra check in `SubmitOrder` that the channel type and confirmation constraints are compatible. For example I cannot use an explicit legacy channel with explicit zero conf.

rpcserver.go
```diff
@@ -95,6 +95,10 @@ func (s *accountStore) PendingBatch() error {
 func newRPCServer(server *Server) *rpcServer {
 	accountStore := &accountStore{server.db}
 	lndServices := &server.lndServices.LndServices
+	batchVersion, err := server.determineBatchVersion()
+	if err != nil {
+		rpcLog.Warnf("unable to determine the batch version: %v", err)
```
guggero, line 100: This should be a hard error, not just logging it IMO.

server.go
```diff
@@ -838,29 +845,46 @@ func (s *Server) syncLocalOrderState() error {
...
 	if configVersion >= 0 {
-		return order.BatchVersion(configVersion)
+		return order.BatchVersion(configVersion), nil
 	}
 
-	isTaprootCompatible := false
-	verErr := lndclient.AssertVersionCompatible(
-		s.lndServices.Version, taprootVersion,
+	ctxt, cancel := context.WithTimeout(
+		context.Background(), time.Second*10,
```
guggero, line 858: Use `getInfoTimeout`?

server.go
```diff
@@ -838,29 +845,46 @@ func (s *Server) syncLocalOrderState() error {
...
 	)
-	if verErr == nil {
-		isTaprootCompatible = true
+	defer cancel()
+
+	// We will try to decided what version to used based on the latest
```
guggero, line 862: nit: try to decide what..

server.go
```diff
@@ -838,29 +845,46 @@ func (s *Server) syncLocalOrderState() error {
...
+	// batch version + what features the node supports.
+	info, err := s.lndClient.GetInfo(ctxt, &lnrpc.GetInfoRequest{})
+	if err != nil {
+		return 0, err
+	}
+
+	// If the node supports ZeroConfChannels use that batch version.
+	_, supportsZeroConf := info.Features[uint32(lnwire.ZeroConfOptional)]
```
guggero, line 870: Should we plan ahead and check if either the `lnwire.ZeroConfReqired` or `lnwire.ZeroConfOptional` is turned on? And same for zero conf?

auctioneer/client.go
guggero: And in addition, I think it would make sense to also make sure that `determineBatchVersion() >= ZeroConf` if either the `bid.ZeroConf` or the `ask.ConfirmationConstraints = ZeroConfOnly`.

guggero (changes requested):
Almost there, just two more small requests. Was able to test end to end this time around, nice work :tada:

rpcserver.go
positiveblue: fixed

auctioneer/client.go
positiveblue: Good call, checks for unannounced channels added too :+1:

server.go
positiveblue: Solved... i think

go.mod
```diff
@@ -4,9 +4,9 @@ go 1.18
 
 require (
 	github.com/btcsuite/btcd v0.23.1
-	github.com/btcsuite/btcd/btcec/v2 v2.2.0
-	github.com/btcsuite/btcd/btcutil v1.1.1
-	github.com/btcsuite/btcd/btcutil/psbt v1.1.4
+	github.com/btcsuite/btcd/btcec/v2 v2.2.1
```
guggero, line 7: This commit should disappear when we rebase the unannounced and zero conf channel PRs on top of the Taproot/MuSig2 one.

rpcserver_test.go
```diff
@@ -70,7 +70,7 @@ var renewAccountTestCases = []struct {
 		if expiryHeight == 0 {
 			expiryHeight = 100 + req.GetRelativeExpiry()
 		}
-		version := account.VersionInitialNoVersion
+		version := account.VersionTaprootEnabled
```
guggero, line 73: Should we add a test for both?
Also (pre-existing), incomplete comment below.

cmd/pool/order.go
```diff
@@ -396,6 +401,14 @@ var ordersSubmitAskCommand = cli.Command{
 				"unannounced ones. The default value is \"no " +
 				"preference\"",
 		},
+		cli.Uint64Flag{
+			Name: "confirmation_constraints",
+			Usage: "specifies if the liquidity must be sold in " +
+				"zeroconf or confirmed channels. Set to 1 " +
+				"for only confirmed channels and 2 for only " +
+				"zeroconf ones. The default value is \"no " +
```
guggero, line 409: Could also set the default value here so it would show up correctly. And we need to remove the "The default value is" sentence.

server.go
```diff
@@ -838,29 +850,52 @@ func (s *Server) syncLocalOrderState() error {
...
-	if isTaprootCompatible {
-		return order.UpgradeAccountTaprootBatchVersion
+	// If the node supports ZeroConfChannels use that batch version.
+	_, zeroConfOpt := info.Features[uint32(lnwire.ZeroConfOptional)]
+	_, zeroConfReq := info.Features[uint32(lnwire.ZeroConfRequired)]
+	supportsZeroConf := zeroConfOpt || zeroConfReq
+
+	_, SCIDAliasOpt := info.Features[uint32(lnwire.ScidAliasOptional)]
+	_, SCIDAliasReq := info.Features[uint32(lnwire.ScidAliasRequired)]
+	supportsSCIDAlias := SCIDAliasOpt || SCIDAliasReq
```
guggero, line 881: :+1:

rpcserver.go
```diff
@@ -1306,15 +1306,76 @@ func assertAccountReady(acct *account.Account) error {
...
+			if !usesAnchors {
+				return fmt.Errorf("zero conf channels need " +
+					"to use anchors. Use the script " +
+					"enforced channel type")
+			}
+		}
+
+	case *order.Bid:
+		if castOrder.UnannouncedChannel &&
+			!batchVersion.SupportsUnannouncedChannels() {
```
guggero, line 1350: I don't think this specific case can happen, since support for unannounced channels is purely a question of using the latest Pool version (which this check also is included in).

guggero: Same with the unannounced check in the ask order.

guggero (approved):
LGTM :tada:

go.mod
positiveblue: Do you mean after we bump to `v0.15.1-beta` in `Taproot/MuSig2 `?

cmd/pool/order.go
positiveblue: modified

go.mod
guggero: Yes

dstadulis:
Need ZC to run in subasta
Need tests to pass

channel_acceptor.go
```diff
@@ -196,5 +196,23 @@ func (s *ChannelAcceptor) acceptChannel(_ context.Context,
 		}, nil
 	}
 
-	return &lndclient.AcceptorResponse{Accept: true}, nil
+	// Check that the channel is a zero conf channel if we were expecitng
```
guggero, line 199: nit: s/expecitng/expecting

server.go
```diff
@@ -838,29 +850,52 @@ func (s *Server) syncLocalOrderState() error {
...
+
+	if supportsZeroConf && supportsSCIDAlias {
+		return order.ZeroConfChannelsBatchVersion, nil
+	}
+
+	// If the node is taproot compatible use the
```
guggero, line 887: We added a commit to set the min required version of `lnd` to be `v0.15.1-beta`. So we can return `order.UpgradeAccountTaprootBatchVersion` here directly and remove the `taprootVersion` variable at the top of the file.

guggero (approved):
Nice work, LGTM :tada: 

Just a few typos and style nits, nothing major. Going to take a look at the server PRs now, might need to circle back if there's something that needs to be adjusted in the client PR.

guggero:
Oh, I think we need to consider which of the two fields (on the bid side, probably both) we need to add to the sidecar ticket? Assuming this might be something that could be offered through the sidecar flow.

dstadulis:
There are 3 new commits

Action Items:
@guggero needs to check these
when review is finished, tag roas for review

sidecar/tlv.go
```diff
@@ -174,21 +177,36 @@ func DeserializeTicket(r io.Reader) (*Ticket, error) {
...
+
 	capacity := uint64(o.Capacity)
 	pushAmt := uint64(o.PushAmt)
 
-	var autoAsInt uint8
 	if o.Auto {
 		autoAsInt = 1
 	}
 
+	if o.UnannouncedChannel {
```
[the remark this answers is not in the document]
guggero: EDIT: I tested this and TLV is forward compatible (meaning it doesn't complain if it encounters types it doesn't know). But to keep the ticket as small as possible I still think we should only include the record if the value is true.

guggero (approved):
Sidecar stuff looks good, just two small comments.
Was able to test sidecars with both an old Pool client (and none of the two flags set) and a private channel through a ticket with a new Pool client :tada:

poolrpc/trader.proto
```diff
@@ -1372,6 +1372,14 @@ message DecodedSidecarTicket {
 
     // The original, base58 encoded ticket.
     string encoded_ticket = 16;
+
+    // If true, the channel acceptor for this ticket will expect an unannounced
+    // channel.
+    bool offer_unannounced_channel = 17;
```
Roasbeef, line 1378: 🙌

sidecar/interfaces.go
```diff
@@ -252,6 +274,16 @@ func (t *Ticket) OfferDigest() ([32]byte, error) {
 			return result, err
 		}
 
+	case VersionUnannouncedZeroConf:
+		err := lnwire.WriteElements(
```
Roasbeef, line 278: In the future, we can switch to a TLV format here to make changes like this a bit easier, but totally not blocking atm.

Roasbeef (approved):
LGTM 🎊

I really dig the sidecar additions as well! I wager unannounced sidecar channels will soon become the new default/norm.

PR #390, opened 2022-09-22
feature: add outbound-liquidity auction market, bidder pays for channel acceptance

positiveblue:
Allow the submission of orders for different auction markets. Expand the current auctions, bidder pays for inbound liquidity, with another market where the bidder pays for channel acceptor, outbound liquidity.

To be able to participate in the new auctions:
The asker needs to:

- Set auction type to `BTCOutboundLiquidity`
- Set order amt to 100K (`BaseSupply`)
- Set `min_chan_size` to 100k or more. This will take into account the bidders funding supply.
 
The bidder needs to:

- Set auction type to `BTCOutboundLiquidity`
- Set order amt to 100K (`BaseSupply`)
- Set `self_chan_balance` to 100k or more.

Sidecar tickets are not supported yet for this new market.

NOTE: Only the last 5 commits need to be reviewed in this PR


dstadulis:
initial notes about the impending design doc on this functionality  https://docs.google.com/document/d/1qM2mCtukMFM0lPHh98nihWCTcQ-qtp8bOIdvw9Trxns/edit#heading=h.4tmfx1hae58w

order/interfaces.go
```diff
@@ -282,6 +282,37 @@ const (
 	OnlyZeroConf ChannelConfirmationConstraints = 2
 )
 
+// AuctionType is a numercal type used to denote in what auction market should
```
guggero, line 285: nit: s/numercal/numerical

order/interfaces.go
```diff
@@ -282,6 +282,37 @@ const (
...
+// this order be considered in.
+type AuctionType uint32
+
+const (
+
+	// BTCInboundLiquidity is an auction type where the bidder pays the
+	// asker a premium to get btc inbound liquidity from him.
+	//
+	// TODO(positiveblue): get a better name for this?
```
guggero, line 294: I think the name is pretty clear and not excessively long, fine with me :+1:

order/interfaces.go
```diff
@@ -844,9 +844,24 @@ func (b *Bid) ValidateSelfChanBalance() error {
...
-		return fmt.Errorf("to use self chan balance the min units " +
-			"match must be equal to the order amount in units")
+	if b.AuctionType == BTCOutboundLiquidity {
+		if b.Units != 1 {
+			return fmt.Errorf("to participate in the outbound " +
+				"liquidity market the order amt should be " +
+				"exactly 100k sats")
+		}
+
+		if b.Units != b.MinUnitsMatch {
```
guggero, line 854: We also need to check this for the inbound liquidity market.

order/batch_verifier.go
```diff
@@ -283,6 +283,14 @@ func (v *batchVerifier) validateMatchedOrder(tally *AccountTally,
 			"orders", ourOrder.Nonce())
 	}
 
+	// Auction types must match.
+	auctionType := ourOrder.Details().AuctionType
+	if auctionType != otherOrder.Order.Details().AuctionType {
+		fmt.Printf("jordi: %v %v", auctionType, otherOrder.Order.Details().AuctionType)
```
guggero, line 289: nit: leftover debug print.

guggero (commented):
Did a quick pass, looks almost ready. Nice go get this with such a small diff.

dstadulis:
after we get ZC and unannounced channels merged, we'll need 2x review on this with Roas to cap it out

cmd/pool/order.go
```diff
@@ -696,8 +696,9 @@ func parseBaseBid(ctx *cli.Context) (*poolrpc.Bid, *sidecar.Ticket, error) {
 	if ctx.IsSet("self_chan_balance") {
 		bid.SelfChanBalance = ctx.Uint64("self_chan_balance")
 		bidAmt := btcutil.Amount(bid.Details.Amt)
-		err := sidecar.CheckOfferParams(
-			bidAmt, btcutil.Amount(bid.SelfChanBalance),
+		err := order.CheckOfferParams(
```
guggero, line 699: Do we also need to make sure here that if an order is created for a sidecar (manually) that the auction type can only be outbound liquidity? Otherwise we create an invalid order (since the sidecar itself doesn't encode the auction type).

cmd/pool/order.go
```diff
@@ -129,6 +132,13 @@ var sharedFlags = []cli.Flag{
 			channelTypePeerDependent,
 			channelTypeScriptEnforced),
 	},
+	cli.StringFlag{
+		Name: "auction_type",
+		Usage: fmt.Sprintf("the auction market where this offer "+
+			"must be considered in during the matching (%q, %q)",
+			auctionTypeInboundLiquidity,
+			auctionTypeOutboundLiquidity),
```
guggero, line 140: So we are also changing the meaning of the `min_chan_amount` on the asker side, right? With the outbound liquidity type the `min_chan_amount` now means "the minimum self channel balance a bidder puts in"? Maybe we should add that to the help somehow? Or update the docs in  `doc/orders.md`?
It took me a while to figure this out from the code, so users might not know this either.

guggero (commented):
Got a few more comments, but we're very close!

cmd/pool/order.go
```diff
@@ -129,6 +132,13 @@ var sharedFlags = []cli.Flag{
 			channelTypePeerDependent,
 			channelTypeScriptEnforced),
 	},
+	cli.StringFlag{
+		Name: "auction_type",
+		Usage: fmt.Sprintf("the auction market where this offer "+
```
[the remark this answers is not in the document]
positiveblue: makes sense

cmd/pool/order.go
positiveblue: Added to the documentation. Also, the cli now returns an error explaining what needs to be set.

docs/orders.md
```diff
@@ -297,6 +297,17 @@ To cancel an order. First get the order_nonce of the to be canceled order by usi
...
 ```
 
+## Auction types
+
+Currently there are two kind of auction types supported in pool:
+
+* **BTC inbound liquidity market**: the bidder pays the asker for getting a channel where most of the funding liquidity comes from the asker. The bidder pays a premium based on the asker liquidity supply.
+
+* **BTC outbound liquidity market**: the bidder pays the asker for opening a channel where most of the funding liquidity comes from the bidder side. The bidder pays a premium based on the bidder liquidity supply. 
+In this kind of auctions the asker needs to set the order amount to 100k and specify the `min_chan_amt` that will take into account the bidder funds. The bidder needs to set the order amount to 100k and a `self_chan_balance` greater or equal than 100k.
```
guggero, line 307: This explains the different restrictions but not really what is matched against what.
How about "that will be matched against the bidder's `self_chan_balance`" instead of "that will take into account the bidder funds."?
Also, I think it's "greater or equal to 100k".

guggero (approved):
LGTM, great work :tada:

rpcserver.go
```diff
@@ -2327,11 +2342,18 @@ func (s *rpcServer) prepareLeasesResponse(ctx context.Context,
 					clearingPrice = batch.ClearingPrices[duration]
 				}
 
+				premiumAmt := chanAmt
+				auctionType := ourOrder.Details().AuctionType
+				if auctionType == order.BTCOutboundLiquidity {
+					premiumAmt += btcutil.Amount(
```
Roasbeef, line 2348: Hmm, so the _buyer_ pays the _seller_ a premium based on the amount of BTC the _buyer_ commits? 

I'd say this is probably the biggest open question I have (and I think my main source of confusion with this concept). One key question I have is: as an asker, why is someone _paying_ me to accept a channel?

rpcserver.go
```diff
@@ -2327,11 +2342,18 @@ func (s *rpcServer) prepareLeasesResponse(ctx context.Context,
...
+						selfChanBalance,
```
Roasbeef, line 2349: In most cases though, we expect the channel amount to be nothing, and the actual value being the self channel balance right?

order/batch_verifier_test.go
```diff
@@ -469,8 +503,8 @@ func TestBatchVerifier(t *testing.T) {
 					// bid1ExecFee - bid2ExecFee - chainFees
 					// - bid2SelfChanBalance
 					// 500_000 - 1000 - 1000 -
-					// 1_110 - 1_110 - 186 - 50
-					Value:    495_544,
+					// 1_110 - 1_110 - 186 - 100_000
+					Value:    395_594,
```
Roasbeef, line 507: Did these values need to be changed for the push amount calculation to work correctly? Unclear why we need to increase this and the one above when they aren't using the new feature.

order/tradingfees.go
```diff
@@ -219,23 +219,23 @@ func (t *AccountTally) CalcMakerDelta(feeSchedule terms.FeeSchedule,
...
 // balance delta, taker fees paid, and execution fee paid.
 func takerDelta(feeSchedule terms.FeeSchedule,
-	price FixedRatePremium, totalSats, selfChanBalance btcutil.Amount,
+	price FixedRatePremium, premiumAmt, takerAmt btcutil.Amount,
 	duration uint32) (btcutil.Amount, btcutil.Amount, btcutil.Amount) {
 
 	// Calculate the premium based on the duration the capital will be
 	// locked for.
-	satsPremium := price.LumpSumPremium(totalSats, duration)
+	satsPremium := price.LumpSumPremium(premiumAmt, duration)
```
Roasbeef, line 227: I think the distinction of channel amt vs premium amount here (and elsewhere in the PR is a bit confusing). For example, in this line, we're taking the premium and computing another premium...

Roasbeef (changes requested):
Did an initial pass, but tbh I'm still working to get an accurate mental model in my head of: the way the new feee accounting works, and scenarios in which a buyer/seller would want to enter into this new engagement. 

Would be it be accurate to say that (assuming a fixed number of connection slots), this lets a party _sell_/auction off an _inbound_ (from their PoV) connection/channel slot?

clientdb/order_test.go
```diff
@@ -192,6 +193,7 @@ func dummyOrder(amt btcutil.Amount, leaseDuration uint32) *order.Kit {
 		panic(fmt.Sprintf("could not create private key: %v", err))
 	}
 	kit := order.NewKitWithPreimage(testPreimage)
+	kit.AuctionType = order.BTCInboundLiquidity
```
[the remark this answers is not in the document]
positiveblue: This is making the default one explicit. I am creating a new kit, so it will have the zero value,  which is `order.BTCInboundLiquidity`. Then for testing that serialization works I create an ask with the default `dummyKit` and then overwrite it with something that is not the zero value, only for that test.

order/interfaces.go
```diff
@@ -850,6 +850,14 @@ func (b *Bid) ValidateSelfChanBalance() error {
 			"match must be equal to the order amount in units")
 	}
 
+	if b.AuctionType == BTCOutboundLiquidity {
+		if b.SelfChanBalance == 0 {
```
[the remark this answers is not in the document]
positiveblue: It's on a higher layer, but I can change this to `< SupplyUnit` :ok_hand:

rpcserver.go
positiveblue: Exactly, for the `BTCOutboundLiquidity`the chan amount needs to be always 100k (1 unit). This way we can roll it out and check the demand for this feature without having to make big refactor

rpcserver.go
positiveblue: Yes, the idea is that I have a node well positioned in the network ( my outbound liquidity is valued by someone else) and they want to open a channel to me so they can route through me.

What I am selling is my outbound liquidity, the amount that you will be able to route through me. So that's what I price (your self balance amt).

We can already see many nodes in the network that do not accept channels by default so nobody messes up with their liquidity balance flows, here node operators will be able to price that, so buyer/seller have a new market where both can benefit without having to use the defensive strategy of "If you want to open a channel to us send us an email"

By now, as an asker,  I can set the min chan amount but eventually I should also be able to set a max amt so I can limit how much of my outbound liquidity I want to put in the market.

dstadulis: >as an asker, why is someone paying me to accept a channel?
you have a scarce good, confirmed channel coins and user/LSP wants to zero conf route through the seller

order/tradingfees.go
dstadulis: +100, naming will be comprehensively defined in a new commit

positiveblue: From the "outside" functions it makes sense to call it "premiumAmt" as "the amount that we take into account to calculate the premium of this match" however I agree it is confusing in the last step, when we calculate the premium that we have to pay for this match. I changed it for `baseAmt` in the last step, not sure what better name could fit here :disappointed:

rpcserver.go
```diff
@@ -1392,6 +1400,13 @@ func (s *rpcServer) validateOrder(o order.Order, acct *account.Account,
 			"the same time")
 	}
 
+	if o.Details().AuctionType == order.BTCOutboundLiquidity &&
+		o.Details().Units != 1 {
+
+		return fmt.Errorf("to participate in the outbound liquidity " +
+			"market the order amt should be exactly 100k sats")
```
ffranr, line 1407: This error message talks about amount whereas the if conditional examines units. I expected something like:

```go
if o.Details().AuctionType == order.BTCOutboundLiquidity &&
	o.Details().Amt != order.BaseSupplyUnit {
        return fmt.Errorf("to participate in the outbound liquidity " +
			"market the order amt should be exactly the base supply amount (%v sats)", order.BaseSupplyUnit)
}
```

cmd/pool/order.go
```diff
@@ -688,6 +751,20 @@ func parseBaseBid(ctx *cli.Context) (*poolrpc.Bid, *sidecar.Ticket, error) {
 		}
 	}
 
+	// Checks for the outbound liquidity market.
+	if bid.Details.AuctionType == auctioneerrpc.AuctionType_AUCTION_TYPE_BTC_OUTBOUND_LIQUIDITY {
+		baseSupply := uint64(order.BaseSupplyUnit)
+		if bid.Details.Amt != baseSupply ||
+			bid.SelfChanBalance < baseSupply {
```
ffranr, line 758: I think this should be `bid.SelfChanBalance > bid.Details.Amt`. Which I got from https://github.com/lightninglabs/pool/blob/46df668589d40a53c800c549800bc5820d73cd8d/poolrpc/trader.pb.go#L2413

order/batch_verifier.go
```diff
@@ -203,7 +203,9 @@ func (v *batchVerifier) Verify(batch *Batch, bestHeight uint32) error {
 				),
 			}
 
-		case unitsFilled < ourOrder.Details().MinUnitsMatch:
+		case ourOrder.Details().AuctionType != BTCOutboundLiquidity &&
+			unitsFilled < ourOrder.Details().MinUnitsMatch:
```
ffranr, line 207: I think these lines are correct. I suppose that if the bid is for oubound liquidity only a single unit will be filled. I wound consider adding a comment.

Further, I wander whether a third case is necessary:
```
		case ourOrder.Details().AuctionType == BTCOutboundLiquidity &&
			unitsFilled == 1
```

order/interfaces.go
```diff
@@ -792,14 +823,77 @@ func (b *Bid) ReservedValue(feeSchedule terms.FeeSchedule,
...
+				"multiple of %d", baseSupplyUnit)
+		}
+	}
+
+	return nil
+}
+
+// CheckOfferParamsForOrder makes sure that the order parameters in a
+// sidecar offer are formally valid, sane and match the order parameters.
+func CheckOfferParamsForOrder(auctionType AuctionType, offer sidecar.Offer,
```
ffranr, line 868: The doc for this function refers to sidecar offers specifically. But this PR moves this function from the sidecar package to the order package. Is the doc still correct?

Ditto for `CheckOfferParams`.

order/interfaces.go
```diff
@@ -792,14 +823,77 @@ func (b *Bid) ReservedValue(feeSchedule terms.FeeSchedule,
...
+			feeSchedule, clearingPrice, premiumAmt,
+			b.SelfChanBalance, b.LeaseDuration,
 		)
 		return delta
 	}, accountVersion)
 }
 
+// CheckOfferParams makes sure the offer parameters of a sidecar ticket are
+// valid and sane.
+func CheckOfferParams(auctionType AuctionType, capacity, pushAmt,
```
ffranr, line 841: The doc for this function refers to sidecar offers specifically. But this PR moves this function from the sidecar package to the order package. Is the doc still correct?

order/rpc_parse.go
```diff
@@ -89,7 +90,9 @@ func ParseRPCOrder(version, leaseDuration uint32,
 		return nil, errors.New("min units match must be greater than 0")
 
 	// The min units match must not exceed the total order units.
-	case details.MinUnitsMatch > uint32(kit.Units):
+	case kit.AuctionType != BTCOutboundLiquidity &&
+		details.MinUnitsMatch > uint32(kit.Units):
```
ffranr, line 94: I think the previous condition was valid even if ```kit.AuctionType == BTCOutboundLiquidity```. But I might be misunderstanding the notion of units as they pertain to the outbound liquidity auction type.

ffranr (changes requested):
I've left a few comments around the parts which I'm uncertain about.

cmd/pool/order.go
```diff
@@ -25,9 +25,34 @@ const (
...
+	auctionTypeOutboundLiquidity = "outbound"
+
 	defaultMaxBatchFeeRateSatPerVByte = 100
 
 	defaultConfirmationConstraints = 1
+
+	defaultAuctionType = auctionTypeInboundLiquidity
+
+	submitOrderDescription = `
+        Submit a new order in the given market. Currently, Pool supports two
```
Roasbeef, line 38: great description 👍

Roasbeef (approved):
LGTM 🐞

ffranr (approved):

PR #391, opened 2022-09-23
multi: make profile IP configurable

guggero:
Replaces #228.

Makes the IP address of the profile listener configurable and uses `localhost` if only a port is specified.

#### Pull Request Checklist
  used.


positiveblue (approved):
:+1:

PR #392, opened 2022-09-29
feature: new `--public` flag for orders.

positiveblue:
Add a way for users to signal that it's ok to share this order details in public.

NOTE: there is a gap in the proto fields to do not collision with #390


cmd/pool/order.go
```diff
@@ -540,6 +547,8 @@ func printOrderDetails(client poolrpc.TraderClient, amt btcutil.Amount,
 		fmt.Printf("Self channel balance: %v\n", selfChanBalance)
 	}
 
+	fmt.Printf("Is public: %t\n", isPublic)
```
guggero, line 550: nit: stray debug log statement.

guggero (approved):
Nice and clean, just a single nit. LGTM :tada: 

I assume we explicitly don't want this to be included in sidecar orders as this will mostly be used for ask orders and sidecar orders are usually more private anyway?

cmd/pool/order.go
positiveblue: That was not supposed to be a debug log but to inform the user in the prompt when substituting an order.

In this way we can change the default value in the future and we make sure that users know what they are doing BEFORE submitting an order.

guggero: Aaah, sorry, I didn't see that we're in the CLI. Never mind, then.

dstadulis:
@jbrill can you verify that you don't need anything else?

levmi:
To clarify, this PR covers the functionality of flagging an order on the Pool side as `public`. To consume those public orders and display them publicly on something like Terminal, we'd need a follow up, likely coming from the Terminal side in terms of resourcing, to grab that information. Just want to make sure we're aligned on the Terminal side so we can plan out work.

PR #393, opened 2022-10-06
taproot: ensure check for expiry script spend correctly handles annex

ffranr:
This PR extends [`poolscript.IsTaprootExpirySpend`](https://github.com/lightninglabs/pool/blob/409d7abf1ee25b4878d50e52c78ecfa9fab58151/poolscript/script.go#L455) such that Taproot witness annex elements are correctly handled (ignored).

### DIFF
--- a/poolscript/script.go
+++ b/poolscript/script.go
@@ -453,8 +453,15 @@ func IsExpirySpend(witness wire.TxWitness) bool {
 // IsTaprootExpirySpend determines whether the provided witness corresponds to
 // the expiration script path of a Taproot enabled (version 1) account.
 func IsTaprootExpirySpend(witness wire.TxWitness) bool {
-	// We still have 3 elements, the trader's signature, the script and the
-	// control block.
+	// At a minimum we expect these three elements: the trader's signature,
+	// the script, and the control block. We also accept an additional
+	// optional fourth element: the annex.
+	if hasAnnex(witness) {
+		// Remove the annex element. If the annex is present, it is
+		// always the last element.
+		witness = witness[:len(witness)-1]
+	}
+
 	if len(witness) != 3 {
 		return false
 	}
@@ -496,6 +503,26 @@ func IsTaprootExpirySpend(witness wire.TxWitness) bool {
 		return false
 	}
 	return true
+}
+
+// hasAnnex returns true if the provided witness includes an annex element,
+// otherwise returns false.
+func hasAnnex(witness wire.TxWitness) bool {
+	// By definition, the annex element can not be the sole element in the
+	// witness stack.
+	if len(witness) < 2 {
+		return false
+	}
+
+	// If an annex element is included in the witness stack, by definition,
+	// it will be the last element and will be prefixed by a Taproot annex
+	// tag.
+	lastElement := witness[len(witness)-1]
+	if len(lastElement) == 0 {
+		return false
+	}
+
+	return lastElement[0] == txscript.TaprootAnnexTag
 }
 
 // IsMultiSigSpend determines whether the provided witness corresponds to the
--- a/poolscript/script_test.go
+++ b/poolscript/script_test.go
@@ -132,6 +132,33 @@ func FuzzWitnessSpendDetection(f *testing.F) {
 	})
 }
 
+// TestHasAnnex tests whether hasAnnex can correctly identify the presence of an
+// annex element within a witness stack.
+func TestHasAnnex(t *testing.T) {
+	t.Parallel()
+
+	// The annex element is the last element and includes the correct
+	// prefix.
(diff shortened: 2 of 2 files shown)


poolscript/script_test.go
```diff
@@ -132,6 +132,32 @@ func FuzzWitnessSpendDetection(f *testing.F) {
...
+	t.Parallel()
+
+	// The annex element is the last element and includes the correct prefix.
+	witness := make([][]byte, 2)
+	witness[0] = []byte{97, 98}
+	witness[1] = []byte{txscript.TaprootAnnexTag, 98}
+	result := hasAnnex(witness)
+	require.True(t, result)
+
+	// The annex element may not be the only element.
```
positiveblue, line 147: :+1:

poolscript/script_test.go
```diff
@@ -132,6 +132,32 @@ func FuzzWitnessSpendDetection(f *testing.F) {
 	})
 }
 
+// TestHasAnnex tests whether hasAnnex can correctly identify
```
positiveblue, line 135: nit: we can wrap the lines at 80 char

```
// TestHasAnnex tests whether hasAnnex can correctly identify the presence of
// an annex element within a witness stack.
```

poolscript/script_test.go
```diff
@@ -132,6 +132,32 @@ func FuzzWitnessSpendDetection(f *testing.F) {
...
+// the presence of an annex element within a witness stack.
+func TestHasAnnex(t *testing.T) {
+	t.Parallel()
+
+	// The annex element is the last element and includes the correct prefix.
```
positiveblue, line 140: nit: 81 char line

poolscript/script.go
```diff
@@ -498,6 +500,22 @@ func IsTaprootExpirySpend(witness wire.TxWitness) bool {
 	return true
 }
 
+// hasAnnex returns true if the provided witness includes an annex element,
+// otherwise returns false.
+func hasAnnex(witness wire.TxWitness) bool {
+	// By definition, the annex element can not be the sole element in the witness stack.
+	if len(witness) < 2 {
+		return false
+	}
```
positiveblue, line 509: nit: we usually separate the blocks with empty lines, this function would be then three brlocks (two breaks)

positiveblue (approved):
Nice work @ffranr :rocket: 

I added a couple of format related nits.


PS: for my future me, here is [some info](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#rationale)

poolscript/script_test.go
ffranr: I'm don't think I understand what you mean. [Line 140](https://github.com/ffranr/pool/blob/8ff3a0e3ecbd14bdd534ff206cee845ccf6b6831/poolscript/script_test.go#L140) is 74 chars by my count.

ffranr: Ah I see what's up, you're counting tabs as 8 characters.

poolscript/script.go
```diff
@@ -453,9 +453,11 @@ func IsExpirySpend(witness wire.TxWitness) bool {
...
 // the expiration script path of a Taproot enabled (version 1) account.
 func IsTaprootExpirySpend(witness wire.TxWitness) bool {
-	// We still have 3 elements, the trader's signature, the script and the
-	// control block.
-	if len(witness) != 3 {
+	// At a minimum we expect these three elements: the trader's signature, the script, and the
+	// control block. We also accept an additional optional fourth element: the annex.
+	if hasAnnex(witness) && len(witness) != 4 {
+		return false
+	} else if len(witness) != 3 {
```
guggero, line 460: nit: in Golang we usually try to avoid using the `else` statement if possible. That way the "happy path" is always to the left while error or special cases are indented.
It's all about readability, so I would probably write this block as follows:
```
if hasAnnex(witness) {
        // Snip off the last element that contains the annex.
        witness = witness[:len(witness)]
}

if len(witness) != 3 {
      return false
}
```

But of course that also comes down to personal taste.

guggero (approved):
Nice work, LGTM :tada: 

Looks good, pending the remaining few nits.
The formatting nits are a bit annoying at first, I'm sure. But setting up the editor to use 8 spaces for tabs and adding a visual line at 80 characters should help.

poolscript/script.go
ffranr: That's good to know, thanks. I've gone with your example.

ffranr:
Thank you @positiveblue and @guggero for the reviews.

I'm happy to hear about any formatting issues you guys might find, no problem at all. I've added an 80 character visual line to my IDE and I've set tabs to 8 characters.

If you notice any further formatting nits please do let me know.

PR #395, opened 2022-10-13
Expose historical account action fees via RPC

ffranr:
#### Pull Request Checklist
  used.


ffranr:
@guggero does my solution look as if it's going in the right direction?

(By the way, I haven't completed the logic yet and the formatting isn't correct.)

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
 	return account.VersionInitialNoVersion, nil
 }
 
+type ActionFee struct {
```
guggero, line 2942: I'd call this something like `ModificationFee`, since Action and Auction look very similar :sweat_smile:

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+	Action  string
+	BlockHeight int32
+	TimeStamp int64
+	OutputAmount int64
+	TotalFee int64
+}
+
+// AccountActionFees returns a map from account key to an ordered list of
+// account action fees.
+func (s *rpcServer) AccountActionFees() (map[string][]ActionFee, error) {
```
guggero, line 2952: :+1:

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+func txOutputAmount(tx *lnrpc.Transaction) (*int64, error) {
+	for _, outputDetail := range tx.OutputDetails {
+		if !outputDetail.IsOurAddress {
+			return &outputDetail.Amount, nil
+		}
+	}
+	return nil, fmt.Errorf("failed to find output amount, tx hash: %v", tx.BlockHash)
+}
+
+var reAccountCreationLabel = regexp.MustCompile(
```
guggero, line 3045: All of this remaining code should probably go into the `account` package.

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+	if len(match) > 0 {
+		accountKey := match[1]
+		return &accountCreationAction{accountKey}, nil
+	}
+	return nil, nil
+}
+
+var reAccountModificationLabel = regexp.MustCompile(
+	`poold -- ` +
+	`AccountModification\(` +
```
guggero, line 3064: Nice, I see you're a man of culture as well (using RegExp) :sunglasses:

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+		}
+		return transactions[i].TimeStamp < transactions[j].TimeStamp
+	})
+
+	// A map from account key to an ordered list of account action fees.
+	accountActions := make(map[string][]ActionFee)
+
+	for _, tx := range transactions {
+		// Handle account creation action.
+		accountCreation, err := parseAccountCreationAction(tx)
```
guggero, line 2974: Could we instead create a function `parseTxLabel()` that finds out what kind of action it was (using the match method of regexp directly) and then returns the correct struct? Might need to put things behind an interface so you can return different kinds of structs.

guggero (commented):
Approach looks good! Allowed myself to add some early comments.

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+		if err != nil {
```
positiveblue, line 2975: can this function return an error?

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
 	return account.VersionInitialNoVersion, nil
 }
 
+type ActionFee struct {
+	Action  string
+	BlockHeight int32
+	TimeStamp int64
```
positiveblue, line 2945: nit: we usually do `Timestamp` and only use uppercase when a new word is starting

rpcserver.go
```diff
@@ -2936,6 +2939,174 @@ func (s *rpcServer) determineAccountVersion(
...
+		accountKey := match[1]
+		return &accountCreationAction{accountKey}, nil
+	}
+	return nil, nil
+}
+
+var reAccountModificationLabel = regexp.MustCompile(
+	`poold -- ` +
+	`AccountModification\(` +
+	`acct_key=(?P<acct_key>[0-9a-z]*), ` +
```
positiveblue, line 3065: q: I think this has never changed, but maybe just in case we add/remove more in the future we can do something like:


```
func (s *rpcServer) AccountActionFees() (map[string][]ActionFee, error) {
    ...
    for _, tx := range transactions {
        switch {
        case isAccountCreation(tx):
            processAccountCreation(tx)
        case is AccountModification(tx)
            processAccountModification(tx)
         ... 
         }
         ...
    }
}

func isAccountModification(tx) bool {
    // only match poold --  AccountModification\(`
}

func ProcessAccountModification(tx) {
    // match acc_key
    // match expiry
    // match deposit
    // match is_close
    // separately, so if we change it in the future we are able to parse old tx without the new fields
}

```

positiveblue (commented):
looks really nice @ffranr :pray:

rpcserver.go
ffranr: The logic which serializes account action data into labels is separate from the logic which deserializes that data. The label can included action data which the parser couldn't parse. That's where the error might arise.

rpcserver.go
ffranr: Will do. I think I was inspired by https://github.com/lightningnetwork/lnd/blob/2e410dd248a7de1beee4e305d510d5b667bb24f5/lnrpc/lightning.pb.go#L1708

rpcserver.go
```diff
@@ -2936,6 +2937,123 @@ func (s *rpcServer) determineAccountVersion(
...
+	}
+
+	// Sort transactions in temporal order.
+	transactions := transactionDetails.Transactions
+	sort.Slice(transactions, func(i, j int) bool {
+		if transactions[i].BlockHeight != transactions[j].BlockHeight {
+			return transactions[i].BlockHeight < transactions[j].BlockHeight
+		}
+		return transactions[i].TimeStamp < transactions[j].TimeStamp
+	})
```
ffranr, line 2959: I think there's a bug here.

ffranr:
```json
{
    "transactions": [
        {
            "tx_hash": "269cc54eee1462379ba0e2c30e47a5c7f51f7d55e785a56335eff18425d69bac",
            "amount": "0",
            "num_confirmations": 0,
            "block_hash": "",
            "block_height": 0,
            "time_stamp": "1666109043",
            "total_fees": "0",
            "dest_addresses": [
                "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8",
                    "pk_script": "51205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d70518353059",
                    "output_index": "0",
                    "amount": "148675",
                    "is_our_address": false
                }
            ],
            "raw_tx_hex": "02000000000101a74ac2b16ace0f001e3b1077693c038595b2df8edac61f00d0041127b4b9ae6301000000000000000001c3440200000000002251205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d705183530590140f37009c3b6272f1099b5b00d0ff86b9bc2fccc9065a2a5a8b83415de99430e58df940d372723773b956ca0b782661ec3dc4db2d2607e7d68a5133e047d60907200000000",
            "label": " poold -- AccountModification(acct_key=0271e9f2bbca55d8a9c7dabd2265de96e063cfb5f6c2bf793025c6157da759c152, expiry=false, deposit=false, is_close=true)",
            "previous_outpoints": [
                {
                    "outpoint": "63aeb9b4271104d0001fc6da8edfb29585033c6977103b1e000fce6ab1c24aa7:1",
                    "is_our_output": false
                }
            ]
        },
        {
            "tx_hash": "63aeb9b4271104d0001fc6da8edfb29585033c6977103b1e000fce6ab1c24aa7",
            "amount": "0",
            "num_confirmations": 0,
            "block_hash": "",
            "block_height": 0,
            "time_stamp": "1666108688",
            "total_fees": "0",
            "dest_addresses": [
                "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8",
                "bcrt1pu7uv5cx644c3mm0zxwvec9733pdd52vjs46a7pdvvrrjktuwju4ssxxhe5"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8",
                    "pk_script": "51205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d70518353059",
                    "output_index": "0",
                    "amount": "50000",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pu7uv5cx644c3mm0zxwvec9733pdd52vjs46a7pdvvrrjktuwju4ssxxhe5",
                    "pk_script": "5120e7b8ca60daad711dede233999c17d1885ada29928575df05ac60c72b2f8e972b",
                    "output_index": "1",
                    "amount": "149230",
                    "is_our_address": false
                }
            ],
            "raw_tx_hex": "02000000000101e09508e7796a66e033d2986a1826f9d2b66ad48a0e28637c1931cbfde5960ac20000000000000000000250c30000000000002251205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d70518353059ee46020000000000225120e7b8ca60daad711dede233999c17d1885ada29928575df05ac60c72b2f8e972b014088ff74cf0ae6a7ecd217bd5f46a6aa5ba460ddcf8797a8bb4f1d72535f351407631f37c9e9b2a67921b0d2a7250f49fc93d9e50c9aeea6a7be634c685539461000000000",
            "label": " poold -- AccountModification(acct_key=0271e9f2bbca55d8a9c7dabd2265de96e063cfb5f6c2bf793025c6157da759c152, expiry=false, deposit=false, is_close=false)",
            "previous_outpoints": [
                {
                    "outpoint": "c20a96e5fdcb31197c63280e8ad46ab6d2f926186a98d233e0666a79e70895e0:0",
                    "is_our_output": false
                }
            ]
        },
        {
            "tx_hash": "c20a96e5fdcb31197c63280e8ad46ab6d2f926186a98d233e0666a79e70895e0",
            "amount": "-101105",
            "num_confirmations": 30,
            "block_hash": "26cf2c2284abefeca0672b0b597a4c8fff3f9f3df33ebb3a9b4fd898f87248f9",
            "block_height": 126,
            "time_stamp": "1666108453",
            "total_fees": "0",
            "dest_addresses": [
                "bcrt1pkz3ennua2fvd890l64kfk3kzv4nuuha8khekk3f3k2f8jh7gyp7s6ch3qa",
                "bcrt1qugnlx7a9palnq3srdy2w459mran9a8klhu3nvz"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pkz3ennua2fvd890l64kfk3kzv4nuuha8khekk3f3k2f8jh7gyp7s6ch3qa",
                    "pk_script": "5120b0a399cf9d5258d395ffd56c9b46c26567ce5fa7b5f36b4531b292795fc8207d",
                    "output_index": "0",
                    "amount": "200000",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_PUBKEY_HASH",
                    "address": "bcrt1qugnlx7a9palnq3srdy2w459mran9a8klhu3nvz",
                    "pk_script": "0014e227f37ba50f7f3046036914ead0bb1f665e9edf",
                    "output_index": "1",
                    "amount": "469781793",
                    "is_our_address": true
                }
            ],
            "raw_tx_hex": "02000000000102458d4214650e3c82a105b4c8898a4137b52ed66201b72ba7de0e16db67f0daf5000000000000000000458d4214650e3c82a105b4c8898a4137b52ed66201b72ba7de0e16db67f0daf501000000000000000002400d030000000000225120b0a399cf9d5258d395ffd56c9b46c26567ce5fa7b5f36b4531b292795fc8207d214d001c00000000160014e227f37ba50f7f3046036914ead0bb1f665e9edf0247304402206a6e212eafc282ab077029d951e4e7cd2954759ed90ca9d3deeb25501123660b022023cb7e331a4a2c4c06a2910abaf78cc266f9340dd41c216617604856826d6d8b012103b4cfd696fbaeaa4b29a27e8a5dc76ed5b3d353d46401cfe199c863ae6c0e20bc0140024b2fc1eb1ec1c26027bdcb608c3823df802c2e11edbd0736a463ce6b2263899bc08e1ba4f60352b7d20ac8891c33561667bb0ed526cb647eb66cd31597df7800000000",
            "label": " poold -- AccountModification(acct_key=0271e9f2bbca55d8a9c7dabd2265de96e063cfb5f6c2bf793025c6157da759c152, expiry=false, deposit=true, is_close=false)",
            "previous_outpoints": [
                {
                    "outpoint": "f5daf067db160edea72bb70162d62eb537418a89c8b405a1823c0e6514428d45:0",
                    "is_our_output": true
                },
                {
                    "outpoint": "f5daf067db160edea72bb70162d62eb537418a89c8b405a1823c0e6514428d45:1",
                    "is_our_output": false
                }
            ]
        },
        {
            "tx_hash": "f5daf067db160edea72bb70162d62eb537418a89c8b405a1823c0e6514428d45",
            "amount": "-101165",
            "num_confirmations": 36,
            "block_hash": "259115a7e264c0890740b02e51b267e3f02bf65742669e7221041ab2f8a15314",
            "block_height": 120,
            "time_stamp": "1666108206",
            "total_fees": "1165",
            "dest_addresses": [
                "bcrt1qntplspfhhjgsyzzetqktsy7j4wljdj8ll8w5np",
                "bcrt1pljv9y72p0fnf6nq3az36ecem337nu423axlxk5llvqlw0ehjzeesc0j0gs"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_PUBKEY_HASH",
                    "address": "bcrt1qntplspfhhjgsyzzetqktsy7j4wljdj8ll8w5np",
                    "pk_script": "00149ac3f80537bc91020859582cb813d2abbf26c8ff",
                    "output_index": "0",
                    "amount": "469882898",
                    "is_our_address": true
                },
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pljv9y72p0fnf6nq3az36ecem337nu423axlxk5llvqlw0ehjzeesc0j0gs",
                    "pk_script": "5120fc985279417a669d4c11e8a3ace33b8c7d3e5551e9be6b53ff603ee7e6f21673",
                    "output_index": "1",
                    "amount": "100000",
                    "is_our_address": false
                }
            ],
            "raw_tx_hex": "010000000001010ca15d27c3b6b4ee71ce997e59d75f9e35a7379fe052b36aae2b4050c8e5b2a90100000000ffffffff0212d8011c000000001600149ac3f80537bc91020859582cb813d2abbf26c8ffa086010000000000225120fc985279417a669d4c11e8a3ace33b8c7d3e5551e9be6b53ff603ee7e6f216730140333795a7a6dbee913274dd967273499aa1a81852c902ed209d7f225d44a14c2f7e08ead7543e66903f72a7069d3aa6b1b30ab87cd028d4f1a5063a3e9536fea900000000",
            "label": " poold -- AccountCreation(acct_key=0271e9f2bbca55d8a9c7dabd2265de96e063cfb5f6c2bf793025c6157da759c152)",
            "previous_outpoints": [
                {
                    "outpoint": "a9b2e5c850402bae6ab352e09f37a7359e5fd7597e99ce71eeb4b6c3275da10c:1",
                    "is_our_output": true
                }
            ]
        },
        {
            "tx_hash": "a9b2e5c850402bae6ab352e09f37a7359e5fd7597e99ce71eeb4b6c3275da10c",
            "amount": "-15007700",
            "num_confirmations": 42,
            "block_hash": "7a23789ff0980deed294d5d89acbd014c225246b856d287f8bd350efb7d89950",
            "block_height": 114,
            "time_stamp": "1666106284",
            "total_fees": "7700",
            "dest_addresses": [
                "bcrt1qxpm6yjnj9f0metvw07l7hxpxsvmjdgzdr7ehhruh06wnqz2c5lzq3fa0g0",
                "bcrt1pxeav0gclxychdfusggapd9ntc8wtchzu6rdgvfymr9lk6uw3lheqg6fpre"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_SCRIPT_HASH",
                    "address": "bcrt1qxpm6yjnj9f0metvw07l7hxpxsvmjdgzdr7ehhruh06wnqz2c5lzq3fa0g0",
                    "pk_script": "00203077a24a722a5fbcad8e7fbfeb9826833726a04d1fb37b8f977e9d300958a7c4",
                    "output_index": "0",
                    "amount": "15000000",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pxeav0gclxychdfusggapd9ntc8wtchzu6rdgvfymr9lk6uw3lheqg6fpre",
                    "pk_script": "5120367ac7a31f313176a790423a16966bc1dcbc5c5cd0da86249b197f6d71d1fdf2",
                    "output_index": "1",
                    "amount": "469984063",
                    "is_our_address": true
                }
            ],
            "raw_tx_hex": "0200000000010189f3c5873bf78c6c17b9f90294d485a1927e332fc255588e6a639a9f91bd483d01000000000000000002c0e1e400000000002200203077a24a722a5fbcad8e7fbfeb9826833726a04d1fb37b8f977e9d300958a7c43f63031c00000000225120367ac7a31f313176a790423a16966bc1dcbc5c5cd0da86249b197f6d71d1fdf20140eee7e0ff898b8c60682d27355348537e3a76d15ae5ea03fddcab12eb8d3b10b80ec89e6ef9ad5ad43029d0a57dec5f5690c41ef0868608634d721dffc200c8b100000000",
            "label": "0:openchannel:shortchanid-125344325632000",
            "previous_outpoints": [
                {
                    "outpoint": "3d48bd919f9a636a8e5855c22f337e92a185d49402f9b9176c8cf73b87c5f389:1",
                    "is_our_output": true
                }
            ]
        },
        {
            "tx_hash": "3d48bd919f9a636a8e5855c22f337e92a185d49402f9b9176c8cf73b87c5f389",
            "amount": "-15008237",
            "num_confirmations": 43,
            "block_hash": "012b0f157e8ae9286ef10006d9b09bc928f28f4813d3e11d493547541b8a22fe",
            "block_height": 113,
            "time_stamp": "1666106278",
            "total_fees": "8237",
            "dest_addresses": [
                "bcrt1qcr5dexw246625nyz0k65dv8jv944k9h9zcl444jxtkatj7u945asmllh8s",
                "bcrt1pqcewggylxn26wjk2ngw4h0cd8gp9layr2llw9q9zpc8nxqmlzejqlvg9pq"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_SCRIPT_HASH",
                    "address": "bcrt1qcr5dexw246625nyz0k65dv8jv944k9h9zcl444jxtkatj7u945asmllh8s",
                    "pk_script": "0020c0e8dc99caaeb4aa4c827db546b0f2616b5b16e5163f5ad6465dbab97b85ad3b",
                    "output_index": "0",
                    "amount": "15000000",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pqcewggylxn26wjk2ngw4h0cd8gp9layr2llw9q9zpc8nxqmlzejqlvg9pq",
                    "pk_script": "51200632e4209f34d5a74aca9a1d5bbf0d3a025ff48357fee280a20e0f33037f1664",
                    "output_index": "1",
                    "amount": "484991763",
                    "is_our_address": true
                }
            ],
            "raw_tx_hex": "020000000001019bf7a67a3ebea27436d617b2ba97ac00de5468c21182e65a455c065565c14d9c01000000000000000002c0e1e40000000000220020c0e8dc99caaeb4aa4c827db546b0f2616b5b16e5163f5ad6465dbab97b85ad3b1363e81c000000002251200632e4209f34d5a74aca9a1d5bbf0d3a025ff48357fee280a20e0f33037f166402483045022100b71ded07728b3f0f2465e10adf42445daac08ac9665fcafc2e46580f2b134f010220119a1c6a56ed36069b3413c24eaed61adff18c72b5579b3253faa42d13de318c0121029da410b3311c2e0e8aac936b060d09cad7add8132e335488285fc26021f5b86300000000",
            "label": "0:openchannel:shortchanid-124244814135296",
            "previous_outpoints": [
                {
                    "outpoint": "9c4dc16555065c455ae68211c26854de00ac97bab217d63674a2be3e7aa6f79b:1",
                    "is_our_output": true
                }
            ]
        },
        {
            "tx_hash": "9c4dc16555065c455ae68211c26854de00ac97bab217d63674a2be3e7aa6f79b",
            "amount": "500000000",
            "num_confirmations": 49,
            "block_hash": "252ec710d23e398ba24bdfcd1864a563ebd935577e437fd3a338cd9f7d908bae",
            "block_height": 107,
            "time_stamp": "1666106245",
            "total_fees": "0",
            "dest_addresses": [
                "bcrt1qx8yvrmlal7ksmsc07zgu3dkqdlcagkcch67vkx",
                "bcrt1q6pkadgfe24ll30d2tfsq6522lyauldnsf8932j"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_PUBKEY_HASH",
                    "address": "bcrt1qx8yvrmlal7ksmsc07zgu3dkqdlcagkcch67vkx",
                    "pk_script": "001431c8c1effdffad0dc30ff091c8b6c06ff1d45b18",
                    "output_index": "0",
                    "amount": "4499995620",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_WITNESS_V0_PUBKEY_HASH",
                    "address": "bcrt1q6pkadgfe24ll30d2tfsq6522lyauldnsf8932j",
                    "pk_script": "0014d06dd6a139557ff8bdaa5a600d514af93bcfb670",
                    "output_index": "1",
                    "amount": "500000000",
                    "is_our_address": true
                }
            ],
            "raw_tx_hex": "02000000014e451544b7ab527da85bcd3667cc30050a15b0d490beeea0ee36dd158f52b6c0000000006a4730440220629e8fd63c510c683ab5c15aab7f54c568c10225fef4743beb3f5f2977ce91be022003836f9f0b2aee84b2bed0dfc9a51ab95179434d67d2bb7d072eb935f2575ba9012103588e20c3ddfcc5ab277e63793321eb6d8106d57680f2aaa935f8a8613266c06dfeffffff02e47b380c0100000016001431c8c1effdffad0dc30ff091c8b6c06ff1d45b180065cd1d00000000160014d06dd6a139557ff8bdaa5a600d514af93bcfb6706a000000",
            "label": "",
            "previous_outpoints": [
                {
                    "outpoint": "c0b6528f15dd36eea0eebe90d4b0150a0530cc6736cd5ba87d52abb74415454e:0",
                    "is_our_output": false
                }
            ]
        }
    ]
}
```

rpcserver.go
ffranr: I've gone with your idea in the latest commit.

account/manager.go
```diff
@@ -798,6 +801,150 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
+// account closure action.
+func isAccountClosure(txLabel string) bool {
+	return strings.HasPrefix(txLabel, " poold -- AccountClosure(")
+}
+
+// These regular expressions (regex) match on transaction label flags.
+var reLabelAccountKey = regexp.MustCompile(`acct_key=(?P<acct_key>[0-9a-z]*)`)
+var reLabelExpiry = regexp.MustCompile(`expiry=(?P<expiry>(false|true))`)
+var reLabelDeposit = regexp.MustCompile(`deposit=(?P<deposit>(false|true))`)
+var reLabelIsClose = regexp.MustCompile(`is_close=(?P<is_close>(false|true))\)`)
```
jamaljsr, line 826: There are a few things I noticed while performing this label parsing on the frontend.

Here are the labels I see for each account action:
```
Create:
  poold -- AccountCreation(acct_key=028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252)
Deposit:
  poold -- AccountModification(acct_key=028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252, expiry=false, deposit=true, is_close=false)
Withdrawal:
  poold -- AccountModification(acct_key=028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252, expiry=false, deposit=false, is_close=false)
Renew:
  poold -- AccountModification(acct_key=028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252, expiry=false, deposit=false, is_close=false)
Close:
  poold -- AccountModification(acct_key=028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252, expiry=false, deposit=false, is_close=true)
```

1. The `expiry` value is never `true`. I'm not sure when this is ever set to `true`. I assumed it would be for account renewals but that isn't the case. Maybe a bug?
2. There is no `withdraw` value in the label.
3. Due to the above, it's not possible to distinguish between withdrawals and renewals from the labels alone.

Just figured I'd point out the issue's I'm currently running into. Maybe they can also be addressed in this PR.

ffranr: Yes, those are the issues which are holding up this PR. Thank you for the clear summary!

I think it should be possible to distinguish between withdraw and renew based on the transaction outputs. But I'm not certain. The transaction outputs that I'm currently looking at (when it comes to withdraw) seem to be corrupted. Oli suspected lnd versioning mismatches.

Could I ask you to post a complete withdraw transaction please (you've given the label only in your comment)? I wander if it might look very different from what I can produce.

ffranr: As you've mentioned, the account renew action does not have a flag of its own and doesn't touch the expiry flag pathway. See: https://github.com/lightninglabs/pool/blob/9280be1cfe939a8fb9bec0bf0eef05376a84cdd7/account/manager.go#L1319-L1322

ffranr: https://github.com/lightninglabs/pool/pull/396

ffranr:
I'm still stuck with this as a withdraw transaction:
```json
{
            "tx_hash": "755b00c399d3bf8eec0fe68156af9b90d443436bbb9f282078aeae855a2a0205",
            "amount": "0",
            "num_confirmations": 0,
            "block_hash": "",
            "block_height": 0,
            "time_stamp": "1666710175",
            "total_fees": "0",
            "dest_addresses": [
                "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8",
                "bcrt1pr9tnxkd5txungm4tl4vjcl0mad3mu8cskyw96axnzla4j0qrwevsq5t4sg"
            ],
            "output_details": [
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1p2agvu57xt3p6wut5xql59qpazhx2j02dswlprm2tv0ts2xp4xpvsed8wg8",
                    "pk_script": "51205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d70518353059",
                    "output_index": "0",
                    "amount": "50000",
                    "is_our_address": false
                },
                {
                    "output_type": "SCRIPT_TYPE_PUBKEY_HASH",
                    "address": "bcrt1pr9tnxkd5txungm4tl4vjcl0mad3mu8cskyw96axnzla4j0qrwevsq5t4sg",
                    "pk_script": "512019573359b459b9346eabfd592c7dfbeb63be1f10b11c5d74d317fb593c037659",
                    "output_index": "1",
                    "amount": "142300",
                    "is_our_address": false
                }
            ],
            "raw_tx_hex": "0200000000010158b13fd365bee3ce6673efaae79bafdd59f7dba31df42b1353503c44a1b672ec0000000000000000000250c30000000000002251205750ce53c65c43a77174303f42803d15cca93d4d83be11ed4b63d70518353059dc2b02000000000022512019573359b459b9346eabfd592c7dfbeb63be1f10b11c5d74d317fb593c037659014047a40ba6132b9ab6da159aa7fd7bb814ee9f58751d5e50db07d92f8b3a7929f853c49f8488d0518a8404bd62ec164b7f724e554fa9c030839e1d8a4e7570da0100000000",
            "label": " poold -- AccountModification(acct_key=038709293544c27d37c632f15bc5a2a2d0ab6d0fc46e6fbc28ec899015d26c3f9c, expiry=false, deposit=false, is_close=false)",
            "previous_outpoints": [
                {
                    "outpoint": "ec72b6a1443c5053132bf41da3dbf759ddaf9be7aaef7366cee3be65d33fb158:0",
                    "is_our_output": false
                }
            ]
        },
```

LND version seems to be correct:
```
user@lightninglabs:~/pool$ reg_alice version
{
    "lncli": {
        "commit": "v0.15.3-beta",
        "commit_hash": "b4e7131bdb47531ad2f00ce345ddcdb58935bba5",
        "version": "0.15.3-beta",
        "app_major": 0,
        "app_minor": 15,
        "app_patch": 3,
        "app_pre_release": "beta",
        "build_tags": [
            "autopilotrpc",
            "signrpc",
            "walletrpc",
            "chainrpc",
            "invoicesrpc",
            "watchtowerrpc",
            "neutrinorpc",
            "monitoring",
            "peersrpc",
            "kvdb_postgres",
            "kvdb_etcd"
        ],
        "go_version": "go1.18.2"
    },
    "lnd": {
        "commit": "v0.15.3-beta",
        "commit_hash": "b4e7131bdb47531ad2f00ce345ddcdb58935bba5",
        "version": "0.15.3-beta",
        "app_major": 0,
        "app_minor": 15,
        "app_patch": 3,
        "app_pre_release": "beta",
        "build_tags": [
            "autopilotrpc",
            "signrpc",
            "walletrpc",
            "chainrpc",
            "invoicesrpc",
            "watchtowerrpc",
            "neutrinorpc",
            "monitoring",
            "peersrpc",
            "kvdb_postgres",
            "kvdb_etcd"
        ],
        "go_version": "go1.18.2"
    }
}
```

account/manager.go
```diff
@@ -52,6 +53,8 @@ const (
 	// both extremes for valid account expirations.
 	minAccountExpiry = 144       // One day worth of blocks.
 	maxAccountExpiry = 144 * 365 // A year worth of blocks.
+
+	txLabelPrefixTag = "poold -- %s"
```
ffranr, line 57: Is there somewhere better to put this? It isn't specific to `account` although it's only used in relation to accounts right now.

account/manager.go
```diff
@@ -165,13 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
-	labelString := fmt.Sprintf(prefixTag, labelJson)
+	labelString := fmt.Sprintf(txLabelPrefixTag, labelJson)
 	return labelString
 }
 
+// IsPoolTx returns true if the given transaction is related to
+// pool.
+func IsPoolTx(tx *lnrpc.Transaction) bool {
+	return strings.HasPrefix(tx.Label, txLabelPrefixTag)
+}
```
ffranr, line 180: Somewhere more general for this?

rpcserver.go
```diff
@@ -3027,6 +3027,88 @@ func (s *rpcServer) determineAccountVersion(
...
+		output := account.SelectOutput(tx, labelData.OutputIndex)
+		if output != nil {
+			return nil, fmt.Errorf("failed to identify account "+
+				"tx output: %v", tx)
+		}
+
+		fee := btcutil.Amount(0)
+		if labelData.TxFee != nil {
+			fee = *labelData.TxFee
+		}
```
ffranr, line 3073: This is a bit ugly. Technically the fee could be missing from the tx label json (if lnd couldn't be reached). Is there a better way to do this?

rpcserver.go
```diff
@@ -3027,6 +3027,88 @@ func (s *rpcServer) determineAccountVersion(
...
+		// Handle all non-create account actions.
+		actions := accountActions[labelData.Key]
+		actions = append(actions, &poolrpc.AccountModificationFee{
+			Action:       string(labelData.Action),
+			BlockHeight:  tx.BlockHeight,
+			Timestamp:    tx.TimeStamp,
+			OutputAmount: output.Amount,
+			Fee:          int64(fee),
+		})
+		accountActions[labelData.Key] = actions
```
ffranr, line 3084: This seems clumsy. Is there a better way to append to a map's list?

account/manager.go
guggero: I think that's fine. We could export it so we could re-use the constant in Faraday. But maybe that's overkill and might produce a circular package dependency.
The on-chain footprint of Pool will always be related to accounts, so the location is okay too IMO.

account/manager.go
```diff
@@ -165,13 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
+
+// ParseTxLabel parses and returns data fields stored in a given account
+// transaction label.
+func ParseTxLabel(label string) (*AccountTxLabel, error) {
+	label = strings.TrimPrefix(label, txLabelPrefixTag)
```
guggero, line 185: I don't think this will work since the tag also contains the `%s`. So probably just need a `fmt.Sprintf(txLabelPrefixTag, "")` instead of the constant directly.

account/manager.go
guggero: I think this is fine.

account/manager.go
```diff
@@ -165,13 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
+
+// SelectOutput returns the account specific output detail given a transaction
+// and the account output index.
+func SelectOutput(tx *lnrpc.Transaction, accountOutputIndex uint32) *lnrpc.OutputDetail {
+	for _, output := range tx.OutputDetails {
+		if output.OutputIndex == int64(accountOutputIndex) {
+			return output
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
guggero: Yes, I'd add an error here just to make the `nil` value less implicit.

poolrpc/trader.proto
```diff
@@ -78,6 +78,13 @@ service Trader {
     rpc RecoverAccounts (RecoverAccountsRequest)
         returns (RecoverAccountsResponse);
 
+    /* pool: `accounts accountmodificationfees`
```
guggero, line 81: The command in the back ticks needs to match an actual CLI sub command. It is parsed by the API docs generator and will fail since the command is called `listfees` instead of `accountmodificationfees`.

poolrpc/trader.proto
```diff
@@ -1011,6 +1018,35 @@ message RecoverAccountsResponse {
...
+    int64 fee = 5;
+}
+
+message ListOfAccountModificationFees {
+    repeated AccountModificationFee account_modification_fee = 1;
+}
+
+message AccountModificationFeesResponse {
+    // A map from account key to an ordered list of account modification fees.
+    map<string, ListOfAccountModificationFees> account_modification_fees = 1;
```
guggero, line 1047: nit: call this just `accounts` and the above message `modification_fees`?
Then on the CLI the JSON looks like this:
```
{
  "accounts": {
    "03xxxx": {
      "modification_fees": [
        { ... }
      ]
    }
  }
}
```

rpcserver.go
```diff
@@ -3027,6 +3027,88 @@ func (s *rpcServer) determineAccountVersion(
 	return account.VersionTaprootEnabled, nil
 }
 
+// AccountModificationFees returns a map from account key to an ordered list of
+// account modifying action fees.
+func (s *rpcServer) AccountModificationFees(ctx context.Context,
+	req *poolrpc.AccountModificationFeesRequest) (
+	*poolrpc.AccountModificationFeesResponse, error) {
+	// Retrieve all lightning wallet transactions.
+	transactionDetails, err := s.lndClient.GetTransactions(
```
guggero, line 3036: nit: formatting of multi-line method declaration (above) and method call (here).

rpcserver.go
guggero: You can do it directly, not sure if that is much nicer (though a bit shorter and easier to parse for me at least):
```
		accountActions[labelData.Key] = append(
			accountActions[labelData.Key],
			&poolrpc.AccountModificationFee{
				Action:       string(labelData.Action),
				BlockHeight:  tx.BlockHeight,
				Timestamp:    tx.TimeStamp,
				OutputAmount: output.Amount,
				Fee:          int64(fee),
			},
		)
```

rpcserver.go
```diff
@@ -3027,6 +3027,88 @@ func (s *rpcServer) determineAccountVersion(
...
+	return prepareAccountModificationFeesRpcResponse(accountActions), nil
+}
+
+// prepareAccountModificationFeesRpcResponse converts a map (trader keys -> list
+// of account modification fees) into a RPC read response where map values are
+// structs rather than lists.
+//
+// Protocol Buffers does not support maps with list values. As a workaround,
+// map values are set to a message type which includes a repeated field.
+func prepareAccountModificationFeesRpcResponse(
```
guggero, line 3096: Given this method is a bit tricky to format and is only used once, I'd do this inline above instead.
So just:
```
	result := make(map[string]*poolrpc.ListOfAccountModificationFees)
	for traderKey, modificationFees := range accountActions {
		result[traderKey] = &poolrpc.ListOfAccountModificationFees{
			AccountModificationFee: modificationFees,
		}
	}

	return &poolrpc.AccountModificationFeesResponse{
		AccountModificationFees: result,
	}
```

guggero (commented):
Left a few comments and tried to answer your questions.

account/manager.go
ffranr: Updated commit includes your suggested change and a unit test.

ffranr:
Thanks for the review @guggero and for answering my questions! I believe my latest update has addressed your comments, please let me know if I've missed anything.

rpcserver.go
```diff
@@ -3030,6 +3030,77 @@ func (s *rpcServer) determineAccountVersion(
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving transactions: %v", err)
+	}
+
+	// A map from account key to an ordered list of account action fees.
+	accountActions := make(map[string][]*poolrpc.AccountModificationFee)
+
+	for _, tx := range transactionDetails.Transactions {
+		// Skip transactions which are
```
positiveblue, line 3051: incomplete sentence?

rpcserver.go
```diff
@@ -3030,6 +3030,77 @@ func (s *rpcServer) determineAccountVersion(
...
+		if !account.IsPoolTx(tx) {
+			continue
+		}
+
+		// Parse account data from transaction label.
+		labelData, err := account.ParseTxLabel(tx.Label)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse account "+
+				"transaction label: %v (error: %v)", tx.Label, err)
+		}
```
positiveblue, line 3061: nit: `\n`

account/manager.go
```diff
@@ -165,10 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
+	return &data.Account, nil
+}
+
+// SelectOutput returns the account specific output detail given a transaction
+// and the account output index.
+func SelectOutput(tx *lnrpc.Transaction, accountOutputIndex uint32) (
+	*lnrpc.OutputDetail, error) {
+
+	for _, output := range tx.OutputDetails {
+		if output.OutputIndex == int64(accountOutputIndex) {
```
positiveblue, line 202: q: aren't the outputs in order? where `tx.OutputDetails[i].OuptutIndex == i`

account/manager_test.go
```diff
@@ -1430,3 +1430,28 @@ func TestMakeTxnLabel(t *testing.T) {
 		require.Equal(t, genLabel, testCase.label)
 	}
 }
+
+// TestParseTxLabel tests whether an example account transaction label can be
+// parsed correctly.
+func TestParseTxLabel(t *testing.T) {
+	t.Parallel()
+
+	label := `poold -- {"account":{"action":"withdraw","key":` +
```
positiveblue, line 1439: nit: i think we use double quotes `"` with `+`

positiveblue (approved):
:ok_hand:

account/manager.go
ffranr: I wasn't sure. Should I assume that they are?

account/manager_test.go
ffranr: I think that would mean adding `\"`, `\}`, and `\{` throughout. Which would be less readable. Should I still go with `"` for consistency?

account/manager.go
```diff
@@ -165,10 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
 		log.Errorf("Internal error: failed to serialize json "+
 			"from %v: %v", label, err)
-		return fmt.Sprintf(prefixTag, action)
+		return fmt.Sprintf(txLabelPrefixTag, action)
+	}
+
+	return fmt.Sprintf(txLabelPrefixTag, labelJson)
+}
+
+// IsPoolTx returns true if the given transaction is related to
```
guggero, line 175: nit: can use the full length of the line for the comment.

account/manager.go
```diff
@@ -165,10 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
+// pool.
+func IsPoolTx(tx *lnrpc.Transaction) bool {
+	return strings.HasPrefix(tx.Label, txLabelPrefixTag)
```
guggero, line 178: This doesn't work, because of the `%s`. Maybe we should remove the formatting part and just use that where we actually write a string? Otherwise we'd have to add the `prefix := fmt.Sprintf(txLabelPrefixTag, "")` here too.

account/manager.go
```diff
@@ -165,10 +166,44 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
...
+}
+
+// ParseTxLabel parses and returns data fields stored in a given account
+// transaction label.
+func ParseTxLabel(label string) (*AccountTxLabel, error) {
+	prefix := fmt.Sprintf(txLabelPrefixTag, "")
+	label = strings.TrimPrefix(label, prefix)
+
+	var data TxLabel
```
guggero, line 187: The data in the label currently isn't written as `TxLabel` but as `AccountTxLabel` instead. So this doesn't work (returns an empty account). Did we decide to put the `"account":{}` level into the label or not? I don't remember.

account/manager.go
guggero: Yes, the outputs are in order, should be able to get rid of the whole function.

account/manager_test.go
guggero, line 1439: The label currently isn't written in this format (see comment above). But given the test assumes it is, we should probably fix that in the code where we create the label.

guggero (changes requested):
I tested this out locally and found a few things that didn't work. Would be nice if you could do a quick manual test run before requesting review, please.
Other than that it looks very good, can get this merged soon I think.

account/manager.go
ffranr: I think we did decide on `"account":{}` as far as I can tell from my notes here: https://github.com/lightninglabs/pool/issues/397

account/manager.go
ffranr: I will remove `%s` from the tag, simpler to keep formatting symbols with formatting function calls.

poolrpc/trader.proto
```diff
@@ -1011,6 +1018,35 @@ message RecoverAccountsResponse {
...
+    int32 block_height = 2;
+
+    // Action transaction timestamp.
+    int64 timestamp = 3;
+
+    // Action transaction output amount.
+    int64 output_amount = 4;
+
+    // Action transaction fee.
+    int64 fee = 5;
```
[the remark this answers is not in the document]
ffranr: I've implemented a `oneof` field in my latest commit:
```
    // Action transaction fee.
    oneof fee {
        // A flag which is true if fee value has not been set, and is otherwise
        // false.
        bool fee_null = 5;

        // Action transaction fee value.
        int64 fee_value = 6;
    }
```

where `fee_null` is a flag that should be considered before reading `fee_value`.

Let me know what you think!

ffranr:
Thank you for your latest review @guggero ! You're right I should always try out the code before I request a review. I'll make sure that I do.

I've tried out the latest commit with account creation and it went as expected.

Here's an example deposit label:
```
 "label": "poold -- {\"account\":{\"key\":\"034b2564b04f24e243ce54366b1b23eb9bf98cd97b2ad13e0b7fa5b093c30171e2\",\"action\":\"deposit\",\"expiry_height\":4439,\"output_index\":0,\"expiry_spend\":false,\"tx_fee\":1105,\"balance_diff\":200000}}",
```

And here's an example account creation label:
```
"label": "poold -- {\"account\":{\"key\":\"034b2564b04f24e243ce54366b1b23eb9bf98cd97b2ad13e0b7fa5b093c30171e2\",\"action\":\"create\",\"expiry_height\":4439,\"output_index\":0,\"expiry_spend\":false,\"tx_fee\":null,\"balance_diff\":200000}}",
```

Note `\"tx_fee\":null` for the account creation transaction because of this TODO: https://github.com/lightninglabs/pool/blob/1e600d5e3dee51ed1e7c2f049a35baba89232245/account/manager.go#L639

poolrpc/trader.proto
```diff
@@ -1011,6 +1018,42 @@ message RecoverAccountsResponse {
...
+    string action = 1;
+
+    // Action transaction block height.
+    int32 block_height = 2;
+
+    // Action transaction timestamp.
+    int64 timestamp = 3;
+
+    // Action transaction output amount.
+    int64 output_amount = 4;
```
guggero, line 1035: I think it would be great to also have the TXID (as a string, otherwise we have the byte reverse order problem) and the amount diff as well. Then we would be able to correlate it to a wallet transaction in the UI if we wanted/needed to. Can be a follow-up PR to unblock this.

guggero (approved):
tACK, LGTM :tada:

poolrpc/trader.proto
ffranr: Locally, I've added the `txid`. Will push an update.

Isn't amount diff redundant in that a consumer could just calculate it from the output amounts the RPC already provides? I don't think the RPC should provide a different view (diffs) onto the same data. Unless it's something that would be needed very commonly I suppose?

ffranr:
@guggero Thanks for the review. I've added `txid` as you suggested. I've tested by checking the RPC return after creating a pool account.

You mentioned a "[byte reverse order problem](https://github.com/lightninglabs/pool/pull/395#discussion_r1022576738)", what is that?

guggero:
> You mentioned a "[byte reverse order problem](https://github.com/lightninglabs/pool/pull/395#discussion_r1022576738)", what is that?

By [convention](https://bitcoin.stackexchange.com/questions/2063/why-does-the-bitcoin-protocol-use-the-little-endian-notation) the ID of a transaction (a.k.a. TXID) is the little endian representation of its SHA256 hash. Which means the hex notation seen in most block explorers is actually the reverse of the hex representation of the hash. So in `lnd` we normally use the raw hash when a gRPC field is of type `bytes` (which corresponds to `[]byte`) but use the human-readable (e.g. reverse order) notation when showing it as a hex string (gRPC type of `string`). In some instances we even re-format things in the CLI just to have nice copy/pasteable TXIDs in the console output: https://github.com/lightninglabs/pool/blob/master/cmd/pool/account.go#L47

ffranr:
Thanks for the explanation!

poolrpc/trader.proto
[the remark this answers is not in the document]
jamaljsr: Just seeing this now. The `fee_null` field works for me 👍

PR #396, opened 2022-10-24
account: improve account onchain transaction labels

ffranr:
Keep prefix tag, but drop the leading space: `poold -- `. Faraday looks for this tag.

Label  structure:

```json
{
  "account":{
    "action": "<action>",
    "key": "<string>",
    "expiry_height": <int>,
    "output_index": <int>,
    "expiry_spend": <bool>,
    "tx_fee": <int>,
    "balance_diff": <int>,
  }
}
```

The `action` field takes one of the following values:
* `create`
* `deposit`
* `withdraw`
* `renew`
* `close`

`key` is the account key.

`expiry_height` is an integer block height at which the account expires. This value is important to ease the process of account recovery, see [here](https://github.com/lightninglabs/pool/pull/396#discussion_r1003366116).

`output_index` is an integer denoting the account balance transaction output.

`expiry_spend` is a boolean denoting whether the transaction spends via the expiry output path.

`tx_fee` is the onchain transaction fee for executing the account action.

`balance_diff` is a signed integer denoting the difference (+/-) in the account balance as a result of the transaction.

Here's a full minimized example (211 characters):
```
poold -- {"account":{"action":"withdraw","key":"028f4218ed13a50d9b6cc2afa947bba40e9336eaedbb10129ac1fda73a3d806252","expiry_height":2400,"output_index":1,"expiry_spend":false,"tx_fee":1027,"balance_diff":10000}}
```

#### Pull Request Checklist
  used.

### DIFF
--- a/account/manager.go
+++ b/account/manager.go
@@ -3,6 +3,7 @@ package account
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -105,6 +106,70 @@ func (wt witnessType) witnessSize() (int, error) {
 	default:
 		return 0, fmt.Errorf("unknown witness type %v", wt)
 	}
+}
+
+// IsExpirySpend returns true if the witness is taking an expiration path.
+func (wt witnessType) IsExpirySpend() bool {
+	switch wt {
+	case expiryWitness, expiryTaproot:
+		return true
+	default:
+		return false
+	}
+}
+
+// Action is a type of account modification.
+type Action string
+
+const (
+	CREATE   Action = "create"
+	DEPOSIT  Action = "deposit"
+	WITHDRAW Action = "withdraw"
+	RENEW    Action = "renew"
+	CLOSE    Action = "close"
+)
+
+type TxLabel struct {
+	Account AccountTxLabel `json:"account"`
+}
+
+type AccountTxLabel struct {
+	Key           string          `json:"key"`
+	Action        Action          `json:"action"`
+	ExpiryHeight  uint32          `json:"expiry_height"`
+	OutputIndex   uint32          `json:"output_index"`
+	IsExpirySpend bool            `json:"expiry_spend"`
+	TxFee         *btcutil.Amount `json:"tx_fee"`
+	BalanceDiff   btcutil.Amount  `json:"balance_diff"`
+}
+
+// actionTxLabel returns a transaction label for use with account
+// modification actions.
+func actionTxLabel(account *Account, action Action, isExpirySpend bool,
+	txFee *btcutil.Amount, balanceDiff btcutil.Amount) string {
+
+	prefixTag := "poold -- %s"
+
+	acctKey := account.TraderKey.PubKey.SerializeCompressed()
+	key := fmt.Sprintf("%x", acctKey)
+	label := AccountTxLabel{
+		Key:           key,
+		Action:        action,
+		ExpiryHeight:  account.Expiry,
+		OutputIndex:   account.OutPoint.Index,
+		IsExpirySpend: isExpirySpend,
+		TxFee:         txFee,
+		BalanceDiff:   balanceDiff,
+	}
+	labelJson, err := json.Marshal(label)
+	if err != nil {
+		log.Errorf("Internal error: failed to serialize json "+
+			"from %v", label)
+		return fmt.Sprintf(prefixTag, action)
+	}
+
+	labelString := fmt.Sprintf(prefixTag, labelJson)
+	return labelString
 }
 
 // ManagerConfig contains all of the required dependencies for the Manager to
@@ -548,10 +613,23 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 					"be greater than 0", acctKey)
 			}
 
-			// If we have a label prefix, then we'll apply that now
-			// and also attach some additional meta data.
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
+			// Attach additional meta data to transaction label.
+			//
+			// TODO(ffranr): Tx inputs are required to calculate the
+			// fee. The fee will be added to the tx label.
+			// m.cfg.Wallet.SendOutputs returns and selects tx
+			// inputs but also requires the label as an argument.
+			// Instead, the tx should be constructed via
+			// m.cfg.Wallet.FundPsbt which would give us an
+			// opportunity to inspect tx inputs before creating a
+			// tx label.
+			var txFee *btcutil.Amount = nil
+
+			balanceDiff := account.Value
+			contextLabel := actionTxLabel(
+				account, CREATE, false, txFee,
+				balanceDiff,
+			)
 			label := makeTxnLabel(m.cfg.TxLabelPrefix, contextLabel)
 
 			// TODO(wilmer): Expose manual controls to bump fees.
@@ -624,9 +702,20 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 				}
 			}
 
-			acctKey := account.TraderKey.PubKey.SerializeCompressed()
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
+			txFee, err := m.deriveFeeFromTx(ctx, accountTx)
+			fee := &txFee
+			if err != nil {
+				log.Errorf(
+					"Failed to derive fee from "+
+						"transaction: %v", accountTx,
+				)
+				fee = nil
+			}
+			balanceDiff := account.Value
+			contextLabel := actionTxLabel(
+				account, CREATE, false, fee,
+				balanceDiff,
+			)
 			label := makeTxnLabel(m.cfg.TxLabelPrefix, contextLabel)
 
 			err = m.maybeBroadcastTx(ctx, accountTx, label)
@@ -766,12 +855,22 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 	// transaction to confirm so that we can transition the account to its
 	// final state.
 	case StatePendingClosed:
-		acctKey := account.TraderKey.PubKey.SerializeCompressed()
-		contextLabel := fmt.Sprintf(" poold -- "+
-			"AccountClosure(acct_key=%x)", acctKey)
+		txFee, err := m.deriveFeeFromTx(ctx, account.LatestTx)
+		fee := &txFee
+		if err != nil {
+			log.Errorf(
+				"Failed to derive fee from "+
+					"transaction: %v", account.LatestTx,
+			)
+			fee = nil
+		}
+		balanceDiff := account.Value
+		contextLabel := actionTxLabel(
+			account, CLOSE, false, fee, balanceDiff,
+		)
 		label := makeTxnLabel(m.cfg.TxLabelPrefix, contextLabel)
 
-		err := m.maybeBroadcastTx(ctx, account.LatestTx, label)
+		err = m.maybeBroadcastTx(ctx, account.LatestTx, label)
 		if err != nil {
 			return err
 		}
@@ -796,6 +895,32 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 	}
 
 	return nil
+}
+
+// deriveFeeFromTx derives the transaction fee from a given transaction message.
+func (m *manager) deriveFeeFromTx(ctx context.Context,
+	tx *wire.MsgTx) (btcutil.Amount, error) {
+
+	// Input and output values are required to calculate the tx fee. However,
+	// tx messages do not contain input values. We will therefore locate those
+	// input values via the outputs that they represent.
+	var sumInputs int64
+	for _, txIn := range tx.TxIn {
+		spendTx, err := m.locateTxByHash(ctx, txIn.PreviousOutPoint.Hash)
+		if err != nil {
+			return 0, nil
+		}
+		txOut := spendTx.TxOut[txIn.PreviousOutPoint.Index]
+		sumInputs += txOut.Value
+	}
+
+	var sumOutputs int64
+	for _, txOut := range tx.TxOut {
+		sumOutputs += txOut.Value
+	}
+
+	fee := sumInputs - sumOutputs
+	return btcutil.Amount(fee), nil
 }
 
 // locateTxByOutput locates a transaction from the Manager's TxSource by one of
@@ -1165,8 +1290,7 @@ func (m *manager) DepositAccount(ctx context.Context,
 	// and assuming it's valid, broadcast the deposit transaction.
 	modifiers = append(modifiers, StateModifier(StatePendingUpdate))
 	modifiedAccount, spendTx, err := m.spendAccount(
-		ctx, account, packet, spendWitnessType, modifiers, false,
-		bestHeight,
+		ctx, account, DEPOSIT, packet, spendWitnessType, modifiers, bestHeight,
 	)
 	if err != nil {
 		releaseInputs()
@@ -1241,8 +1365,7 @@ func (m *manager) WithdrawAccount(ctx context.Context,
 	// and assuming it's valid, broadcast the withdrawal transaction.
 	modifiers = append(modifiers, StateModifier(StatePendingUpdate))
 	modifiedAccount, spendTx, err := m.spendAccount(
-		ctx, account, packet, spendWitnessType, modifiers, false,
-		bestHeight,
+		ctx, account, WITHDRAW, packet, spendWitnessType, modifiers, bestHeight,
 	)
 	if err != nil {
 		return nil, nil, err
@@ -1317,8 +1440,7 @@ func (m *manager) RenewAccount(ctx context.Context,
 	// and assuming it's valid, broadcast the update transaction.
 	modifiers = append(modifiers, StateModifier(StatePendingUpdate))
 	modifiedAccount, spendTx, err := m.spendAccount(
-		ctx, account, packet, spendWitnessType, modifiers, false,
-		bestHeight,
+		ctx, account, RENEW, packet, spendWitnessType, modifiers, bestHeight,
 	)
 	if err != nil {
 		return nil, nil, err
@@ -1454,8 +1576,7 @@ func (m *manager) CloseAccount(ctx context.Context, traderKey *btcec.PublicKey,
 		ValueModifier(0), StateModifier(StatePendingClosed),
 	}
 	_, spendTx, err := m.spendAccount(
-		ctx, account, packet, spendWitnessType, modifiers, true,
-		bestHeight,
+		ctx, account, CLOSE, packet, spendWitnessType, modifiers, bestHeight,
 	)
 	if err != nil {
 		return nil, err
@@ -1472,13 +1593,13 @@ func (m *manager) CloseAccount(ctx context.Context, traderKey *btcec.PublicKey,
 // able to resume the spend of an account upon restarts if they happen to
 // shutdown mid-process.
 func (m *manager) spendAccount(ctx context.Context, account *Account,
-	packet *psbt.Packet, witnessType witnessType, modifiers []Modifier,
-	isClose bool, bestHeight uint32) (*Account, *wire.MsgTx, error) {
+	action Action, packet *psbt.Packet, witnessType witnessType,
+	modifiers []Modifier, bestHeight uint32) (*Account, *wire.MsgTx, error) {
 
 	// In case we're not closing the account, let's now locate our
 	// re-created account output.
 	accountOutputIdx := -1
-	if !isClose {
+	if action != CLOSE {
 		// The account output should be recreated, so we need to locate
 		// the new account outpoint.
 		newAccountOutput, err := account.Copy(modifiers...).Output()
@@ -1514,7 +1635,7 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
 	var lockTime uint32
 	switch witnessType {
 	case expiryWitness, expiryTaproot:
-		if !isClose {
+		if action != CLOSE {
 			return nil, nil, errors.New("modifications for " +
 				"expired accounts are not currently supported")
 		}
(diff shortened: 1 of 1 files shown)


account/manager.go
```diff
@@ -1546,19 +1561,12 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
...
 	if err := m.cfg.Store.UpdateAccount(account, modifiers...); err != nil {
 		return nil, nil, err
 	}
 
-	// As this is a generic account modification, we'll add some additional
-	// information to make accounting for this transaction a bit easier.
-	deposit := prevAccountState.Value < account.Value
 	acctKey := account.TraderKey.PubKey.SerializeCompressed()
-	contextLabel := fmt.Sprintf(" poold -- AccountModification(acct_key=%x, "+
-		"expiry=%v, deposit=%v, is_close=%v)", acctKey,
```
positiveblue, line 1559: i think we should leave the `expiry` tag here.

For [legacy](https://github.com/lightninglabs/pool/blob/master/poolscript/script.go#L58) and [taproot](https://github.com/lightninglabs/pool/blob/master/poolscript/script.go#L93) accounts the expiry date is part of the scrip. That means that the user needs that parameter to recreate the spending script.

If some day our server goes down, the client-side only recovery  account process could have a nice speed up by using this data, instead of having to perform a brute force (as we have right now).

ffranr: That's a good point, thank you!

Currently the expiry tag is just a bool. Which is never set to true as far as I can tell. If I understand you correctly, perhaps it should be a date?

account/manager.go
```diff
@@ -107,6 +108,65 @@ func (wt witnessType) witnessSize() (int, error) {
...
+	IsExpirySpend bool           `json:"expiry_spend"`
+	TxFee         btcutil.Amount `json:"tx_fee"`
+	BalanceDiff   btcutil.Amount `json:"balance_diff"`
+}
+
+// actionTxLabel returns a transaction label for use with account
+// modification actions.
+func actionTxLabel(account Account, action Action, isExpirySpend bool,
+	txFee btcutil.Amount, balanceDiff btcutil.Amount,
+) (*string, error) {
```
guggero, line 140: nit: formatting, should break either between arguments or return types to avoid this non-indented line here.

account/manager.go
```diff
@@ -107,6 +108,65 @@ func (wt witnessType) witnessSize() (int, error) {
...
+
+	prefixTag := "poold -- "
+
+	acctKey := account.TraderKey.PubKey.SerializeCompressed()
+	label := AccountTxLabel{
+		Key:           string(acctKey),
```
guggero, line 146: The key should be hex formatted, otherwise this will directly convert the byte slice into ASCII characters.

account/manager.go
```diff
@@ -107,6 +108,65 @@ func (wt witnessType) witnessSize() (int, error) {
...
+		BalanceDiff:   balanceDiff,
+	}
+	labelJson, err := json.Marshal(label)
+	if err != nil {
+		return nil, err
+	}
+
+	// Transaction labels are limited in size. Therefore, compact the JSON.
+	labelCompactJson := new(bytes.Buffer)
+	err = json.Compact(labelCompactJson, labelJson)
```
guggero, line 161: I think by default `json.Marshal()` already produces non-indented JSON, so no need to compact again.

account/manager.go
guggero: Also, we don't normally use string pointers, no problem just returning `""` in the error case.

account/manager.go
```diff
@@ -107,6 +108,65 @@ func (wt witnessType) witnessSize() (int, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	labelString := prefixTag + labelCompactJson.String()
```
guggero, line 166: Should use `fmt.Sprintf(prefixTag, string(labelJson))` here and add a `%s` to the `prefixTag` to avoid some gotchas around string conversion in Golang.

account/manager.go
```diff
@@ -548,11 +608,20 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
-			// and also attach some additional meta data.
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
-			label := makeTxnLabel(m.cfg.TxLabelPrefix, contextLabel)
+			// Attach additional meta data to transaction label.
+			isExpirySpend := false
+			txFee := btcutil.Amount(1)
+			balanceDiff := account.Value
+			contextLabel, err := actionTxLabel(
+				*account,
```
guggero, line 616: We usually write the arguments a bit more compact (while still respecting the 80 character limit). And the closing parenthesis should be on a new line for multi-line arguments.
So I'd write this like this:
```
		contextLabel, err := actionTxLabel(
				*account, CREATE, isExpirySpend,
				txFee, balanceDiff,
		)
```

account/manager.go
```diff
@@ -107,6 +108,65 @@ func (wt witnessType) witnessSize() (int, error) {
...
+	ExpiryHeight  uint32         `json:"expiry_height"`
+	OutputIndex   uint32         `json:"output_index"`
+	IsExpirySpend bool           `json:"expiry_spend"`
+	TxFee         btcutil.Amount `json:"tx_fee"`
+	BalanceDiff   btcutil.Amount `json:"balance_diff"`
+}
+
+// actionTxLabel returns a transaction label for use with account
+// modification actions.
+func actionTxLabel(account Account, action Action, isExpirySpend bool,
```
guggero, line 138: Can just pass in a pointer to the account, no need to de-reference.

account/manager.go
```diff
@@ -1316,10 +1399,8 @@ func (m *manager) RenewAccount(ctx context.Context,
 	// rest of the flow. This should request a signature from the auctioneer
 	// and assuming it's valid, broadcast the update transaction.
 	modifiers = append(modifiers, StateModifier(StatePendingUpdate))
-	modifiedAccount, spendTx, err := m.spendAccount(
-		ctx, account, packet, spendWitnessType, modifiers, false,
-		bestHeight,
-	)
+	modifiedAccount, spendTx, err := m.spendAccount(ctx, account, RENEW, packet,
```
guggero, line 1402: nit: don't change the formatting for a multi-line method call, here and a few other places in this file.

account/manager.go
```diff
@@ -1471,14 +1550,20 @@ func (m *manager) CloseAccount(ctx context.Context, traderKey *btcec.PublicKey,
...
 	packet *psbt.Packet, witnessType witnessType, modifiers []Modifier,
-	isClose bool, bestHeight uint32) (*Account, *wire.MsgTx, error) {
+	bestHeight uint32) (*Account, *wire.MsgTx, error) {
+
+	spendingAccount := *account
 
 	// In case we're not closing the account, let's now locate our
 	// re-created account output.
+	//
+	// TODO(ffranr): accountOutputIdx == -1 shouldn't be used to signal that
```
guggero, line 1562: It's actually only used in a single place deep down in the call stack. And the actual value is used there as well if it's not -1. So I don't think it make sense to thread through the action all the way.

guggero (commented):
Looks pretty good, I like the new label format!

I'm wondering if we should also start writing the same account modification data into the database alongside the accounts to make retrieval even easier and independent of lnd's label database (re-scanning the wallet can drop the labels, which users do sometimes).
Then we kind of have redundant information, but the new label is also just nice for the user to see what happened, so I don't think it's overkill.

account/manager.go
ffranr: My thought process: `account Account` would make it clear that `actionTxLabel` does not modify `account`.

But I can see that passing pointers is common in the codebase. Will change as you suggest.

account/manager.go
ffranr: Ah not intentional, IDE formatting. Will fix.

account/manager.go
ffranr: I'm going to try `fmt.Sprintf(prefixTag, labelJson)` because I think the `%s` should take care of converting `labelJson` into a string.

account/manager.go
```diff
@@ -548,10 +601,24 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 					"be greater than 0", acctKey)
 			}
 
-			// If we have a label prefix, then we'll apply that now
-			// and also attach some additional meta data.
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
+			// Attach additional meta data to transaction label.
+			//
+			// TODO(ffranr): Tx inputs are required to calculate the fee. The fee
```
guggero, line 606: :+1:

account/manager.go
```diff
@@ -548,10 +601,24 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
+			// will be added to the tx label. m.cfg.Wallet.SendOutputs returns
+			// and selects tx inputs but also requires the label as an argument.
+			// Instead, the tx should be constructed via m.cfg.Wallet.FundPsbt
+			// which would give us an opportunity to inspect tx inputs before
+			// creating a tx label.
+			txFee := btcutil.Amount(0)
+
+			balanceDiff := account.Value
+			contextLabel, err := actionTxLabel(
+				account, CREATE, false,
```
guggero, line 616: nit: should use the full 80 characters limit for the argument list, so I think it would fit even compacter:
```
			contextLabel, err := actionTxLabel(
				account, CREATE, false, txFee, balanceDiff,
			)
```
Same below.

account/manager.go
```diff
@@ -624,9 +691,18 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 				}
 			}
 
-			acctKey := account.TraderKey.PubKey.SerializeCompressed()
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
+			txFee, err := m.deriveFeeFromTx(ctx, accountTx)
+			if err != nil {
```
guggero, line 695: Since this depends on the TX fee being calculated via a call to `lnd`, we shouldn't abort/return here in case of an error. I guess just logging and setting the fee to zero is sufficient.

account/manager.go
```diff
@@ -107,6 +108,58 @@ func (wt witnessType) witnessSize() (int, error) {
...
+		Key:           key,
+		Action:        action,
+		ExpiryHeight:  account.Expiry,
+		OutputIndex:   account.OutPoint.Index,
+		IsExpirySpend: isExpirySpend,
+		TxFee:         txFee,
+		BalanceDiff:   balanceDiff,
+	}
+	labelJson, err := json.Marshal(label)
+	if err != nil {
```
guggero, line 155: Since this error should only happen when using the JSON API incorrectly, I think we could simplify this helper function by removing the error return type and instead just logging the error here and return a default label (e.g. `pool -- <action>`) as a fall back. That would remove some of the `if err != nil ` checks further down in the diff.

account/manager.go
```diff
@@ -798,6 +883,32 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 	return nil
 }
 
+// deriveFeeFromTx derives the transaction fee from a given transaction message.
+func (m *manager) deriveFeeFromTx(ctx context.Context,
+	tx *wire.MsgTx) (btcutil.Amount, error) {
+
+	// Input and output values are required to calculate the tx fee. However,
```
guggero, line 890: nit: line length in several places in the file. You need to set up your IDE to use 8 spaces per tab character ("Tab size" in "Editor -> Code Style -> Go", tab "Tabs and Indents" in GoLand).

guggero: Also, feel free to create a PR that adds the same `.editorconfig` from `lnd` which GoLand seems to use to auto-configure this.

account/manager.go
```diff
@@ -1568,6 +1678,22 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
 	return account, spendTx, nil
 }
 
+// deriveFeeFromPsbtPacket returns the transaction fee from a given PSBT packet.
+func deriveFeeFromPsbtPacket(packet psbt.Packet) btcutil.Amount {
```
guggero, line 1682: Same comment here about using a pointer directly. I get that it makes it more obvious that we're not modifying the packet here, but we don't really follow that pattern throughout the code base. Also, if you look at `psbt.Packet` it basically consists of pointer types only (a slice is a pointer), so this wouldn't protect the content of the package from being modified really anyway.

This could also just be called `deriveFeeFromPsbt()`. The `Packet` name is just something that's being used in the `psbt` package to make the naming nicer, otherwise it would be called `psbt.Psbt` which is discouraged by the Golang naming suggestions. So basically the `Packet` in this function name is kind of an implementation detail that can be omitted :)

guggero (commented):
Very nice work, looks very good!
Just a few comments about error handling and nits, nothing major :tada:

account/manager.go
ffranr: Good idea!

account/manager.go
ffranr: I'm concerned that a user might not notice that an error has occurred if the fee is a valid value (0). Is a log message really sufficient? For example, the sum total fee might still seem reasonable.

But I'm not sure what we can do otherwise! What are your thoughts please?

ffranr: I suppose ideally we would set the fee to json `null`. Which I think means using a `*btcutil.Amount` here: https://github.com/lightninglabs/pool/pull/396/files#diff-94465ef38c605b8bddd6e99ee32de13f57b8d6135d6be738f750498e82f19c8eR132

Do you agree?

guggero: Given that a fee of 0 sats is an invalid value for any on-chain transaction, I think it's enough to not break the account funding flow because of a call to lnd that is only for statistics reason. And given that we are going to migrate this to PSBT funding which gives us way more info anyway, I think that's okay for now.

ffranr: I'll go with your suggestion and use 0 fee.

> I think it's enough to not break the account funding flow because of a call to lnd that is only for statistics reason

I'm not sure I completely understand what you mean. Wouldn't the use of `*btcutil.Amount` would be isolated to labels? Shouldn't touch the rest of the account funding flow.

guggero: Ah, my comment was more about whether or not we should return an error or just log something.

Okay, I think I know what you mean now why `nil` is better than 0. That will cause any calculation in the UI to fail (or require special handling) instead of just silently summing up with 0. So yeah, maybe we should actually do that. Sorry for the back and forth, context switching a bit too much at the moment.

account/manager.go
```diff
@@ -1568,6 +1672,22 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
 	return account, spendTx, nil
 }
 
+// deriveFeeFromPsbt returns the transaction fee from a given PSBT packet.
+func deriveFeeFromPsbt(packet *psbt.Packet) btcutil.Amount {
```
positiveblue, line 1676: comment: not for this PR but this code is going to be replicated in multiple code repo ([ex in Taro](https://github.com/lightninglabs/taro/pull/184/files#diff-3ef0dca269260a596c1904601701d590c3e180e566016a8de2ced5fb3423fa3cR746). It may be worth adding it to btcd or lnd libs.

```
var packet *psbt.Packet
packet.GetTxFee() 
```

account/manager.go
```diff
@@ -1551,14 +1656,13 @@ func (m *manager) spendAccount(ctx context.Context, account *Account,
...
-	// As this is a generic account modification, we'll add some additional
-	// information to make accounting for this transaction a bit easier.
-	deposit := prevAccountState.Value < account.Value
-	acctKey := account.TraderKey.PubKey.SerializeCompressed()
-	contextLabel := fmt.Sprintf(" poold -- AccountModification(acct_key=%x, "+
-		"expiry=%v, deposit=%v, is_close=%v)", acctKey,
-		witnessType == expiryWitness, deposit, isClose)
-
+	// Generate transaction label.
+	isExpirySpend := witnessType == expiryWitness || witnessType == expiryTaproot
```
positiveblue, line 1660: nit:

```
// IsExpirySpend returns true if the witness is taking an expiration path.
func (wt witnessType) IsExpirySpend() bool {
	switch wt {
	case expiryWitness, expiryTaproot:
		return true
	default:
		return false
	}
}
```

positiveblue (approved):
LGTM :100:

account/manager.go
guggero: I guess we could use the `psbt.SumUtxoInputValues()` function here that is slightly safer as it checks whether there is a witness UTXO. I forgot I added that a while ago :sweat_smile: 
But then we have the question of how to handle the error again...

ffranr: btcd PR: https://github.com/btcsuite/btcd/pull/1916

account/manager.go
ffranr: Good idea, and thank you for the code.

account/manager.go
```diff
@@ -107,6 +108,70 @@ func (wt witnessType) witnessSize() (int, error) {
 	}
 }
 
+// IsExpirySpend returns true if the witness is taking an expiration path.
+func (wt witnessType) IsExpirySpend() bool {
```
guggero, line 112: This should be moved to where the `witnessType` is declared.

account/manager.go
```diff
@@ -107,6 +108,70 @@ func (wt witnessType) witnessSize() (int, error) {
...
+	}
+	labelJson, err := json.Marshal(label)
+	if err != nil {
+		log.Errorf("Internal error: failed to serialize json "+
+			"from %v", label)
+		return fmt.Sprintf(prefixTag, action)
+	}
+
+	labelString := fmt.Sprintf(prefixTag, labelJson)
+	return labelString
```
guggero, line 172: nit: just return `fmt.Sprintf()` directly?

account/manager.go
```diff
@@ -624,9 +702,20 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
 				}
 			}
 
-			acctKey := account.TraderKey.PubKey.SerializeCompressed()
-			contextLabel := fmt.Sprintf(" poold -- "+
-				"AccountCreation(acct_key=%x)", acctKey)
+			txFee, err := m.deriveFeeFromTx(ctx, accountTx)
+			fee := &txFee
```
guggero, line 706: I guess in this case we should just return a pointer to `btcutil.Amount` so we don't need to do this here.

account/manager.go
```diff
@@ -624,9 +702,20 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
+			if err != nil {
+				log.Errorf(
```
guggero, line 708: Ah, all calls to `log.Xxxf()` and `fmt.Errorf()` should use the compact formatting (that way they're easily visually distinguishable from an actual method/function call). See lnd's formatting guide.

account/manager.go
```diff
@@ -624,9 +702,20 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
+					"Failed to derive fee from "+
+						"transaction: %v", accountTx,
+				)
+				fee = nil
+			}
+			balanceDiff := account.Value
+			contextLabel := actionTxLabel(
+				account, CREATE, false, fee,
```
guggero, line 716: nit: I think `balanceDiff` should also fit on the previous line.

account/manager.go
```diff
@@ -766,12 +855,22 @@ func (m *manager) resumeAccount(ctx context.Context, account *Account, // nolint
...
 	case StatePendingClosed:
-		acctKey := account.TraderKey.PubKey.SerializeCompressed()
-		contextLabel := fmt.Sprintf(" poold -- "+
-			"AccountClosure(acct_key=%x)", acctKey)
+		txFee, err := m.deriveFeeFromTx(ctx, account.LatestTx)
+		fee := &txFee
+		if err != nil {
+			log.Errorf(
+				"Failed to derive fee from "+
+					"transaction: %v", account.LatestTx,
```
guggero, line 863: nit: also log the causing error in these cases. So `... from transaction %v: %v, account.LatestTx, err)`).

guggero (commented):
Getting very close here! Only a few nits remaining :tada:

account/manager.go
ffranr: `witnessType` is declared within this file here: https://github.com/lightninglabs/pool/pull/396/files/d76f0f11114ba07e0d54a63ca9d233f37a8f2286#diff-94465ef38c605b8bddd6e99ee32de13f57b8d6135d6be738f750498e82f19c8eR64

Above the `IsExpirySpend()` declaration are other `witnessType` methods:
```go
func (wt witnessType) scriptVersion() poolscript.Version {...}
...
func (wt witnessType) witnessSize() (int, error) {...}
```

Perhaps constants defined in between lines 71 and 81 should be moved elsewhere, is that what you mean? Let me know if there's a better place for `IsExpirySpend()` or if I've missed something.

guggero: Oh, I'm sorry, that's my fault... I thought the `witnessType` was declared in a different file (but the same package). Should've clicked the arrow on the diff to show more of the code before L111. Never mind :see_no_evil:

account/manager.go
ffranr: `eDiff,` would be over the 80 char column (according to my IDE). Is that acceptable? Asking because I noticed that some lines within the codebase do go over.

I see that the guide asks for a "best effort basis": https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md#80-character-line-length

account/manager.go
ffranr: No problem! Review tool has room for improvement to be honest. Not sure if I know of better.

account/manager.go
guggero: Hmm, weird. Then maybe some setting isn't right in your IDE perhaps?
This is what it looks in my IDE after merging L716 and L717 (visual aid set at 80 characters).
Screenshot from 2022-11-04 19-42-53


Yes, it's best effort and we usually don't make a big deal out of a few characters.

dstadulis:
prematurely merged, follow up PR to continue improvements and reference this PR

account/manager.go
ffranr: My mistake. I had inlay parameter hints turned on.

ffranr:
@guggero I've opened this PR to address the rest of your comments: https://github.com/lightninglabs/pool/pull/413

If you have further comments you are welcome to leave them on either PR. Sorry for the inconvenience!

PR #403, opened 2022-10-31
Add channel reserve as a constraint during order matching/expression

ffranr:
#### Pull Request Checklist
  used.


auctioneerrpc/auctioneer.proto
```diff
@@ -1083,6 +1083,11 @@ message ServerBid {
     Signals if this bid is interested in a zero conf channel or not.
     */
     bool zero_conf_channel = 9;
+
+    /*
+    Signals if this bid is interested in a minimal reserve channel.
+    */
+    bool min_reserve_channel = 10;
```
dstadulis, line 1090: My understanding is that there isn't anything endogenous to a "minimum reserve" channel that distinguishes it from another channel.  That the distinguishing factor need to be produced by Pool, is the bidder is in essence setting a stipulation on the asker's performance, that their node update channel balances to channel-reserve levels Pool hasn't supported before (and hence what is considered "minimum reserve" is super nebulous and could change in the future).

Let's discuss internally how to best create matches around this property

dstadulis:
For the record, per 2022-11-08 meeting this feature will be removed from sprints until work could resume in 2023-Q1

PR #404, opened 2022-10-31
cmd: use `--amt` in submit order description

### LINKED ISSUE #401: Replace --amount by --amt in outbound order description
The "pool orders submit" description mentions `--amount` for the outbound market, but I believe it should be `--amt`.

NB: I prefer `--amount` over `--amt`, but changing the API seems to be a different issue.

pool v0.6.0-beta

positiveblue:
Use the right flag in the subcommand help message.

Fix #401


guggero (approved):
LGTM :tada:

ffranr (approved):

PR #406, opened 2022-10-31
cmd: allow `min_chan_amt > amt` for outbound liquidity orders

positiveblue:
Skip the extra check in the cli validation flow for the markets where this option is a valid.

This bug only applied to the validation logic in the cli package, poold and the backend services validate the orders correctly.

### DIFF
--- a/cmd/pool/order.go
+++ b/cmd/pool/order.go
@@ -210,6 +210,105 @@ func promptForConfirmation(msg string) bool {
 	}
 }
 
+// isBidOrder returns true for the Bid type.
+func isBidOrder(orderType order.Type) bool {
+	return orderType == order.TypeBid
+}
+
+// isInboundLiquidityOrder returns true when the `auction_type` param is set
+// to the inbound liquidity market.
+func isInboundLiquidityOrder(ctx *cli.Context) bool {
+	return ctx.String("auction_type") == auctionTypeInboundLiquidity
+}
+
+// isOutboundLiquidityOrder returns true when the `auction_type` param is set
+// to the outbound liquidity market.
+func isOutboundLiquidityOrder(ctx *cli.Context) bool {
+	return ctx.String("auction_type") == auctionTypeOutboundLiquidity
+}
+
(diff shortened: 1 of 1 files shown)


cmd/pool/order.go
```diff
@@ -210,6 +210,42 @@ func promptForConfirmation(msg string) bool {
...
 }
 
+// parseMinChanAmount parses and validates the value for the `min_chan_amt`
+// parameter.
+func parseMinChanAmount(ctx *cli.Context, amt uint64) (btcutil.Amount, error) {
+	minChanAmt := btcutil.Amount(ctx.Uint64("min_chan_amt"))
+
+	auctionType := ctx.String("auction_type")
+	if auctionType == auctionTypeInboundLiquidity {
+		if minChanAmt == 0 {
```
guggero, line 220: Can we preserve the comment about the default of `10%` here please? Just to give this some context.

guggero (approved):
LGTM, just one nit :tada:

cmd/pool/order.go
```diff
@@ -210,6 +210,44 @@ func promptForConfirmation(msg string) bool {
 	}
 }
 
+// parseMinChanAmount parses and validates the value for the `min_chan_amt`
+// parameter.
+func parseMinChanAmount(ctx *cli.Context, amt uint64) (btcutil.Amount, error) {
+	minChanAmt := btcutil.Amount(ctx.Uint64("min_chan_amt"))
```
ffranr, line 216: I don't think this function does much parsing of `min_chan_amt`. It might be a good idea if this function was
```
validateMinChanAmount(minChanAmount btcutil.Amount, amt btcutil.Amount, auctionType) error
```

and move the `getMinChanAmount` functionality into a separate function. And then it would make more sense that a validator accepts `amt` as an argument. And testing might be simpler.

ffranr (changes requested):

ffranr (approved):
Looks good! Nice refactoring.

I've only read the code. Let me know if I should do more.

PR #413, opened 2022-11-07
account: transaction labels code cleanup

ffranr:
PR https://github.com/lightninglabs/pool/pull/396 was merged prematurely. This PR addresses any remaining review comments.

### DIFF
--- a/account/manager.go
+++ b/account/manager.go
@@ -164,12 +164,11 @@ func actionTxLabel(account *Account, action Action, isExpirySpend bool,
 	labelJson, err := json.Marshal(label)
 	if err != nil {
 		log.Errorf("Internal error: failed to serialize json "+
-			"from %v", label)
+			"from %v: %v", label, err)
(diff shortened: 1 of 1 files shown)


account/manager.go
```diff
@@ -908,7 +898,7 @@ func (m *manager) deriveFeeFromTx(ctx context.Context,
 	for _, txIn := range tx.TxIn {
 		spendTx, err := m.locateTxByHash(ctx, txIn.PreviousOutPoint.Hash)
 		if err != nil {
-			return 0, nil
+			return nil, nil
```
guggero, line 901: Oh, I guess this slipped by in my last review. This should return the error.

guggero (commented):
Thanks for the fixes. Looks good, just one small issue.

account/manager.go
ffranr: Good catch. Thanks.

guggero:
Trivial diff, going to merge with just a single review.

PR #416, opened 2022-11-08
script: add `verify-install.sh` for signed releases

positiveblue:
The process is similar to the one used in LND. The repo includes a list of keys that can be used to sign a valid relsease.

By now, we need a singature of 2 out of 4  to mark the release as valid. This number can increase when more keys are added to the list.

Added keys: @ffranr @guggero @Roasbeef @positiveblue 

NOTE: we need to ensure that we have reproducible builds before merging this.


scripts/keys/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Verifying the Release
+
+In order to verify the release, you'll need to have gpg or gpg2 installed on 
+your system. Once you've obtained a copy (and hopefully verified that as well),
+you'll first need to import the keys that have signed this release if you 
+haven't done so already:
+
+```
+curl https://raw.githubusercontent.com/lightningnetwork/lnd/master/scripts/keys/guggero.asc | gpg --import
```
guggero, line 23: nit: change this to the Pool repo? Or use a key server as shown [here](https://github.com/lightninglabs/faraday/releases/tag/v0.2.8-alpha)?

guggero: Or, thinking about it, should we instead use the keys in the lnd repo for verification in the first place? Since if you trust the lnd signers, you probably also trust them to sign for Pool?
Meaning we wouldn't commit any keys to this repo in the first place (and update the `verify-install.sh` script accordingly)?

guggero (commented):
Thanks for adding the script and keys! I verified the keys match the ones in `lnd` and the digests in the `verify-install.sh` script.

Left a suggestion about not committing the keys to this repo, would make the replacement of a key a bit simpler too. Let me know what you think.

scripts/verify-install.sh
```diff
@@ -0,0 +1,324 @@
+#!/bin/bash
```
guggero, line 1: This file needs `chmod +x` by the way.

dstadulis:
This PR won't have pay off until the next gittag occurs, correct? e.g. to verify the tags and signatures the verifier will need sigs and a gittag to run the script over.

If this the case, prioritization should drop to P3.
There are a few places in the PR cryptographic integrity can be improved (e.g. only import public keys which match stipulated fingerprints) and security improved (minimum TLS version to prevent transport encryption downgrade).

Let's prio on this during some of the time slack before our next release

PR #418, opened 2022-11-08
scripts: use `git describe --tags`

### LINKED ISSUE #417: build: use latest tag, even when it has no annotations
The output for the latest pool version (`v0.6.1-beta`) to `./pool --version` is `pool version 0.6.1-beta commit=v0.6.0-beta-18-g5e95640ec64759681b49c0f49adf52abd06cb45c` when it should be `pool version 0.6.1-beta commit=v0.6.1-beta`.

The problem is that the tags where not annotated (`git tag -S v0.6.1-beta`) when I created them and we are using `git describe` in our release scripts.

positiveblue:
When building a release we want to use the right tag, even if it was not annotated.

Fix #417


guggero (approved):
LGTM :tada:

ffranr (approved):

PR #421, opened 2022-12-02
multi: configurable RequestShutdown call-back in config

ellemouton:
Add `RequestShutdown` option to Pool's config that callers running pool as a subserver can set to whatever they want. By default the function will be a no-op. In the case where Pool is running as a standalone binary, the call-backs will call the corresponding signal interceptor function.

### DIFF
--- a/config.go
+++ b/config.go
@@ -171,6 +171,11 @@ type Config struct {
 	// the server is going to use to listen for (and issue) shutdown
 	// commands on.
 	ShutdownInterceptor signal.Interceptor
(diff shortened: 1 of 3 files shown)


positiveblue (approved):
LGTM :rocket: 

We can rethink in the future a better default behaivour for when it runs as a sub server (so we do not have critical errors that should make the app stop being totally ignored)

guggero (commented):
Post merge ACK :tada:

PR #422, opened 2022-12-02
version: bump to v0.6.2-beta

positiveblue:
The previous version of pool could not be used properly as a subserver because of how it was handling the (sub)-server shutdown.

`v0.6.2-beta` includes a fix for it.


ellemouton (approved):
Thanks @positiveblue  :)

Issue #405, opened 2022-10-31
Outbound ask order: allow any confirmation constraint


C-Otto:
When creating an outbound ask order, the only options for `--confirmation_constraints` are "only confirmed channels" and "only zero conf channels". If I allow zero conf channels, I probably also accept confirmed channels. As such, it should be possible to specify this constraint as a third option.

pool v0.6.0-beta

positiveblue:
Thank you for noticing @C-Otto.

The API allows set `NO_PREFERENCE`. The default value in the cli is `ONLY_CONFIRMED` (`confirmation_constraints=1)`. You can set `confirmation_constraints=0` to express that you want to match any type here.


I will update the cmd helper description.

C-Otto:
@positiveblue please reopen (see comment in PR)

guggero:
Fixed (for real now) by #423.

PR #423, opened 2022-12-05
cmd/pool: fix confirmation constraint description

guggero:
Fixes the rest of https://github.com/lightninglabs/pool/issues/405.

#### Pull Request Checklist
  used.


ffranr (approved):

ffranr:
I would have thought the options should be: 
* accept zero conf
* require confirmations

"2 for only zero conf" -> does that mean that if there is a conf then the liquidity wont be sold?

guggero:
This is for matchmaking only. With "q for only zero conf" the asker only wants to be matched with bids that are request zero conf, no "normal" orders.
A zero conf channel becomes a normal channel after 6 confirmations.

PR #424, opened 2022-12-11
Bumped reccommended LND version

rstmsn:
Bumped reccommended LND version to 0.15.4-beta

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,5 +95,5 @@ $ make install
 This will install the binaries into your `$GOPATH/bin` directory.
 
 ## Compatibility
-Lightning Pool requires `lnd` version `0.14.3-beta` or higher (`v0.15.0-beta` or
+Lightning Pool requires `lnd` version `0.14.3-beta` or higher (`v0.15.4-beta` or
 later is recommended).


guggero (changes requested):
Thanks for flagging this outdated sentence. The minimum required version [was bumped to `v0.15.1-beta`](https://github.com/lightninglabs/pool/blob/master/server.go#L43) a while back, so the first part of the sentence should also be updated. Maybe we can even link to the section in the code so it's easier to check?

rstmsn:
> Thanks for flagging this outdated sentence. The minimum required version [was bumped to `v0.15.1-beta`](https://github.com/lightninglabs/pool/blob/master/server.go#L43) a while back, so the first part of the sentence should also be updated. Maybe we can even link to the section in the code so it's easier to check?

Would you agree that it would make more sense to also bump the minimum required version in the code to 0.15.4-beta, as this is the update that patches the large taproot transaction bug ?
Presumably loop will fail to run if it's operating on top of an lnd that is not patched for this issue?

Grateful for thoughts in case I have overlooked something.

guggero:
> Would you agree that it would make more sense to also bump the minimum required version in the code to 0.15.4-beta, as this is the update that patches the large taproot transaction bug ?

Yes, I think that makes sense.

guggero:
Replaced by https://github.com/lightninglabs/pool/pull/435.

PR #426, opened 2022-12-29
rpcserver: format rpcLog error

### LINKED ISSUE #425: broken debug string
I see the following log message, where "%v" should have been replaced by some more helpful string.

```
[ERR] RPCS: Unknown server error: %v server sent unexpected error
```

poold version 0.6.1-beta commit=v0.6.0-beta-18-g5e95640ec64759681b49c0f49adf52abd06cb45c-dirty

positiveblue:
We were using `Error(<string>, <err>)` instead of `Errorf(<string>, <err>)`. The error was appended at the end of the message instead of replacing the `%v` tag in the message.

Fixes #425

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -11,6 +11,9 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
   && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+  && chmod -R 777 /tmp/build/ \
+  && git config --global --add safe.directory /build 
+  # The last line is needed to ensure that go build is able to gather 
+  # information from the vsc used in the builds to get the commit hash. 
 
 WORKDIR /build
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -238,7 +238,9 @@ func (s *rpcServer) Stop() error {
 }
 
 // serverHandler is the main event loop of the server.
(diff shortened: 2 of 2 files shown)


tools/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
   && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+  && chmod -R 777 /tmp/build/ \ 
+  && git config --global --add safe.directory /build
```
guggero, line 15: Doesn't seem to be working just yet...

guggero (approved):
LGTM :tada:

tools/Dockerfile
positiveblue: Fixed in #427

I am still not sure how it can be solved in lnd without making the folder "safe" for git :confused:

guggero (approved):

tools/Dockerfile
```diff
@@ -11,6 +11,9 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
   && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+  && chmod -R 777 /tmp/build/ \
+  && git config --global --add safe.directory /build
```
guggero, line 15: I guess this was a rebase issue. I don't think we actually still need this.

positiveblue: Yes it was T.T I will delete it in a next PR

PR #427, opened 2023-01-05
build: update go to v1.19.4-beta

positiveblue:
- Since Go v1.19.0 `go fmt` also formats commented code so files need to be reformated.
- Includes the fix for the lint complains about vsc

P.S: I can extract some of the steps in the actions in a comment one for  `Seting up Golang environment` if needed in this PR too

PR #429, opened 2023-01-05
rpc: populate allowed/not allowed node ids when listing orders

### LINKED ISSUE #428: rpc: allowed/not Allowed node ids not populated when listing orders.
The fields are stored properly in the db and properly taken into account during the matches but the `order list` rpc method does not populate the values in the response.

```
pool o s bid  --allowed_node_id=${ALLOWED_NODE_ID} ...
// submission succeeds

pool o l | jq .bids[].allowed_node_ids
// returns null
```

positiveblue:
Fixes #428 
Depends on #427

### DIFF
--- a/order/rpc_parse.go
+++ b/order/rpc_parse.go
@@ -127,23 +127,19 @@ func ParseRPCOrder(version, leaseDuration uint32,
 			"at the same time")
 	}
 
-	kit.AllowedNodeIDs = make([][33]byte, len(details.AllowedNodeIds))
-	for idx, nodeID := range details.AllowedNodeIds {
-		if _, err := btcec.ParsePubKey(nodeID); err != nil {
-			return nil, fmt.Errorf("invalid allowed_node_id: %x",
-				nodeID)
-		}
-		copy(kit.AllowedNodeIDs[idx][:], nodeID)
+	allowedNodeIDs, err := UnmarshalNodeIDSlice(details.AllowedNodeIds)
+	if err != nil {
+		return nil, fmt.Errorf("invalid allowed_node_ids: %v", err)
 	}
+	kit.AllowedNodeIDs = allowedNodeIDs
 
-	kit.NotAllowedNodeIDs = make([][33]byte, len(details.NotAllowedNodeIds))
-	for idx, nodeID := range details.NotAllowedNodeIds {
-		if _, err := btcec.ParsePubKey(nodeID); err != nil {
-			return nil, fmt.Errorf("invalid not_allowed_node_id: "+
-				"%x", nodeID)
-		}
-		copy(kit.NotAllowedNodeIDs[idx][:], nodeID)
+	notAllowedNodeIDs, err := UnmarshalNodeIDSlice(
+		details.NotAllowedNodeIds,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid not_allowed_node_ids: %v", err)
 	}
+	kit.NotAllowedNodeIDs = notAllowedNodeIDs
 
 	kit.IsPublic = details.IsPublic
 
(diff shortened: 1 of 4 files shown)


guggero (changes requested):

rpcserver.go
```diff
@@ -1698,6 +1698,20 @@ func (s *rpcServer) ListOrders(ctx context.Context,
 			}
 		}
 
+		allowedNodeIDs := make(
+			[][]byte, 0, len(dbOrder.Details().AllowedNodeIDs),
+		)
+		for _, nodeID := range dbOrder.Details().AllowedNodeIDs {
+			allowedNodeIDs = append(allowedNodeIDs, nodeID[:])
```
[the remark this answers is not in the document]
positiveblue: Right, tests added for it :+1:

To have this done everywhere with the same structure we could have a `serialize.go` in each package that has the logic to serialize/deserialize obj from/to rpc and `rpcserver.go` calls those. In that way we can keep the testing (with invalid inputs) in the package itself. Similar to the current `rpc_parse.go` in orders but without splitting the logic between that file and the `rpcserver.go`.

order/rpc_parse.go
```diff
@@ -524,6 +520,43 @@ func ParseRPCSign(signMsg *auctioneerrpc.OrderMatchSignBegin) (AccountNonces,
...
+		copy(nodeID, nodeIDs[i][:])
+
+		res = append(res, nodeID)
+	}
+
+	return res
+}
+
+// UnmarshalNodeIDSlice returns a slice of node ids from a flatten version.
+func UnmarshalNodeIDSlice(slice [][]byte) ([][33]byte, error) {
```
guggero, line 539: Nice! We can re-use this on the server side as well.

order/rpc_parse.go
```diff
@@ -524,6 +520,43 @@ func ParseRPCSign(signMsg *auctioneerrpc.OrderMatchSignBegin) (AccountNonces,
 	return nonces, prevOutputs, nil
 }
 
+// MarshalNodeIDSlice returns a flatten version of an slice of node ids to be
```
guggero, line 523: nit: s/flatten/flattened/

guggero (approved):

tools/Dockerfile
```diff
@@ -11,6 +11,9 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
   && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+  && chmod -R 777 /tmp/build/ \
+  && git config --global --add safe.directory /build 
+  # The last line is needed to ensure that go build is able to gather
```
guggero, line 16: Can be removed.

tools/Dockerfile
```diff
@@ -11,9 +11,6 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
   && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/ \
-  && git config --global --add safe.directory /build 
-  # The last line is needed to ensure that go build is able to gather 
-  # information from the vsc used in the builds to get the commit hash. 
+  && chmod -R 777 /tmp/build/
```
positiveblue, line 14: reverts a commit that I pushed by error in #426

jbrill (approved):
LGTM 👍 very comprehensive tests. I was unable to get the fields showing in an end to end test as it looks like there's some additional field marshaling we need to do subasta-side.

Issue #431, opened 2023-01-10
ListLeases: fix lease chain fee estimation for Taproot accounts


guggero:
There is a [TODO](https://github.com/lightninglabs/pool/blob/master/rpcserver.go#L2424) in the code from when we introduced Taproot channels. The chain fees paid for executing an order is dependent on the account version (P2WSH vs. P2TR). But at the time of the Taproot upgrade PR it looked like it wouldn't be easy to find out what version an account was _at the time of the executed batch_, so the calculation still assumes P2WSH for every batch. But it looks like we can just use the `Version` of the account in the `Accounts` map of the batch.
I'm not sure what version of the account will be stored in the batch if in that batch the account is auto-upgraded from P2WSH (version 0) to P2TR (version 1). So that might be an edge case to consider and test (there are integration tests covering all of this).

PR #434, opened 2023-01-19
Lease chain fee estimation calculation includes account version

ffranr:
I need to add a test to cover this aspect:

> what version of the account will be stored in the batch if in that batch the account is auto-upgraded from P2WSH (version 0) to P2TR (version 1)

This PR addresses this issue: https://github.com/lightninglabs/pool/issues/431

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2266,6 +2266,9 @@ func (s *rpcServer) prepareLeasesResponse(ctx context.Context,
 		// paid.
 		totalNumChannels := 0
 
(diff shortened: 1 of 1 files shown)


ffranr:
I think the code should be correct at this point. I need to add an itest to subasta to make sure that account updates are handled correctly.

I think that the chain fee estimation code should be refactored out of the `prepareLeasesResponse` method. And then unit tests can be written that the correct fees are estimated for a given batch.

guggero:
!lightninglabs-deploy mute

PR #436, opened 2023-02-08
account recovery: increase default number of keys to try

guggero:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/470.

#### Pull Request Checklist
  used.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -3,6 +3,7 @@ FROM golang:1.19.4
 RUN apt-get update && apt-get install -y git
 ENV GOCACHE=/tmp/build/.cache
 ENV GOMODCACHE=/tmp/build/.modcache
+ENV GOFLAGS="-buildvcs=false"
 
 COPY . /tmp/tools
 
--- a/auctioneer/client.go
+++ b/auctioneer/client.go
@@ -37,12 +37,12 @@ import (
 const (
 	reconnectRetries = math.MaxInt16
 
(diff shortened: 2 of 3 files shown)


positiveblue (approved):
Thanks :pray:

tools/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM golang:1.19.4
 RUN apt-get update && apt-get install -y git
 ENV GOCACHE=/tmp/build/.cache
 ENV GOMODCACHE=/tmp/build/.modcache
+ENV GOFLAGS="-buildvcs=false"
```
positiveblue, line 6: should we add it to the dockerfile in the same way that is done in the other projects?

guggero: Do you mean the `git config --global --add safe.directory /build`? (We fixed it a number of times, not sure which one actually fixes what).
To be hones, this seems clearer as it directly fixes the issue and makes it more clear that it's a Golang issue (that might be fixed in the future), rather than implying it was some sort of `git` problem. I can go ahead and fix it in all other repos in this manner.

positiveblue: `fixes the issue and makes it more clear that it's a Golang related issue` :1st_place_medal: 

 Do not fix it in all the other repos, I will change them and tag you in the PR :+1:

ffranr (approved):

ffranr (approved):
Thank you!

PR #446, opened 2023-03-07
build: remove Pool release targets no longer built for lnd

dstadulis:
lnd PR https://github.com/lightningnetwork/lnd/pull/7252/ removed support for 10 OS / architectures

As Pool requires lnd to run, this PR replicates the build list of lnd for Pool.

It is assumed that if a developer builds for an unsupported OS/architecture for their lnd binaries, the same will be done for Pool


positiveblue (approved):
[no diff](https://github.com/lightningnetwork/lnd/blob/master/make/release_flags.mk#L14)

guggero (approved):

PR #449, opened 2023-05-16
multi: stop using lnwire.WriteElement

morehouse:
We intend to eventually remove `lnwire.WriteElement` from LND in favor of the faster statically-typed write functions.  Switch to using those faster functions in Pool.

#### Pull Request Checklist
  used.


ffranr (approved):
Looks good to me! Thanks for the well organised commit structure :+1:

PR #457, opened 2023-06-05
server: specify gRPC keepalive params

guggero:
With this commit we instruct the gRPC client to ping the server every 10 seconds to make sure the TCP connection is still alive.

Tested with Wireshark.

#### Pull Request Checklist
  used.

### DIFF
--- a/server.go
+++ b/server.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	proxy "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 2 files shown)


positiveblue (approved):
LGTM  :+1:

server.go
```diff
@@ -68,6 +70,11 @@ var (
 			"signrpc", "walletrpc", "chainrpc", "invoicesrpc",
 		},
 	}
+
+	// defaultClientPingTime is the default time we'll use for the client
+	// keepalive ping time. This means the client will ping the server every
+	// 10 seconds to make sure the TCP connection is still alive.
```
ffranr, line 76: nit: I don't think it will ping the server every 10 seconds. I think it will ping if it hasn't seen any activity after 10 seconds and then it will wait 20 seconds for a response.
```
	// After having pinged for keepalive check, the client waits for a duration
	// of Timeout and if no activity is seen even after that the connection is
	// closed.
	Timeout time.Duration // The current default value is 20 seconds.
```

from `google.golang.org/grpc@v1.39.0/keepalive/keepalive.go`

We might want to set `PermitWithoutStream` to `true` also so that pings occur even without active RPC.

ffranr (approved):
:hourglass: Nice! :watch:

server.go
guggero: You're right, it's only when there is no other activity. Updated the comment.

I don't think we want the keepalive without a stream. Because at some point the number of open TCP connections will become a limiting factor for the number of Pool clients we can support. So we don't want to allow the client to keep a TCP connection open for longer than necessary. So if they don't have an account and therefore no account subscription stream, they shouldn't be able to keep open the connection artificially.

Issue #458, opened 2023-06-17
Logs missing crucial information


C-Otto:
I'm using pool 0.6.4-beta. When I start poold via systemd, it fails after a second or so. The logs don't indicate anything useful:

```
2023-06-17 10:35:17.439 [INF] RPCS: Starting trader server
2023-06-17 10:35:17.522 [INF] RPCS: Connected to lnd node c-otto.de with pubkey 027ce055380348d7812d2ae7745701c9f93e70c1adeb2657f053f91df4f2843c71
2023-06-17 10:35:17.980 [INF] AUCT: Shutting down auctioneer client
```

However, when I start poold manually, I get this:

```
unable to start server: unable to start account manager: unable to resume account xxx: rpc error: code = Unknown desc = transaction rejected: output already spent
```

Please consider showing important information like this in the logs.

guggero:
@C-Otto thanks for the report. Do you know if `systemd` captures `stderr`? Because if there is an error _before_ `poold` has set up the logging system (e.g. configuration file cannot be parsed), then we now log to `stderr` (see https://github.com/lightninglabs/pool/pull/459) instead of `stdout`.

C-Otto:
I'm pretty sure systemctl is able to capture stderr, but in this case (copied over from my lnd setup) I disabled capturing output as I already have the log file. If errors like this aren't supposed to end up in the log file, I might have to tweak my setup. Otherwise, I'd like this message to appear in the log file (additionally to stderr).

C-Otto:
Sounds good to me! It's not too hard to check stdout/stderr, especially if the logs are empty.

PR #459, opened 2023-06-19
Fix startup logging issues

guggero:
Fixes https://github.com/lightninglabs/pool/issues/458.

#### Pull Request Checklist
  used.

### DIFF
--- a/cmd/poold/main.go
+++ b/cmd/poold/main.go
@@ -18,7 +18,7 @@ var (
 func main() {
 	err := start()
 	if err != nil {
-		fmt.Println(err)
+		_, _ = fmt.Fprintf(os.Stderr, "startup error: %v\n", err)
 		os.Exit(1)
 	}
 }
--- a/run.go
+++ b/run.go
@@ -51,15 +51,26 @@ func Run(cfg *Config) error {
 			)
 			http.Handle("/", profileRedirect)
 			//nolint:gosec
-			fmt.Println(http.ListenAndServe(cfg.Profile, nil))
+			err := http.ListenAndServe(cfg.Profile, nil)
+			if err != nil {
+				log.Errorf("Unable to run profiler: %v", err)
+			}
 		}()
 	}
 
 	trader := NewServer(cfg)
 	err = trader.Start()
 	if err != nil {
+		log.Errorf("Error starting server: %v", err)
 		return fmt.Errorf("unable to start server: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


cmd/poold/main.go
```diff
@@ -18,7 +18,7 @@ var (
 func main() {
 	err := start()
 	if err != nil {
-		fmt.Println(err)
+		_, _ = fmt.Fprintf(os.Stderr, "[poold] %v\n", err)
 		os.Exit(1)
```
ffranr, line 22: Consider the following so we don't have to specify a log format literal?
```
log.SetOutput(os.Stderr)
log.Fatalf("startup error: %v", err)
```

(`log.Fatalf` will call `os.Exit(1)`)

I'm not sure where the log formatted is initiated.

run.go
```diff
@@ -58,8 +58,16 @@ func Run(cfg *Config) error {
...
 	if err != nil {
+		log.Errorf("Error starting server: %v", err)
 		return fmt.Errorf("unable to start server: %v", err)
 	}
 	<-cfg.ShutdownInterceptor.ShutdownChannel()
-	return trader.Stop()
+
+	err = trader.Stop()
+	if err != nil {
+		log.Errorf("Error starting server: %v", err)
```
ffranr, line 68: `starting` -> `stopping`

ffranr (changes requested):
:+1: :+1:

cmd/poold/main.go
guggero: We don't really use the standard library logger anywhere. And if we did that here, then it would add its own header (date/time, file and line number) and stack trace. And the date format will be in another format, so setting that up would be additional code. We just want to make sure things end up in `stderr`, which we do here as well: https://github.com/lightningnetwork/lnd/blob/f9f8f6406b110168553d4e4b6aae7a857ef9fe3c/config.go#L800

run.go
```diff
@@ -58,8 +58,16 @@ func Run(cfg *Config) error {
 	trader := NewServer(cfg)
 	err = trader.Start()
 	if err != nil {
+		log.Errorf("Error starting server: %v", err)
```
positiveblue, line 61: :+1:

positiveblue (approved):
LGTM

I think we should make sure that we always print with `log.Errorf` before returning an error in this function (including the one in the "profile handler goroutine")

ffranr (approved):

PR #465, opened 2023-08-24
build: update to build against lnd v0.17.0

Roasbeef:
In this commit, we update the project to build against lnd v0.17.0. The `lnwire` signature parsing routines were changed, so we update all the relevant call sites to use the new functions.

Due to [this
PR](https://github.com/lightningnetwork/lnd/pull/7659/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R183), we now need to inherit the replace directive set, otherwise this build fails. IIUC, this now goes for any other project that uses the `lnrpc` package. We may want to reconsider the replace for lnd now that we're in the rc process.


positiveblue:
Some unit tests need to be updated `error storing sidecar ticket: malformed signature: too short: 71 bytes`

Should we bump `lndclient` first  and update it in this PR too?

go.mod
```diff
@@ -178,3 +179,5 @@ require (
 replace google.golang.org/grpc => google.golang.org/grpc v1.39.0
 
 replace github.com/lightninglabs/pool/auctioneerrpc => ./auctioneerrpc
+
+replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.30.0-hex-display
```
guggero, line 183: Should we also add the reason for the replace as a comment? See https://github.com/lightninglabs/taproot-assets/blob/main/go.mod#L208

And yes, unfortunately we need this replace in every project now... But at least we can remove the other one for `google.golang.org/grpc` if we update to the latest Aperture version.

guggero (changes requested):

Roasbeef:
Refreshed!

sidecar/tlv.go
```diff
@@ -350,11 +350,11 @@ func ESig(w io.Writer, val interface{}, buf *[8]byte) error {
 		var s [64]byte
 		if *v != nil {
 			rawSig := (*v).Serialize()
-			sig, err := lnwire.NewSigFromRawSignature(rawSig)
+			sig, err := lnwire.NewSigFromWireECDSA(rawSig)
```
guggero, line 353: Also need to use `NewSigFromECDSARawSignature` here.

sidecar/tlv.go
```diff
@@ -372,13 +372,22 @@ func DSig(r io.Reader, val interface{}, buf *[8]byte, l uint64) error {
 		}
 
 		if s != ZeroSignature {
-			wireSig := lnwire.Sig(s)
+			wireSig, err := lnwire.NewSigFromWireECDSA(s[:])
```
guggero, line 375: Also need to use `NewSigFromECDSARawSignature` here.

guggero:
Replaced by https://github.com/lightninglabs/pool/pull/468.

Issue #467, opened 2023-09-20
multi: add support for creating orders for unadvertised taproot channels


Roasbeef:
With lnd 0.17 nearly finalized, we should add support for making orders for taproot channels. For now, there's now script enforced lease variant, so it'll need to work with the normal frozen co-op close scheme. We'll also want to ensure that if a user sets the new order type, that we require the channel to _not_ be announced. 

Some relevant call sites we'll need to change: 

* https://github.com/lightninglabs/pool/blob/70dd1b107e9b19c25554b670952f4753e999766b/order/interfaces.go#L231-L247
* https://github.com/lightninglabs/pool/blob/70dd1b107e9b19c25554b670952f4753e999766b/chaninfo/channel_info.go#L147-L160
* https://github.com/lightninglabs/pool/blob/70dd1b107e9b19c25554b670952f4753e999766b/auctioneerrpc/auctioneer.proto#L363-L376

PR #468, opened 2023-09-22
Add support for matching Simple Taproot channels

guggero:
Bumps `lnd` to `v0.17.0-beta.rc4` and prepares to allow matching of Simple Taproot Channels.

~Depends on https://github.com/lightninglabs/pool/pull/470~

Fixes https://github.com/lightninglabs/pool/issues/467.
Replaces https://github.com/lightninglabs/pool/pull/465.

#### Pull Request Checklist
  used.


ffranr (approved):
I think that all looks good to me.

auctioneerrpc/auctioneer.proto
```diff
@@ -373,6 +373,11 @@ enum ChannelType {
     outputs that pay directly to the channel initiator (the seller).
     */
     SCRIPT_ENFORCED_LEASE = 2;
+
+    /*
+    A channel type that uses a Pay-to-Taproot output for the funding output.
+    */
+    SIMPLE_TAPROOT = 3;
```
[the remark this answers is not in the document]
guggero: Yes, you are correct... Unfortunately we started this file/enum before we were aware of that. So in this specific case I'd vote for consistency with the existing enum names rather than having one name being different than the others (we can't rename the existing ones without breaking older gRPC or REST clients).

funding/manager.go
```diff
@@ -689,14 +699,20 @@ func (m *Manager) BatchChannelSetup(
...
+			)
 			private := matchedOrderBid.UnannouncedChannel
+
+			log.Debugf("Opening channel to node=%x, private=%v, "+
+				"chan_point=%v, chan_amt=%v, commit_type=%v, "+
+				"musig2=%v, zero_conf=%v, thaw_height=%v",
+				matchedOrder.NodeKey[:], private, chanPoint,
+				chanAmt, commitmentType, chanPointShim.Musig2,
+				matchedOrderBid.ZeroConfChannel,
+				chanPointShim.ThawHeight)
```
Roasbeef, line 715: Style nit: missing newline below here.

rpcserver.go
```diff
@@ -148,9 +186,19 @@ func (s *rpcServer) Start() error {
 		return fmt.Errorf("error in GetInfo: %v", err)
 	}
 
-	_, s.wumboSupported = info.Features[uint32(lnwire.WumboChannelsRequired)]
+	_, s.wumboSupported = info.Features[wumboRequired]
 	if !s.wumboSupported {
-		_, s.wumboSupported = info.Features[uint32(lnwire.WumboChannelsOptional)]
+		_, s.wumboSupported = info.Features[wumboOptional]
+	}
+	_, s.taprootChansSupported = info.Features[taprootChansStagingRequired]
```
Roasbeef, line 193: I think we just want something like: 
`s.taprootChansSupported = info.Features[taprootChansStagingRequired] != nil || info.Features[taprootChansStagingOptional] != nil`

etc

rpcserver.go
```diff
@@ -1369,6 +1424,38 @@ func (s *rpcServer) validateOrder(o order.Order, acct *account.Account,
...
+	if o.Details().ChannelType == order.ChannelTypeSimpleTaproot {
+		taprootNotSupportedErr := fmt.Errorf("connected lnd version "+
+			"%v does not support taproot channels, need at least "+
+			"v%d.%d.%d with --protocol.simple-taproot-chans "+
+			"turned on", s.server.lndServices.Version.Version,
+			taprootChannelsVersion.AppMajor,
+			taprootChannelsVersion.AppMinor,
+			taprootChannelsVersion.AppPatch)
+
+		// First check the lnd version itself.
```
Roasbeef, line 1437: I think we can check this on start up, the cache the response so we don't need to hit version each time for an order submission.

guggero: We do exactly that. The variable `s.server.lndServices.Version` contains the cached version and `lndclient.AssertVersionCompatible()` just does the comparison of the two versions.

Roasbeef (approved):
LGTM 🧿

PR #485, opened 2024-05-31
Bump lnd to version v0.18.0-beta

guggero:
#### Pull Request Checklist
  used.


account/watcher/watcher_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	secp "github.com/decred/dcrd/dcrec/secp256k1/v4"
-	gomock "github.com/golang/mock/gomock"
+	gomock "go.uber.org/mock/gomock"
```
bhandras, line 12: Interesting, so this is a fork?

bhandras (approved):
LGTM 🎉

account/watcher/watcher_test.go
guggero: Yeah, apparently the original one was archived. Found out while attempting to fix the broken unit test by upgrading the mock.

go.mod
```diff
@@ -181,4 +197,6 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 
 replace github.com/lightninglabs/pool/auctioneerrpc => ./auctioneerrpc
 
-go 1.19
+go 1.21.4
+
+toolchain go1.22.2
```
ViktorTigerstrom, line 202: Hmm shouldn't this be:
```suggestion
toolchain go1.22.3
```

Although that change should maybe be in the last commit?

ViktorTigerstrom (approved):
LGTM except one minor comment :rocket:! Thanks a lot

go.mod
guggero: Was generated automatically, but yes you're right. Updated.

PR #487, opened 2024-08-19
poolrpc: convert poolrpc into a go module

bhandras:
This commit transforms the `poolrpc` subpackage into a versioned module. By doing so, projects that rely solely on the generated RPC code from the proto files will benefit from significantly reduced dependencies during compilation.

#### Pull Request Checklist
  used.


go.mod
```diff
@@ -26,9 +27,9 @@ require (
 	github.com/urfave/cli v1.22.9
 	go.etcd.io/bbolt v1.3.7
 	go.uber.org/mock v0.4.0
-	golang.org/x/sync v0.6.0
-	google.golang.org/grpc v1.59.0
-	google.golang.org/protobuf v1.33.0
+	golang.org/x/sync v0.8.0
+	google.golang.org/grpc v1.64.1
```
guggero, line 31: Same question here re the version we choose for `google.golang.org/grpc` and `google.golang.org/protobuf`. IMO they should be the same for all projects, as we've had issues with mismatches and updates to new versions before...

guggero (commented):
Thanks for the PR.

go.mod
bhandras: Given that `frdrpc` depends on GRPC v1.65.0 and protobuf v1.34.2 I bumped to those versions. Also added a new forked hex display (see here: https://github.com/lightninglabs/protobuf-go-hex-display/pull/3). PTAL

poolrpc/go.mod
```diff
@@ -0,0 +1,20 @@
...
+
+require (
+	golang.org/x/net v0.27.0 // indirect
+	golang.org/x/sys v0.22.0 // indirect
+	golang.org/x/text v0.17.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240814211410-ddb44dafa142 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240814211410-ddb44dafa142 // indirect
+)
+
+replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.34.2-hex-display
```
guggero, line 20: I think we should probably also add the replace rule for `auctioneerrpc`? `replace github.com/lightninglabs/pool/auctioneerrpc => ../auctioneerrpc`

guggero (approved):
One comment around a local replace directive. Otherwise LGTM :tada:

poolrpc/go.mod
bhandras: SGTM, done.

PR #495, opened 2025-01-03
chore: fix some typos

peicuiping:
#### Pull Request Checklist
- [x] LndServices minimum version has been updated if new lnd apis/fields are
  used.
 fix some typos

### DIFF
--- a/docs/batch_execution.md
+++ b/docs/batch_execution.md
@@ -22,7 +22,7 @@ Note that the trader can reject signing the batch for any reason, even when the 
 
 ### Batch Publication
 
-When all participating traders have signed their inputs in the Batch Execution Transaction, the auctioneer can sign the final input and broadcast the transaction. This transaction can be large, and serve as the funding transaction for potentially hundres of channels! The participating traders only pay chain fees for their inputs and outputs in the transaction, so everybody is saving substantially on fees compared to individually funding channels. If the trader supports account autorenewal and the account was close to expire, its expiry height will be automatically extended after the batch is sucessfully executed.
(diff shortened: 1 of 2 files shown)


docs/batch_execution.md
```diff
@@ -22,7 +22,7 @@ Note that the trader can reject signing the batch for any reason, even when the
...
+When all participating traders have signed their inputs in the Batch Execution Transaction, the auctioneer can sign the final input and broadcast the transaction. This transaction can be large, and serve as the funding transaction for potentially hundres of channels! The participating traders only pay chain fees for their inputs and outputs in the transaction, so everybody is saving substantially on fees compared to individually funding channels. If the trader supports account autorenewal and the account was close to expire, its expiry height will be automatically extended after the batch is successfully executed.
```
dstadulis, line 25:
```suggestion
When all participating traders have signed their inputs in the Batch Execution Transaction, the auctioneer can sign the final input and broadcast the transaction. This transaction can be large, and serve as the funding transaction for potentially hundreds of channels! The participating traders only pay chain fees for their inputs and outputs in the transaction, so everybody is saving substantially on fees compared to individually funding channels. If the trader supports account autorenewal and the account was close to expire, its expiry height will be automatically extended after the batch is successfully executed.
```

Another typo in the same line can be corrected by amending your commit.

Thanks for the fixes!

dstadulis (changes requested):
Thanks!

docs/batch_execution.md
peicuiping: Thanks. Modified!

peicuiping:
Thanks for your review!

PR #496, opened 2025-03-04
slog: update deps for LND, aperture, lndclient and btclog

ellemouton:
Depends on: 
- https://github.com/lightningnetwork/lnd/pull/9576
- https://github.com/lightninglabs/aperture/pull/148

Update the LND, aperture, lndclient and btclog deps. 
This includes upgrading the logging framework to the V2 framework. This
new framework lets us deprecate the `maxlogfiles` and
`maxlogfilesize` config options.


.golangci.yml
```diff
@@ -112,6 +97,9 @@ linters:
     - interfacebloat
 
 issues:
+  # Only show newly introduced problems.
+  new-from-rev: c5b37702e0bead2408ba72aad792ea4ff9fc930f
```
ellemouton, line 101: hope this is ok?
(see num of issues given the linter version bump: https://github.com/lightninglabs/pool/actions/runs/13659731510/job/38187830849)

ellemouton: hmm is something other than this needed to get it to do the thing? linter step still seems to be including the whole code-base

guggero: Yeah, probably need to add a couple of new linters to the ignore list.

auctioneerrpc/go.mod
```diff
@@ -1,8 +1,16 @@
 module github.com/lightninglabs/pool/auctioneerrpc
 
-go 1.15
+go 1.22.6
```
guggero, line 3: nit: 1.23.6?

tools/go.mod
```diff
@@ -1,188 +1,203 @@
 module github.com/lightninglabs/pool/tools
 
-go 1.18
+go 1.23.0
```
guggero, line 3: nit: use `1.23.6` here and remove `toolchain` below?

poolrpc/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/lightninglabs/pool/poolrpc
 
-go 1.22.3
+go 1.22.6
```
guggero, line 3: nit: use 1.23.6 and remove toolchain below?

guggero (commented):
Thanks a lot!

Can we also bump the Go version to 1.23.6 in the different Docker files (and everywhere else we might have an old version)?

bhandras (approved):
☕

PR #517, opened 2026-05-15
.github: add gateway workflow

saubyk:
Adds the gateway consumer workflow shim that triggers on PR comments, review re-requests, and PR close events.


Roasbeef (approved):
LGTM 🐧

PR #518, opened 2026-05-20
auctioneer: reconnect on EOF instead of silently leaving the subscribe stream dead

djkazic:
## Summary

Treat `io.EOF` on the `SubscribeBatchAuction` stream the same as any other transport-level stream error: surface it as `ErrServerErrored` so the rpcserver consumer triggers a reconnect via `HandleServerShutdown`. Previously EOF was emitted as a separate `ErrServerShutdown` sentinel that the consumer silently ignored, on the (incorrect) assumption that the client had already scheduled its own reconnect.

The result was a permanently dead subscription stream after any clean stream close, with the auctioneer filtering the trader's orders out of matching as "offline" until the process was restarted.

## Background

`auctioneer.Client.readIncomingStream` distinguishes two stream-failure cases:

- **Non-EOF errors** (`transport is closing`, `Unavailable`, etc.) emit `ErrServerErrored` to `StreamErrChan`. The consumer in `rpcserver.go` sees this and calls `HandleServerShutdown`, which closes the dead stream, re-dials, and re-runs `StartAccountSubscription` for every previously subscribed account.
- **EOF** emitted `ErrServerShutdown`, which the consumer treated as a no-op with the comment "the client has already scheduled a restart."

The comment was only accurate for *one* specific code path: when the auctioneer explicitly sends a `SubscribeError_SERVER_SHUTDOWN` *application* message, the read loop handles it inline (calls `HandleServerShutdown(nil)` directly and returns). For any *transport-level* EOF / proxy / load-balancer idle timeout, gRPC `MaxConnectionAge`, server stream handler returning for some other reason, server crash without graceful shutdown etc., no reconnect was ever scheduled.

Compounding this, `closeStream` was also never called on the EOF path, so any subsequent subscribe attempt from elsewhere short-circuited at the "already subscribed" guard in `connectAndAuthenticate` without ever sending a fresh Commit message on a new stream.

In production this manifested as: the trader's persistent stream died, the client process kept running and continued serving unary RPCs (so account / order CLI calls worked normally), but every batch attempt flagged `offline trader` against that account until the `poold` process was restarted.

## Tests

Adds `auctioneer/client_test.go` with three table-style tests against `readIncomingStream` driven by a fake stream:
  - `TestReadIncomingStreamEOFTriggersReconnect` — the regression case; fails on pre-fix code with `expected ErrServerErrored on EOF, got: server shutting down`.
  - `TestReadIncomingStreamTransportErrorTriggersReconnect` — locks in the pre-existing behaviour for non-EOF transport errors so the two paths stay unified.
  - `TestReadIncomingStreamContextCanceledDoesNotReconnect` — asserts client-initiated cancels don't double-trigger reconnect.

### DIFF
--- a/auctioneer/client.go
+++ b/auctioneer/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand/v2"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -929,6 +930,16 @@ func (c *Client) IsSubscribed() bool {
 	return c.serverStream != nil
 }
 
+// jitterBackoff returns backoff with up to 25% additive jitter so reconnect
+// attempts from a population of traders observing the same disconnect event
+// fan out over a window rather than spike at one instant. The jitter is
+// one-sided so we never wait less than the operator's configured floor.
+//
+//nolint:gosec // Backoff jitter doesn't need cryptographic randomness.
+func jitterBackoff(backoff time.Duration) time.Duration {
+	return backoff + time.Duration(rand.Int64N(int64(backoff)/4+1))
+}
+
 // connectServerStream opens the initial connection to the server for the stream
 // of account updates and handles reconnect trials with incremental backoff.
 func (c *Client) connectServerStream(initialBackoff time.Duration,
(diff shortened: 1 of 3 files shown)


djkazic:
Adding `jitter` for reconnects to avoid thundering herds (the motive of the original change that introduced this behavior).

auctioneer/client.go
```diff
@@ -945,8 +956,11 @@ func (c *Client) connectServerStream(initialBackoff time.Duration,
 	)
 	for i := 0; i < numRetries; i++ {
 		// Wait before connecting in case this is a reconnect trial.
+		// Apply additive jitter so a population of traders that all
+		// hit the same disconnect event don't fan in on the server at
+		// exactly the same instant.
 		if backoff != 0 {
-			err = c.wait(backoff)
+			err = c.wait(jitterBackoff(backoff))
```
Roasbeef, line 963: Could expand this with an exponential back off on top of the jitter.

Roasbeef (approved):
Exp backoff already in place 

LGTM 🕊️

PR #520, opened 2026-05-21
auctioneer: fix initial per-account subscribe failure

djkazic:
## Problem

`auctioneer.Client.connectAndAuthenticate` registers a new `acctSubscription` in `c.subscribedAccts` *before* running the 3-way auth handshake, and only removes it again via `closeStream` (called on full client shutdown or server-disconnect reconnect).

Every per-account error path returns without deleting the entry, including the legitimate `ACCOUNT_DOES_NOT_EXIST` response that `RecoverAccounts` triggers when probing keys that aren't known to the auctioneer yet.

The next call to `StartAccountSubscription` for that account key, typically from `handleStateOpen` after the on-chain funding tx confirms, hits the "already subscribed" early-return guard, returns `nil` without sending a fresh `Commit`, and the per-account handshake never runs. The client thinks the account is subscribed; the auctioneer has never authenticated it. The trader is silently filtered as `offline trader` at matching time and stays that way until the process is restarted (which clears the in-memory map).

## Symptoms
The bug fires whenever RecoverAccounts probes a key the auctioneer hasn't seen yet. This happens naturally when:
    - A user opens an account and runs recovery before the funding tx confirms.
    - A user restores from seed and has accounts in flight.
    - A user has had failed account creation attempts (the recovery loop walks past "holes" in the key sequence; every probed-but-not-found key seeds a stale entry).

Litd / lightning-terminal users disproportionately affected. The terminal UI invokes `RecoverAccounts` as part of its account-list refresh flow, so any restart of litd with a not-yet-confirmed account is enough to set the trap.

## Fix
Add a `success` sentinel plus a `defer` in `connectAndAuthenticate` that removes the entry from `c.subscribedAccts` unless the handshake reached the `GetSuccess` branch. The defer is set up *after* the "already-subscribed" early-return guard, so it does not delete a pre-existing entry on that path.

`srvMsg.GetSuccess() != nil` is the only branch that sets `success = true`; every other exit (authenticate error, unrecognized response, recovery-mode `INCOMPLETE_ACCOUNT_RESERVATION`, recovery-mode `ACCOUNT_DOES_NOT_EXIST`, unknown error, unknown message) cleans up. The cleanup is keyed on `acctPubKey`, captured at the top of the function and never re-derived, so the deferred delete cannot remove an entry for a different account.

## Tests
Adds `TestConnectAndAuthenticateCleansUpOnAccountDoesNotExist` in `auctioneer/client_test.go`. It drives a full `connectAndAuthenticate` call in recovery mode against a scripted fake stream: capture the Commit, echo its `commitHash` back in a Challenge, drain the Subscribe, then respond with `SubscribeError_ACCOUNT_DOES_NOT_EXIST`. Asserts `connectAndAuthenticate` returns `(sub, false, nil)` *and* `c.subscribedAccts` is empty afterwards.

Verified the test catches the bug: against the unmodified `connectAndAuthenticate`, it fails with `subscribedAccts entry was not cleaned up after ACCOUNT_DOES_NOT_EXIST; later subscribes for the same account would silently no-op`.

### DIFF
--- a/auctioneer/client.go
+++ b/auctioneer/client.go
@@ -643,6 +643,25 @@ func (c *Client) connectAndAuthenticate(ctx context.Context,
 	c.subscribedAcctsMtx.Lock()
 	c.subscribedAccts[acctPubKey] = sub
 	c.subscribedAcctsMtx.Unlock()
+
+	// The subscription needs to be in the map before authenticate runs so
+	// that readIncomingStream can route the server's Challenge/Error
+	// responses back to it. But if the handshake doesn't reach the
+	// GetSuccess branch below, the entry is stale — a later subscribe
+	// attempt for the same account would hit the "already subscribed"
+	// guard at the top of this function and silently no-op without ever
+	// sending a fresh Commit. Clean up here so only genuinely live
+	// subscriptions remain in the map.
+	success := false
+	defer func() {
+		if success {
+			return
+		}
+		c.subscribedAcctsMtx.Lock()
+		delete(c.subscribedAccts, acctPubKey)
+		c.subscribedAcctsMtx.Unlock()
+	}()
+
 	err := sub.authenticate(ctx)
 	if err != nil {
 		log.Errorf("Authentication failed for account %x: %v",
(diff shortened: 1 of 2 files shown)


Roasbeef (approved):
LGTM 🥬

